PR #689, opened 2014-11-21
EclipseWriter: add possibility to write out the transmissibilities which are used by the simulator

andlaus:
the quantities are attributed to the positive faces of logically
Cartesian cells. I have not found a way to visualize the quanties in
the INIT file using ResInsight, so this is written once per timestep
for now...

Note that the TRANNNC stuff is a bit of a work-arround, but I could not figure out a canonical way to query the grid if a face is a NNC or not. (any hints are appreciated.)


atgeirr:
I have tentatively put the upcoming release milestone on this PR, but I need some guidance: should it be merged, postponed or dropped?

atgeirr:
Since I had no replies for this at (or since) the time of last release I will close this PR tomorrow unless asked not to. It may be of special interest to @qilicun @totto82 @joakim-hove and @andlaus.

andlaus:
I propose closing this: IIRC it is intended to be a debugging tool and not same the output as produced by E100.

joakim-hove:
I Agree.
19. okt. 2015 16:51 skrev "Andreas Lauser" <notifications@github.com>
følgende:

PR #700, opened 2014-12-09
POC: components for ERT

akva2:
Use this to not announce ecl_well as a dependency across all projects since symbols only end up in a unit test of opm-core.

@bska POC of what we discussed yesterday. this way the find rule can be synced across projects just fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,8 @@ macro (files_hook)
 endmacro (files_hook)
 
 macro (tests_hook)
+	find_package(ERT COMPONENTS ecl_well REQUIRED)
+	list(APPEND opm-core_LIBRARIES ${ERT_LIBRARY})
 	cond_disable_test ("ERT")
 endmacro (tests_hook)
 
--- a/cmake/Modules/FindERT.cmake
+++ b/cmake/Modules/FindERT.cmake
@@ -60,53 +60,68 @@ find_path (ERT_GEN_INCLUDE_DIR
 if (CMAKE_SIZEOF_VOID_P)
   math (EXPR _BITS "8 * ${CMAKE_SIZEOF_VOID_P}")
 endif (CMAKE_SIZEOF_VOID_P)
-find_library (ERT_LIBRARY_ECL
-  NAMES "ecl"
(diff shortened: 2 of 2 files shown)


andlaus:
> Use this to not announce ecl_well as a dependency across all projects since symbols only end up in a unit test of opm-core.

hm, as far as I can see the point of this PR, it is about not linking to some ERT libraries if they have not been explicitly specified as prerequisites, analogous to the boost `COMPONENTS` mechanism?

I think this is a worthy cause and should be merged, but I have two questions: 

1) why is the stuff in the `test_hook()` in 'CMakeLists.txt' necessary? shouldn't the `ERT COMPONENTS $FOO` directive be in `opm-core-prereqs.cmake`?
2) If a downstream module needs more ERT components than opm-core, are the appropriate tests run?

akva2:
On 26/03/15 12:12, Andreas Lauser wrote:
> I think this is a worthy cause and should be merged, but I have two 
> questions:
>
> 1) why is the stuff in the |test_hook()| in 'CMakeLists.txt' 
> necessary? shouldn't the |ERT COMPONENTS $FOO| directive be in 
> |opm-core-prereqs.cmake|?
> 2) If a downstream module needs more ERT components than opm-core, are 
> the appropriate tests run?
>
1) see POC prefix. it was a quick way to test it..
2) yes, they should be, though i haven't actually tried it.

atgeirr:
Since this is a POC, @akva2, does that mean it should *not* be merged as is?

akva2:
On 26/03/15 12:40, Atgeirr Flø Rasmussen wrote:
>
> Since this is a POC, @akva2 <https://github.com/akva2>, does that mean 
> it should /not/ be merged as is?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/OPM/opm-core/pull/700#issuecomment-86478021>.
>
definitely not as-is.

atgeirr:
> definitely not as-is.

Thanks for clarifying. I'll move this to the fall release to see if anyone can and will finish it before then.

atgeirr:
I am not against merging the feature, but the target for this PR has moved to opm-common, so I will close this PR.

PR #705, opened 2014-12-12
FindAlugrid.cmake: do not try to include the whole grid

andlaus:
first this makes the test very slow, second (and more importantly) it
makes the test fail on dune master as the header relies on the
presence of the DUNE_GRID_VERSION* macros which are not present at the
configure stage of the build...


andlaus:
this PR still works as advertised, but there still is the issue of whether the new test program (which basically checks if `dune/alugrid/grid.hh` is includable) is sufficient. I suppose that it is, but @dr-robertk is the authoritative person for this.

PR #708, opened 2014-12-16
build system: do not separate the debugging symbols by default

andlaus:
the reason is that the tools for profiling and for coverage analysis
(i.e., gprof and gcov) have trouble with this and crash. if you want
this to be enabled for whatever reason, use the cmake flag
-DSTRIP_DEBUGGING_SYMBOLS=ON or pass configure the
--enable-strip-debug option.

### DIFF
--- a/cmake/Modules/OpmCompile.cmake
+++ b/cmake/Modules/OpmCompile.cmake
@@ -1,5 +1,7 @@
 # - Compile main library target
 
+option (STRIP_DEBUGGING_SYMBOLS "use separate files for the executable code and the debugging symbols" OFF)
+
 macro (opm_compile opm)
   # some CMake properties do not do list expansion
   string (REPLACE ";" " " ${opm}_LINKER_FLAGS_STR "${${opm}_LINKER_FLAGS}")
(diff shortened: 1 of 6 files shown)


andlaus:
from the technical POV this one seems to be ready to go in: I verified that it still compiles and that it works with `-DSTRIP_DEBUG_SYMBOLS={ON|OFF}`

atgeirr:
I am inclined to agree that we should change the default to make it less likely that tools will fail etc. I will therefore merge this now. For those who want or need the old behaviour: if you experience trouble using the STRIP_DEBUG_SYMBOLS option to achieve it, pipe up and we'll try to fix it, or even revert if absolutely necessary.

GitPaean:
It looks like under `RelWithDebugInfo` mode, it uses `-DNDEBUG`, which makes it skip all the `assert`s. Preferably, the `assert` checking will be kept for the mode `RelWithDebugInfo` for debugging purpose.

akva2:
changing this would be a bit problematic as RelWithDebug is used in packaging (and then symbols are extracted into a separate -dbg package). not unsolvable but it would be nonstandard behavior.

atgeirr:
I agree with @akva2, since NDEBUG will be used to disable time-consuming checks it would be better to keep it for the RelWithDebugInfo mode. If there was a DebugWithOptimization or similar mode it would be appropriate (to keep those checks by not defining NDEBUG) for that though... In any case, a developer should be able to override this to enable those checks, right?

GitPaean:
`DebugWithOptimization` will be good. I was not aware of the standard meaning of these options. I was assuming that the `RelWithDebugInfo` is also for debugging purpose like for some relatively bigger models that were impossible to debug without optimization.

atgeirr:
Unfortunately there is no cmake standard build type DebugWithOptimization, it was just a wish. But as long as you can manually change the options when you run cmake you should be able to achieve what you want.

Issue #268, opened 2013-06-28
Change the source term in tutorial2?


rolk:
tutorials/tutorial2.cpp declares a grid of 40x40x1 blocks using the GridManager constructor where dx, dy is implicitly set to 1. Thus the area of each side is 1m^2.

It then proceeds to set a source term on the first and last block, which in the Cartesian grid are two corners. The active surface of those two blocks are thus 2m^2 each.

Now, from examples/sim_2p_incomp.cpp and computePorevolume in opm/core/utility/miscUtilities.cpp we see evidence that the source term is actually a volumetric flux in volume units (meters) per second, meaning that tutorial2 has a source term of 100m^3/s, yielding a velocity of 50m/s (and a Reynolds number on the order 10^7).

Isn't this a tad... unrealistic?

PR #713, opened 2015-01-09
Introduce interface for SimulatorTimers and support writing of substeps with EclipseWriter.

dr-robertk:
This PR introduces an interface for SimulatorTimers implemented by SimulatorTimer and AdaptiveSimulatorTimer. 

The EclipseWriter has been revised to support writing of substeps when used with the AdaptiveTimeStepping.

It has been tested with the SPE9 case and the solution as well as the dates look ok.

### DIFF
--- a/opm/core/io/eclipse/EclipseWriter.hpp
+++ b/opm/core/io/eclipse/EclipseWriter.hpp
@@ -1,6 +1,7 @@
 /*
   Copyright (c) 2013 Andreas Lauser
   Copyright (c) 2013 Uni Research AS
+  Copyright (c) 2014 IRIS AS
 
   This file is part of the Open Porous Media project (OPM).
 
@@ -39,10 +40,12 @@ namespace Opm {
 // forward declarations
 namespace EclipseWriterDetails {
 class Summary;
+struct WriterTimer;
 }
 
 class SimulatorState;
 class SimulatorTimer;
+class AdaptiveSimulatorTimer;
 class WellState;
 
 namespace parameter { class ParameterGroup; }
@@ -91,10 +94,33 @@ public:
      * ERT or ECLIPSE. Note that calling this method is only
      * meaningful after the first time step has been completed.
      *
+     * \param[in] timer          The timer providing time step and time information
      * \param[in] reservoirState The thermodynamic state of the reservoir
-     * \param[in] wellState The production/injection data for all wells
+     * \param[in] wellState      The production/injection data for all wells
      */
     virtual void writeTimeStep(const SimulatorTimer& timer,
+                               const SimulatorState& reservoirState,
+                               const WellState& wellState);
+
+
+    /*!
+     * \brief Write a reservoir state and summary information to disk.
+     *
+     *
+     * The reservoir state can be inspected with visualization tools like
+     * ResInsight.
+     *
+     * The summary information can then be visualized using tools from
+     * ERT or ECLIPSE. Note that calling this method is only
+     * meaningful after the first time step has been completed.
+     *
+     * \param[in] timer          The timer providing time step and time information
+     * \param[in] subStepTimer   The timer providing sub step information
+     * \param[in] reservoirState The thermodynamic state of the reservoir
+     * \param[in] wellState      The production/injection data for all wells
+     */
+    virtual void writeTimeStep(const SimulatorTimer& timer,
+                               const AdaptiveSimulatorTimer& subStepTimer,
                                const SimulatorState& reservoirState,
                                const WellState& wellState);
 
@@ -117,6 +143,13 @@ private:
     std::shared_ptr<EclipseWriterDetails::Summary> summary_;
 
     void init(const parameter::ParameterGroup& params);
+    // implementation of writeInit
+    void writeInit(const EclipseWriterDetails::WriterTimer &timer);
+    // implementation of writeTimeStep
+    void writeTimeStep(const EclipseWriterDetails::WriterTimer& timer,
+                       const SimulatorState& reservoirState,
+                       const WellState& wellState);
+
 };
 
 typedef std::shared_ptr<EclipseWriter> EclipseWriterPtr;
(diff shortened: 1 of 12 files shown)


dr-robertk:
I just double checked. This PR also depends on PR #268 from opm-autodiff.

atgeirr:
Thanks for the PR, I am reviewing it now. I'll make comments in the diff.

opm/core/simulator/SimulatorTimerInterface.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <boost/date_time/posix_time/posix_time_types.hpp>
+#include <boost/date_time/posix_time/conversion.hpp>
+
+namespace Opm
+{
+
+    namespace parameter { class ParameterGroup; }
+
+    /// Interface class for SimulatorTimer objects, to be improved.
+    class SimulatorTimerInterface
```
atgeirr: I notice that the class does not contain the operator++. Is this because its semantics is different between subclasses? I am not sure I have understood fully how the AdaptiveSimulatorTimer is supposed to be used: does ++ increase its report or substep?

opm/core/simulator/SimulatorTimerInterface.hpp
```diff
@@ -0,0 +1,103 @@
...
+        /// beginning of the current time step [s].
+        virtual double simulationTimeElapsed() const = 0;
+
+        /// Return true if timer indicates that simulation of timer interval is finished
+        virtual bool done() const = 0;
+
+        /// Return start date of simulation
+        virtual boost::posix_time::ptime startDateTime() const = 0;
+
+        /// Return the current time as a posix time object.
```
atgeirr: It seems to me that this method and currentPosixTime() need not be virtual and perhaps should not be.

opm/core/simulator/SimulatorTimer.hpp
```diff
@@ -21,20 +21,23 @@
...
 {
 
     namespace parameter { class ParameterGroup; }
 
-    class SimulatorTimer
+    class SimulatorTimer : public SimulatorTimerInterface
     {
     public:
+        // use default implementation of these methods
+        using SimulatorTimerInterface::currentDateTime;
```
atgeirr: I think these using statements are superfluous and only needed if there is an overload with the same name in this class.

opm/core/io/eclipse/EclipseWriter.hpp
```diff
@@ -39,10 +41,10 @@ namespace Opm {
 // forward declarations
 namespace EclipseWriterDetails {
 class Summary;
+struct WriterTimer;
```
atgeirr: Is something missing? I do not see any definition of WriterTimer, yet it is used in two functions.

opm/core/io/eclipse/EclipseWriter.hpp
```diff
@@ -110,13 +114,20 @@ class EclipseWriter : public OutputWriter
...
-    int reportStepIdx_;
+    int writeStepIdx_;
     std::string outputDir_;
     std::string baseName_;
     PhaseUsage phaseUsage_; // active phases in the input deck
     std::shared_ptr<EclipseWriterDetails::Summary> summary_;
 
     void init(const parameter::ParameterGroup& params);
+    // implementation of writeInit
+    void writeInit(const EclipseWriterDetails::WriterTimer &timer);
```
atgeirr: I cannot see that this matches the implementation? Also, the ampersand should be attached to the type.

opm/core/simulator/AdaptiveSimulatorTimer.cpp
```diff
@@ -86,12 +88,23 @@ namespace Opm
...
     double AdaptiveSimulatorTimer::currentStepLength () const
     {
         assert( ! done () );
         return dt_;
     }
 
+    double AdaptiveSimulatorTimer::stepLengthTaken() const
+    {
+        assert( ! steps_.empty() );
+        return *(steps_.rbegin());
```
atgeirr: I suggest steps_.back() for readability.

opm/core/simulator/SimulatorTimer.hpp
dr-robertk: I think it increases readability of the code, but I can remove them if needed.

opm/core/simulator/AdaptiveSimulatorTimer.cpp
dr-robertk: Fixed.

opm/core/simulator/SimulatorTimerInterface.hpp
dr-robertk: operator++ returns a reference to the object. Personally, I would rename the method to advance and make it void, and the one could also add a virtual method in the interface class. In general I see the SimulatorTimerInterface more as the read-only part of the timers.

opm/core/io/eclipse/EclipseWriter.hpp
dr-robertk: This is a left over from the previous implementation. I removed it.

opm/core/simulator/SimulatorTimerInterface.hpp
dr-robertk: That's a matter of taste. If somebody wants to overload this for some reason then that would be a nice source of errors.

opm/core/simulator/SimulatorTimerInterface.hpp
atgeirr: I like the advance() suggestion, if you want to do it I'm in favor. Keeping this as submitted is also ok.

atgeirr:
I'm fine with this PR now (and its cousin in opm-autodiff). Do you want it merged now, or would you like to  do any more changes?

dr-robertk:
I would like to have this merged. Next week I'll try to improve the output in opm-autodiff to be more encapsulated. But for now the way it is is good enough.

PR #714, opened 2015-01-12
Added WellCompletion namespace

joakim-hove:
Renames due to: https://github.com/OPM/opm-parser/pull/394


bska:
Thanks!

bska:
PR OPM/opm-parser#394 appears to be as yet unmerged.  Therefore, this breaks the build.  Moreover, I think you missed the declaration of `WellsManagerDetail::computeWellIndex()` in [`WellsManager_impl.hpp`](https://github.com/OPM/opm-core/blob/master/opm/core/wells/WellsManager_impl.hpp).

PR #718, opened 2015-01-12
Avoid multiple symbol definitions for MinPvProcessor.

blattms:
The non-template class MinpvProcessor contains all function
definitions inside of the header file without inlining. The
results in the symbols being in the library of opm-core and
in case of using CpGrid an addtional occurence in other
libraries (opm-autodiff?).

This commit changes this by simply inlining all these functions.

Note that another (better?) option would be to move at least
the bigger once into a cpp file.

### DIFF
--- a/opm/core/grid/MinpvProcessor.hpp
+++ b/opm/core/grid/MinpvProcessor.hpp
@@ -53,7 +53,7 @@ namespace Opm
         std::array<int, 3> delta_;
     };
 
-    MinpvProcessor::MinpvProcessor(const int nx, const int ny, const int nz)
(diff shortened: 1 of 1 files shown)


atgeirr:
Thanks for noticing, my mistake in the first place. I am not 100% sure if it is sufficient to declare the functions inline inside the class definition, and not in the function definitions. Do you know for sure the answer?

blattms:
I must admit that I do not know that. But specifying inline with definition (and not the declaration) will surely be ok. I changed the PR in this regard

atgeirr:
Thank you, will merge!

PR #719, opened 2015-01-13
Adds a free function to UgGridHelpers to access the face tag.

blattms:
This is needed for grid independant transmissibility calculations


opm/core/grid/GridHelpers.cpp
```diff
@@ -63,6 +63,11 @@ double faceArea(const UnstructuredGrid& grid, int face_index)
     return grid.face_areas[face_index];
 }
 
+int faceTag(const UnstructuredGrid& grid, int face_index)
+{
+    return grid.cell_facetag[face_index];
```
bska: Alas, this is semantically wrong--or at least misleading.  The `cell_facetag` table is supposed to be indexed by "half face indices" that designate pairs of cells and faces.  Creating an implementation that is linear in the number of connections on a face is trivial with the altered interface
```C++
int
faceTag(const UnstructuredGrid& grid, const int cell, const int face)
{
    assert (0 <= cell);  assert (cell < grid.number_of_cells);
    assert (0 <= face);  assert (face < grid.number_of_faces);

    int i = grid.cell_facepos[cell];
    for (int e = grid.cell_facepos[cell + 1]; i < e; ++i) {
        if (face == grid.cell_faces[i]) {
            break;
        }
    }

    if (i == grid.cell_facepos[cell + 1]) {
        OPM_THROW(std::logic_error,
                  "Connection '" << face << "' not on cell '" << cell << "'");
    }

    return grid.cell_facetag[i];
}
```
but a constant-time implementation either requires that `i` be passed as input data or that we create an ancillary table of mappings from pairs of (cell,face) to "half face" indices.

blattms: Hugh? I must admit that I do not get how your solution is supposed to work, ad cell_facetag needs to indexed with the "half face" index, doesn't it?

blattms:
It is not wrong (but the CpGrid version is), just improper documentation. face_index is suppoed to be the "half face" index as used in the transmissibility calculations.

blattms:
Anyway It seems I have to think about it a bit longer to come up with a solution that fits both Cpgrid and Ug. Be patient.

opm/core/grid/GridHelpers.cpp
bska: > I do not get how your solution is supposed to work,

My suggestion was based on the assumption that `face_index` really was a face index (renamed `face` for brevity).  My two-parameter (excluding the grid) function computes the requisite half-face index from a pair of cell and face indices and returns the appropriate tag.

blattms:
@bska Cool that is a nice hint to get both UgGrid and CpGrid working. Thanks!

bska:
> that is a nice hint to get both `UnstructuredGrid` and `CpGrid` working. Thanks!

Yeah, at least as a first step.  I would generally prefer a constant time implementation, especially given the fact that the way we usually traverse the grid constructs `i` as a by-product, but that requires changing at least one of the grid implementations.

With a view towards parallel implementations, it would perhaps be even better to consider an interface of the form
```C++
int
cellFaceTag(const G& grid, const int cell, const int i);
```
with `G` being a grid (`UnstructuredGrid` or `CpGrid`), `cell` being a cell index and `i` denoting the local ID number of the connection in question.  Then the `UnstructuredGrid` implementation of `cellFaceTag()` would be something like
```C++
int
cellFaceTag(const UnstructuredGrid& grid, const int cell, const int i)
{
    assert (0 <= cell);  assert (cell < grid.number_of_cells);
    assert (0 <= i   );  assert (i + grid.cell_facepos[cell] <
                                 grid.cell_facepos[cell + 1]);

    return grid.cell_facetag[grid.cell_facepos[cell] + i];
}
```
and we could add a `SparseTable<int> cell_facetag_` to class `CpGridData`, populated with the appropriate values (using for instance a construction similar to function `fill_cell_topology()` in [`opm/core/grid/cornerpoint_grid.c`](https://github.com/OPM/opm-core/blob/master/opm/core/grid/cornerpoint_grid.c)) and expose the data as a `cellFaceTag()` method of the form
```C++
int
cellFaceTag(const int cell, const int i) const
{
    return current_view_data_->cell_facetag_[cell][i];
}
```
Of course, implementing `boundaryId()` in terms of `cellFaceTag()` in the case of `!uniqueBoundaryIds()` is then either a linear operation in terms of the number of connections on a cell or will require some other, pre-computed translation table.  There are benefits and drawbacks to both alternatives.

In any case, design decisions and cost/benefit analysis abounds in this area.

blattms:
I opened a new pull request that will solve the problems for CpGrid and UG.

PR #720, opened 2015-01-13
Adds a free function to UgGridHelpers to access the face tag.

blattms:
To identify the face tag we use the iterator over the cell faces. This provides all the information needed both for UG and CpGrid.

For UG we can directly compute the half face index from by subtracting
```C++
&(*cellFaces()[0].begin())
```
from it. For `CpGrid` the iterator will provide the cell index and we will use Bard's proposal to compute the face tag correctly.

Supersedes and closes #719

### DIFF
--- a/opm/core/grid/GridHelpers.cpp
+++ b/opm/core/grid/GridHelpers.cpp
@@ -63,6 +63,12 @@ double faceArea(const UnstructuredGrid& grid, int face_index)
     return grid.face_areas[face_index];
 }
 
+int faceTag(const UnstructuredGrid& grid,
+            boost::iterator_range<const int*>::const_iterator face)
+{
+    return grid.cell_facetag[&(*face)-&(*(cell2Faces(grid)[0].begin()))];
+}
+
 SparseTableView cell2Faces(const UnstructuredGrid& grid)
 {
     return SparseTableView(grid.cell_faces, grid.cell_facepos, numCells(grid));
--- a/opm/core/grid/GridHelpers.cpp
+++ b/opm/core/grid/GridHelpers.cpp
@@ -66,7 +66,7 @@ double faceArea(const UnstructuredGrid& grid, int face_index)
 int faceTag(const UnstructuredGrid& grid,
             boost::iterator_range<const int*>::const_iterator face)
 {
-    return grid.cell_facetag[&(*face)-&(*(cell2Faces(grid)[0].begin()))];
+    return grid.cell_facetag[face-cell2Faces(grid)[0].begin()];
 }
 
 SparseTableView cell2Faces(const UnstructuredGrid& grid)
--- a/opm/core/grid/GridHelpers.hpp
+++ b/opm/core/grid/GridHelpers.hpp
@@ -183,7 +183,6 @@ double faceArea(const UnstructuredGrid& grid, int face_index);
 
 /// \brief Get Eclipse Cartesian tag of a face
 /// \param grid The grid that the face is part of.
-/// \param cell_index The index of the cell that the face is attached to.
 /// \param cell_face The face attached to a cell as obtained from cell2Faces()
 /// \return 0, 1, 2, 3, 4, 5 for I-, I+, J-, J+, K-, K+
 int faceTag(const UnstructuredGrid& grid, boost::iterator_range<const int*>::const_iterator cell_face);
--- a/opm/core/grid/GridHelpers.hpp
+++ b/opm/core/grid/GridHelpers.hpp
@@ -181,6 +181,13 @@ const double* faceNormal(const UnstructuredGrid& grid, int face_index);
 /// \param face_index The index of the face in the grid.
 double faceArea(const UnstructuredGrid& grid, int face_index);
 
+/// \brief Get Eclipse Cartesian tag of a face
+/// \param grid The grid that the face is part of.
+/// \param cell_index The index of the cell that the face is attached to.
+/// \param cell_face The face attached to a cell as obtained from cell2Faces()
+/// \return 0, 1, 2, 3, 4, 5 for I-, I+, J-, J+, K-, K+
+int faceTag(const UnstructuredGrid& grid, boost::iterator_range<const int*>::const_iterator cell_face);
+
 /// \brief Maps the grid type to the associated type of the cell to faces mapping.
 ///
 /// Provides a type named Type.


blattms:
Please wait until I posted similar PRs for dune-cornerpoint and opm-autodiff to get the whole picture.

opm/core/grid/GridHelpers.hpp
```diff
@@ -181,6 +181,13 @@ const double* faceNormal(const UnstructuredGrid& grid, int face_index);
 /// \param face_index The index of the face in the grid.
 double faceArea(const UnstructuredGrid& grid, int face_index);
 
+/// \brief Get Eclipse Cartesian tag of a face
+/// \param grid The grid that the face is part of.
+/// \param cell_index The index of the cell that the face is attached to.
```
bska: There is no `cell_index` parameter to `faceTag()`.  Its documentation should therefore be removed.

opm/core/grid/GridHelpers.cpp
```diff
@@ -63,6 +63,12 @@ double faceArea(const UnstructuredGrid& grid, int face_index)
     return grid.face_areas[face_index];
 }
 
+int faceTag(const UnstructuredGrid& grid,
+            boost::iterator_range<const int*>::const_iterator face)
+{
+    return grid.cell_facetag[&(*face)-&(*(cell2Faces(grid)[0].begin()))];
```
atgeirr: While I think the pointer arithmetic is correct, I think using std::distance would be more readable. But perhaps I am a little optimistic: the iterators are not strictly from the same containers, is that why you convert it all to pointers?

opm/core/grid/GridHelpers.hpp
blattms: @bska Fixed in a new commit. Upon request I will fixup both to be one
commit.

@atgeirr I have used pointer arithmetic, because the underlying data
is a C-array. The iterators are over the range of the cells of one
face only. Unfortunately```boost::iterator_range::iterator``` is what
they call a forward traversal iterator (no ```operator-(iter1,iter2)``` 
which would give us linear complexity when using ```std::distance```.
Currently pointer arithmetic seems the only choice.

Using boost::iterator_range might not been the smartest choice in the
first place. But we should change that in another PR. Feel free to
open a feature request.

atgeirr: I assumed the boost::iterator_range iterator would be an RA iterator, but if not it would be a mistake to use distance(), I agree. Apart from that, I think using boost for this is good, let's not reinvent more wheels than required.

bska: > I assumed the `boost::iterator_range` iterator would be an RA iterator,

"It depends".  Specifically, it depends on the category of the underlying iterator.  As can be inferred from the documentation for class [`boost::iterator_range`](  http://www.boost.org/doc/libs/1_57_0/libs/range/doc/html/range/reference/utilities/iterator_range.html) (link to Boost 1.57, Boost 1.44 is the same as far as I can tell), the `iterator_range<>::iterator` has the category of the underlying iterator.

In other words, if `boost::iterator_range<>` is instantiated with a random access iterator, then the derived iterator is random access too.  I'll furthermore point out that the `iterator_range` offers the methods
```C++
reference operator[]( difference_type at ) const;
value_type operator()( difference_type at ) const;
size_type size() const;
```
but these are specific to random access iterators.

opm/core/grid/GridHelpers.cpp
bska: > I think the pointer arithmetic is correct

I'm not convinced, but I'd be happy to be enlightened.  The `iterator_range` is defined, in parts, as
```C++
template <class ForwardTraversalIterator>
class iterator_range {
public:
    // ...
    typedef ForwardTraversalIterator const_iterator;
    // ...
};
```
In other words, `boost::iterator_range<const int*>::const_iterator` is an alias of `const int*`.  Consequently, `*face` is "simply" a temporary `const int` and taking its address yields a pointer to that temporary.  The relation to the underlying contiguous array is thereby lost as far as I can tell.

Similarly, `cell2Faces(const UnstructuredGrid& grid)` returns a `SparseTableView` (see [`GridHelpers.hpp`](https://github.com/OPM/opm-core/blob/master/opm/core/grid/GridHelpers.hpp) for detailed definition) whose `operator[]()` method returns an `iterator_range<const int*>` corresponding to the specified row (i.e., cell).

Dereferencing that range's `begin()` method yields another temporary `const int` whose address is a pointer to that single `const int` without any relation to the underlying contiguous array.  If (and that's a big if) this analysis is correct, I'd feel a lot more comfortable with a slightly different implementation.  For instance, something along the lines of
```C++
int
faceTag(const UnstructuredGrid&                   grid,
        SparseTableView::row_type::const_iterator cface)
{
    auto start = cell2Faces(grid)[0].begin();
    auto i     = std::distance(start, cface);

    return grid.cell_facetag[i];
}
```
This code obviously assumes that `face` is valid iterator into the connection array of one of the `grid`'s cells, but that was the case before too.

Comments?

blattms:
@bska You are right. I did read the documentation but somehow the first sentences confused me into thinking that I only get the forward iterator interface.

Just added a commit that skips the pointer arithmetic.

opm/core/grid/GridHelpers.cpp
atgeirr: I am pretty sure that the `&*iter` ideom is sound, and indeed is the standard way of getting a pointer to the element, the dereferencing should return a reference.

bska: > the dereferencing should return a reference.

Well, in this case it doesn't because the iterator is an unadorned `const int*`.

atgeirr: I just had to test this... The following program:
```
#include <iostream>

int main()
{
    const int q = 7;
    const int* pq = &q;
    const int* refderefpq = &(*pq);
    std::cout << pq << ' ' << refderefpq << std::endl;
}
```
produces the output
```
0x7fff504faa9c 0x7fff504faa9c
```
on my machine which seems to indicate that indeed it behaves as I expect. Also see the page https://www.sgi.com/tech/stl/trivial.html in the Invariants section. I think that also supports my view. It is probably written in the standard somewhere, but I wouldn't know where.

blattms: So we agree that both versions do the same thing, right?

The new one is just a bit shorter and IMHO nicer. But I can revert it if you wish.

atgeirr:
I'm fine with leaving the face - something[0].begin() as it is. There is only one theoretical problem, that a debugging implementation could check that the iterator operator- operates on iterators from the same iterator_range. I assumed this is why you did the pointer-stuff.

blattms:
As Bard noted, the iterator_range::{begin|end}() return the underlying
iterator, in this case const int*. I assume that the compiler or
stdlib has no way of checking whether these are in the range.

atgeirr:
True, it would only be possible if the iterators came from a checked class such as vector.

blattms:
Thanks a lot!

PR #721, opened 2015-01-14
Protecting against use of unassigned deck values.

joakim-hove:
This should fix: https://github.com/OPM/opm-parser/issues/396. The problem is default handling - that is seriously difficult to get Right&trade; The current approach is that:

  1. All default values mentioned in the Json config should be sensible, i.e. use values like -1 to indicate default behavior should be avoided.
  2. A `DeckItem` which does not have value (coming either from the deck or as a default) will throw an exception when you ask for the data - that was what happened here.

The clear ambition is that all these details should be handled 100% in the parser, and the external modules should only use the "fully processed" `EclipseState` object - that is probably the long term for this problem as well.


andlaus:
besides my  difference in opinion of whether the parser should be allowed to throw on `getItem()` for successfully created decks, the patch looks fine...

joakim-hove:
OK - with positive test results from @totto82 I'll merge this.

PR #723, opened 2015-01-16
AdaptiveTimeStepping improvements.

dr-robertk:
This PR introduced a max growth factor for the timestep size as well as some fixes for the times step computatation and improvement of the output.


totto82:
Tested on norne. This PR make the mini-steps more distributed and the time-step growth more robust after a convergence failure. i.e it works as it should.

atgeirr:
Looks good, will merge.

PR #724, opened 2015-01-16
Not creating Logger instance explicitly.

joakim-hove:
Followup to: https://github.com/OPM/opm-parser/pull/398


joakim-hove:
Merging this now in agreement with @atgeirr. There is certainly more to with the logging framework itself, in particular *verbosity*, but the changes in downstream modules `opm-core` and `opm-autodiff` are very small anyway.

PR #725, opened 2015-01-19
Make MinpvProcessor ignore inactive cells.

atgeirr:
This changes significantly the volume of some cells after processing. Before this, large inactive cells with low or zero porosity could contribute large pore volumes to a simulation since their volumes could be added to active cells in MINPV processing.

Please report if this improves pore volume match, do not merge this (or its companion OPM/dune-cornerpoint#122) until it has been tested.

### DIFF
--- a/opm/core/grid/GridManager.cpp
+++ b/opm/core/grid/GridManager.cpp
@@ -158,7 +158,7 @@ namespace Opm
 
         if (!poreVolumes.empty() && eclipseGrid->isMinpvActive()) {
             MinpvProcessor mp(g.dims[0], g.dims[1], g.dims[2]);
-            mp.process(poreVolumes, eclipseGrid->getMinpvValue(), zcorn.data());
+            mp.process(poreVolumes, eclipseGrid->getMinpvValue(), actnum, zcorn.data());
         }
 
         const double z_tolerance = eclipseGrid->isPinchActive() ?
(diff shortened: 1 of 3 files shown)


qilicun:
The API changes should also adapted to ``test_minpvprocessor.cpp``

qilicun:
I did the comparison, total PV in OPM is **673740009.896**, in ECL is **673148278.062**, the relative error is 
**0.000879051248205**. For absolute error and relative error of every cells, I made two plots.
abs_pv
rel_pv

atgeirr:
> The API changes should also adapted to test_minpvprocessor.cpp

Oops, I forgot to commit that. Should be in now.

Judging from your figures, we still have a few cells with significantly different pore volume, but a lot less difference overall than before, right?

qilicun:
>Judging from your figures, we still have a few cells with significantly different pore volume, but a lot less difference overall than before, right?

Yes, I output these cells  with cartesian index and other information, you can find [here](https://gist.github.com/qilicun/8c864776cfa440662ea6). The criterion is ``|opm_pv -ecl_pv| >10``

bska:
I'm not entirely convinced that we're not missing something fundamental in this part, but as this improves a real case and causes no fall-out I'll merge this and the companion PR OPM/dune-cornerpoint#122 into master.

Thanks a lot.

PR #728, opened 2015-01-21
Added methods for computing global reductions.

blattms:
We need to compute quite a few global reductions in the Newton method of opm-autodiff. This commit adds the functionality to compute several reductions combined using only one global communication. Compiles and test succeeds with one or more process.

I have added a test for the reductions and moved some common functions with test_parallel_linearsolver.cpp to a new header DuneIstlTestHelper.hpp to prevent code duplication.

### DIFF
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -158,7 +158,7 @@ public:
                 }
             }
         }
-    };
+    }
     /// \brief Compute one or more global reductions.
     ///
     /// This function can either be used with a container, an operator, and an initial value
@@ -223,7 +223,6 @@ private:
         std::tuple<ReturnValues...> init=values;
         updateOwnerMask(std::get<0>(containers));
         computeLocalReduction(containers, operators, values);
-        auto val=std::get<0>(values);
         std::vector<std::tuple<ReturnValues...> > receivedValues(communicator_.size());
         communicator_.allgather(&values, 1, &(receivedValues[0]));
         values=init;
@@ -468,7 +467,9 @@ namespace Opm
 /// \param grid The grid to inspect.
 
 inline void extractParallelGridInformationToISTL(boost::any& anyComm, const UnstructuredGrid& grid)
-{}
+{
+    (void)anyComm; (void)grid;
+}
 } // end namespace Opm
 
 #endif
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -345,7 +345,7 @@ private:
         template<class T, class T1>
         T operator()(const T& t1, const T& t2, const T1& mask)
         {
-            b_(t1, maskValue(t2, mask));
+            return b_(t1, maskValue(t2, mask));
         }
         template<class T, class T1>
         T maskValue(const T& t, const T1& mask)
@@ -377,7 +377,7 @@ private:
         template<class T, class T1>
         T operator()(const T& t1, const T& t2, const T1& mask)
         {
-            b_(t1, maskValue(t2, mask));
+            return b_(t1, maskValue(t2, mask));
         }
         template<class T, class T1>
         T maskValue(const T& t, const T1& mask)
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -130,8 +130,10 @@ public:
       OwnerOverlapSet sourceFlags;
       AllSet destFlags;
       Dune::Interface interface(communicator_);
-      if(!remoteIndices_->isSynced())
+      if( !remoteIndices_->isSynced() )
+      {
           remoteIndices_->rebuild<false>();
+      }
       interface.build(*remoteIndices_,sourceFlags,destFlags);
       Dune::BufferedCommunicator communicator;
       communicator.template build<T>(interface);
@@ -141,14 +143,20 @@ public:
     template<class T>
     void updateOwnerMask(const T& container)
     {
-        if(! indexSet_)
+        if( ! indexSet_ )
+        {
             OPM_THROW(std::runtime_error, "Trying to update owner mask without parallel information!");
-        if(container.size()!= ownerMask_.size())
+        }
+        if( container.size()!= ownerMask_.size() )
         {
             ownerMask_.resize(container.size(), 1.);
-            for(auto i=indexSet_->begin(), end=indexSet_->end(); i!=end; ++i)
+            for( auto i=indexSet_->begin(), end=indexSet_->end(); i!=end; ++i )
+            {
                 if (i->local().attribute()!=Dune::OwnerOverlapCopyAttributeSet::owner)
+                {
                     ownerMask_[i->local().local()] = 0.;
+                }
+            }
         }
     };
     /// \brief Compute one or more global reductions.
@@ -207,8 +215,10 @@ private:
         static_assert(std::tuple_size<std::tuple<Containers...> >::value==
                       std::tuple_size<std::tuple<ReturnValues...> >::value,
                       "We need the same number of containers and return values");
-        if(std::tuple_size<std::tuple<Containers...> >::value==0)
+        if( std::tuple_size<std::tuple<Containers...> >::value==0 )
+        {
             return;
+        }
         // Copy the initial values.
         std::tuple<ReturnValues...> init=values;
         updateOwnerMask(std::get<0>(containers));
@@ -217,9 +227,11 @@ private:
         std::vector<std::tuple<ReturnValues...> > receivedValues(communicator_.size());
         communicator_.allgather(&values, 1, &(receivedValues[0]));
         values=init;
-        for(auto rvals=receivedValues.begin(), endvals=receivedValues.end(); rvals!=endvals;
-            ++rvals)
+        for( auto rvals=receivedValues.begin(), endvals=receivedValues.end(); rvals!=endvals;
+             ++rvals )
+        {
             computeGlobalReduction(*rvals, operators, values);
+        }
     }
     /// \brief TMP for computing the the global reduction after receiving the local ones.
     ///
@@ -257,7 +269,7 @@ private:
                           std::tuple<ReturnValues...>& values)
     {
         const auto& container = std::get<I>(containers);
-        if(container.size())
+        if( container.size() )
         {
             auto& reduceOperator  = std::get<I>(operators);
             auto newVal = container.begin();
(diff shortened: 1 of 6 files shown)


atgeirr:
Thanks for the contribution! I would assume that this sort of utility would live in dune-istl usually, but you put it here since we have to be compatible with older dune versions?

While variadic templates should be part of gcc 4.4, I would appreciate if @bska could verify that this PR builds on the older RHEL system.

I do not feel competent to review the MPI parts, but I trust that with a unit test in place the code should be fine, also it does not affect any existing code (I think). I will make a few comments in the diff though.

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -1,5 +1,7 @@
 /*
-  Copyright 2014 Dr. Markus Blatt - HPC-Simulation-Software & Services
+  Copyright 2014, 2015 Dr. Markus Blatt - HPC-Simulation-Software & Services
+  Copyright 2014 Statoil ASA
+  Copyright 2015 NTNU
```
atgeirr: Attention to legal details, that is good...!

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -115,9 +137,143 @@ class ParallelISTLInformation
...
+            return;
+        // Copy the initial values.
+        std::tuple<ReturnValues...> init=values;
+        updateOwnerMask(std::get<0>(containers));
+        computeLocalReduction(containers, operators, values);
+        auto val=std::get<0>(values);
+        std::vector<std::tuple<ReturnValues...> > receivedValues(communicator_.size());
+        communicator_.allgather(&values, 1, &(receivedValues[0]));
+        values=init;
+        for(auto rvals=receivedValues.begin(), endvals=receivedValues.end(); rvals!=endvals;
```
atgeirr: Please use braces to delimit code blocks (also the if () above and some other places). We have seen bugs due to this before.

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -153,8 +309,141 @@ class ParallelISTLInformation
 
     std::shared_ptr<ParallelIndexSet> indexSet_;
     std::shared_ptr<RemoteIndices> remoteIndices_;
-    MPI_Comm communicator_;
+    Dune::CollectiveCommunication<MPI_Comm> communicator_;
+    mutable std::vector<double> ownerMask_;
```
atgeirr: Why is `double` used for the mask? I would imagine that `char` would be better (assuming we want to stay away from the non-container std::vector<bool>)?

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -153,8 +309,141 @@ class ParallelISTLInformation
...
+        }
+        BinaryOperator& localOperator()
+        {
+            return b_;
+        }
+    private:
+        BinaryOperator b_;
+    };
+
+    /// \brief An operator that converts the values where mask is 0 to the minimum value
```
atgeirr: I was a little surprised to see that the masked operators are applied where the mask is 0, not 1. Is it a common convention and I'm just not familiar with it? Or is there a particular reason for it in this context?

tests/DuneIstlTestHelpers.hpp
```diff
@@ -0,0 +1,220 @@
...
+    int procs, rank;
+    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+    MPI_Comm_size(MPI_COMM_WORLD, &procs);
+    int n = N/procs; // number of unknowns per process
+    int bigger = N%procs; // number of process with n+1 unknows
+
+
+    int start, end, istart, iend;
+    // Compute owner region
+    if(rank<bigger) {
```
atgeirr: I do very much prefer spaces around binary ops and next to if/for/while etc. (but not function calls). I'm considering it a lost cause to have a uniform style however...

bska:
> I would appreciate if @bska could verify that this PR builds on the older RHEL system.

I have launched a build test including downstream modules in
```
{ debug, release } x { sequential, MPI } x { Dune 2.2.1, Dune 2.3.1 }
```
on the CentOS 5.11 system we have.  At the current rate I estimate it will take at least an hour to complete.?

bska:
I said
> I have launched a build test including downstream modules [...] At the current rate I estimate it will take at least an hour to complete

It actually went through a bit faster.  I got some diagnostics however, and I wonder if the "no return statement in function returning non-void" might be problematic.  They are all in the implementations of `*Operator::operator()` e.g.,
```C++
template<typename BinaryOperator>
struct MaskToMinOperator
{
    // ...
    template<class T, class T1>
    T operator()(const T& t1, const T& t2, const T1& mask)
    {
        b_(t1, maskValue(t2, mask));
    }

    // ...
private:
    BinaryOperator b_;
};
```

Here's the complete list of diagnostics with the initial number being number of occurrences
```
3 ./opm-core/opm/core/linalg/ParallelIstlInformation.hpp:153: warning: extra ‘;’
2 ./opm-core/opm/core/linalg/ParallelIstlInformation.hpp:216: warning: unused variable ‘val’
1 ./opm-core/opm/core/linalg/ParallelIstlInformation.hpp:336: warning: no return statement in function returning non-void
1 ./opm-core/opm/core/linalg/ParallelIstlInformation.hpp:368: warning: no return statement in function returning non-void
1 ./opm-core/opm/core/linalg/ParallelIstlInformation.hpp:412: warning: no return statement in function returning non-void
3 ./opm-core/opm/core/linalg/ParallelIstlInformation.hpp:445: warning: unused parameter ‘anyComm’
3 ./opm-core/opm/core/linalg/ParallelIstlInformation.hpp:445: warning: unused parameter ‘grid’
2 ./opm-core/tests/DuneIstlTestHelpers.hpp:180: warning: no previous declaration for ‘std::tuple<int, int, int, int> computeRegions(int)’
2 ./opm-core/tests/DuneIstlTestHelpers.hpp:59: warning: extra ‘;’
1 ./opm-core/tests/test_parallel_linearsolver.cpp:47: warning: no previous declaration for ‘void run_test(const Opm::parameter::ParameterGroup&)’
```

A bit more worrying is the fact that I got the following failure in `make test` in release mode of the MPI build when targeting Dune 2.3.1.  Is there some additional information I can provide that will help analyse the problem?
```
12/41 Testing: parallelistlinformation
12/41 Test: parallelistlinformation
Command: "/home/bska/work/opm/build/opm/opt/mpi/dune-2.3.1/opm-core/bin/test_parallelistlinformation"
Directory: /home/bska/work/opm/build/opm/opt/mpi/dune-2.3.1/opm-core/tests
"parallelistlinformation" start time: Jan 22 11:20 CET
Output:
----------------------------------------------------------
Running 2 test cases...
/home/bska/work/opm/src/opm/opm-core/tests/test_parallelistlinformation.cpp(57): error in "tupleReductionTest": check std::get<0>(values)==std::get<0>(oldvalues)+((N-1)*N)/2 failed
/home/bska/work/opm/src/opm/opm-core/tests/test_parallelistlinformation.cpp(79): error in "singleContainerReductionTest": check value==oldvalue+((N-1)*N)/2 failed

*** 2 failures detected in test suite "OPM-ParallelIstlInformation"
<end of output>
Test time =   0.10 sec
----------------------------------------------------------
Test Failed.
"parallelistlinformation" end time: Jan 22 11:20 CET
"parallelistlinformation" time elapsed: 00:00:00
----------------------------------------------------------
```

tests/DuneIstlTestHelpers.hpp
blattms: Don't give up on style guideline. I added some chages to adhere to your guideline

opm/core/linalg/ParallelIstlInformation.hpp
blattms: Not sure whether there is a convention for masks, but the approach used is convenient. Maybe the name is just missleading.
I am using it to select the dofs owned by process which should provide one partition of a partitioning. With one for owner and zero for the other nodes I can easily select the interesting ones with an axpy.

blattms:
@bard I am unable to reproduced the problem. Do you used the official tarballs? What build options (flags, build type)?

bska:
> Do you used the official tarballs?

Yes--at least I think so.  I don't think I've contaminated my Dune install directory.

> What build options (flags, build type)?

These are the build flags for Dune:
```sh
$ cat opts-release-mpi
CONFIGURE_FLAGS="--enable-fieldvector-size-is-method --disable-documentation --disable-mpiruntest --disable-expressiontemplates --enable-parallel --with-superlu=/work/ --with-superlu-lib=libsuperlu_4.3.a --prefix=/home/bska/work/opm/inst/dune/2.3.1/opt/mpi CXX=mpic++ CC=mpicc F77=mpif77 FC=mpif90 CXXFLAGS='-DNOOUTPUT -pipe -O3 -DNDEBUG -Wall -Wextra -pedantic -fno-strict-aliasing -march=native -mtune=native' CFLAGS='-DNOOUTPUT -pipe -std=c99 -O3 -DNDEBUG -Wall -Wextra -pedantic -fno-strict-aliasing -march=native -mtune=native' "
MAKE_FLAGS="-j 4"
BUILDDIR=BUILD-OPT-MPI

$ mpic++ --version
g++44 (GCC) 4.4.7 20120313 (Red Hat 4.4.7-1)
Copyright (C) 2010 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Here are the build flags for Opm:
```

CMAKE_CXX_FLAGS:STRING=-std=c++0x -pipe -pedantic -Wall -Wextra -Wcast-align -Wcast-qual -Wchar-subscripts -Wformat-nonliteral -Wmissing-declarations -Wpointer-arith -Wredundant-decls -Wundef -Wwrite-strings -Wno-unknown-pragmas -ggdb3

//Flags used by C++ compiler during Debug builds
CMAKE_CXX_FLAGS_DEBUG:STRING=-ggdb3 -O0 -UNDEBUG -DDEBUG

//Flags used by the compiler during release minsize builds.
CMAKE_CXX_FLAGS_MINSIZEREL:STRING=-O3 -DNDEBUG -mtune=native

//Flags used by the compiler during release builds (/MD /Ob1 /Oi
// /Ot /Oy /Gs will produce slightly less optimized but smaller
// files).
CMAKE_CXX_FLAGS_RELEASE:STRING=-O3 -DNDEBUG -mtune=native

//Flags used by the compiler during Release with Debug Info builds.
CMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING=-O3 -DNDEBUG -mtune=native
```
This targets OpenMPI 1.8.0 built with the same compiler (`mpic++`).

bska:
> I am unable to reproduced the problem.

As a matter of fact, so am I.  As of commit 0fb801d `make test` completes without error on my test system.  Moreover, it appears that 0fb801d fixes a real problem on this system because 420586d leads to failure in `parallelistlinformation`.

bska:
I'm going to trust you on the MPI bits.  As of commit 0fb801d everything builds and all tests run on the reference system, so I'm going to merge this into master.

Thanks a lot.

blattms:
It turns out, that you should not trust me. There are still some severe bugs, that my tests did not trigger.  (Wouldn't have found them without your report).

I will post another PR once I am sure that they are gone.

PR #729, opened 2015-01-21
Build system update

andlaus:
this PR contains a few smallish build system updates which enable ZOLTAN for alugrid, fix the METIS detection on some distributions and work around a dune 2.4 deprecation warning (the last patch is pretty hacky, but the dune build system does this in a similar way). most of the work was done by @dr-robertk.


atgeirr:
Thanks, looks fine to my eyes.

PR #730, opened 2015-01-22
More features for flow diagnostics

atgeirr:
This PR collects the following features:
 - `IncompTpfaSinglePhase` and `IncompPropertiesSinglePhase`, new classes intended to make it easy to run simplified diagnostic computations on more or less arbitrary cases.
 - A rewritten `compute_tof` example program, that should now be able to run with deck input containing arbitrary number of phases and related properties (used to require 2-phase oil-water), using the above mentioned classes.
 - New diagnostics functions `computeFandPhi()`, `computeLorenz()` and `computeSweep()` in `FlowDiagnostics.hpp` with unit tests.

While somewhat large, the PR contains no changes to existing features, other than to the compute_tof program. Some caveats relating to `compute_tof`:
 - The `compute_tof` program treats the reservoir in a very simplified fashion: using oil-phase properties as basis for single-phase pressure solve, and treating all injectors and producers with simplified BHP controls instead of the controls given in the deck (which may be inappropriate for a single-phase solve, such as LRAT). This simplification is intended, allowing a quick look at a case straight from the deck.
 - The program does not incorporate the full complexity of transmissibility multipliers and pore volume manipulation that is available for example in opm-autodiff's `sim_fibo_ad`. Partially because some of this treatment is currently implemented in opm-autodiff and not in opm-core.
 - The program treats the well situation as existing at the initial time step. Given a complex schedule it is anyway more appropriate to do diagnostics from fluxes and mobilities as computed by a simulator.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -112,6 +112,7 @@ list (APPEND MAIN_SOURCE_FILES
 	opm/core/simulator/SimulatorTimer.cpp
 	opm/core/tof/AnisotropicEikonal.cpp
 	opm/core/tof/DGBasis.cpp
+	opm/core/tof/FlowDiagnostics.cpp
 	opm/core/tof/TofReorder.cpp
 	opm/core/tof/TofDiscGalReorder.cpp
 	opm/core/transport/TransportSolverTwophaseInterface.cpp
@@ -164,6 +165,7 @@ list (APPEND TEST_SOURCE_FILES
   tests/test_ug.cpp
 	tests/test_cubic.cpp
 	tests/test_event.cpp
+	tests/test_flowdiagnostics.cpp
 	tests/test_nonuniformtablelinear.cpp
 	tests/test_sparsevector.cpp
 	tests/test_sparsetable.cpp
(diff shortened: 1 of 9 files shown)


bska:
```
/opm/core/tof/FlowDiagnostics.cpp:57: error: ‘sort’ is not a member of ‘std’
```

atgeirr:
Looks like <numeric> perhaps includes <algorithm> or something like that on my system. Thanks for catching, should be fixed now.

bska:
> Looks like `<numeric>` perhaps includes `<algorithm>` or something like that on my system.

Could be.  Anyway, this looks mostly mechanical in nature with some refinement and bug-fixing so I'm inclined to merge once I've verified that all tests pass on CentOS 5.11.

I do notice what appears to be space-versus-tab issues in the `IncompTpfaSinglePhase.hpp` header so maybe you could fix those in the interim?

bska:
Oh, and on another code formatting issue, I **personally** prefer to list the return type on a separate line when it takes up as much horizontal space as
```C++
std::pair<std::vector<double>, std::vector<double>>
```
Otherwise the function name and parameters get flushed too far to the right in my opinion.  I know that we usually disagree on that though.

atgeirr:
Reformatting done.

bska:
> Reformatting done.

Thanks.  Everything builds and all tests run.  I'm going to merge this into master.

PR #732, opened 2015-01-23
Fixes global reductions

blattms:
Correctly compute the minimum and maximum values.

As there are no functors for computing the minimum and maximum,
we convert the std::max and std::min function pointers to
functors (which is not really nice.) Previously we were somehow
tricked into using std::greater and std::less, which of course do
return true or false and not what we need. Additionally, do more
excessive testing with different ranges.

This is a follow up to #728 that fixes some additional issues. Sorry for not finding this earlier.


atgeirr:
Thanks for the fix, will merge.

PR #734, opened 2015-01-26
Add another way of MINPV process.

qilicun:
If the cell's pv is less than MINPV, set it to be non-actived.

After this PR, the relative PV error is **3.36752074434e-06**, the maximum cell pv relative error is less than **3.5%**

pv


joakim-hove:
Iff this is indeed the final verdict on `MINPV` processing I would argue that it should be in the parser.

alfbr:
We still have PINCH, which needs to know about MINPV (and MINPVV) cell removal, so the parser will have to provide that too. Generally speaking I agree, applying MINPV, and passing which cells were removed, seems to fit best in the parser.

qilicun:
Please hold on merging after PINCH is checked.

atgeirr:
This will not connect cells across the MINPV cell, which should be happening if PINCH also is active, so I think the topology will be wrong after this. To obtain the exact cell volumes eclipse uses I propose instead adding a little loop to the bottom of GridManager::readFromEclipseGrid() instead, writing PORV/PORO to the grid cell volume array: then we have the right topology already, and can get by simply overwriting the volumes.

joakim-hove:
>To obtain the exact cell volumes eclipse uses I propose instead adding a little loop to the bottom 

I don't really understand what you are proposing here; but to me it looks like you might be running around in circles instead of using the values from the parser?

atgeirr:
We are indeed running around in a circle, but we have to: we need to modify the zcorn to obtain the right topology during processing, but may not want that to influence the pore volumes.

qilicun:
As you discussed, I thinks this PR is less than full consideration, it's better not to merge it. Any suggestion is appreciated.

alfbr:
> we need to modify the zcorn to obtain the right topology during processing

Why do you need to modify zcorn?

atgeirr:
> Why do you need to modify zcorn?

The current processor first scans zcorn to identify nodes that are equal (within a tolerance if PINCH is active), then topology is created using these node identities to deduce connection information. If we just deactivate the MINPV cells, there will be no connection across the removed cell.

alfbr:
So you do not actually modify zcorn? It is rather that you need to know which cells MINPV removed when you apply PINCH. I haven't looked into the implementation details here, but naively it should be possible for the parser to pass on that information to the grid processor.

atgeirr:
We can of course rewrite the cornerpoint grid processor, I was just suggesting a way to obtain the result we want without doing that.

andlaus:
> If we just deactivate the MINPV cells, there will be no connection across the removed cell.

I think that (depending on transmissibilities) deactivating cells sometimes is the better option than pinching them: A large cell with very low porosity and transmissibility should better be deactivated than pinched out.

just my 2c...

alfbr:
> We can of course rewrite the cornerpoint grid processor

Yes, I believe adapting the processor is necessary if we move MINPV handling to the parser. Hopefully not a huge effort though.

> I think that (depending on transmissibilities) deactivating cells sometimes is the better option than pinching them:

I believe PINCH is typically an effective way of getting convergence, so it is more of a necessary step than to make the results more representative. In any case, it is a specific functionality. Either we support it, or we don't. It is widely used, so I suggest we do.

andlaus:
> I believe PINCH is typically an effective way of getting convergence

I know, but if the pinched out cell represents a blocking fracture, pinching them out will produce wrong results. So far, it was my assumption than we preferred correctness to performance?

alfbr:
> I know, but if the pinched out cell represents a blocking fracture, pinching them out will produce wrong results. So far, it was my assumption than we preferred correctness to performance?

We probably have differing views here. Once PINCH is set, pinching is the correct behaviour. Right now we are trying to sort out incorrect implementation of MINPV, and still ensure that PINCH behaves correctly. You are basically arguing that we should not support PINCH at all.

andlaus:
> We probably have differing views here. Once PINCH is set, pinching is the correct behaviour.

actually, I don't really care if implemented incorrectly (as long as I don't need to assume responsibility for such things): if Eclipse does it like this and you prioritize Eclipse behaviour above all else, that's perfectly okay with me.

My point is that this is often physically "wrong" in the sense that it leads to significantly different flow behaviour in the case of removing low permeability "fractures"...

> You are basically arguing that we should not support PINCH at all.

nope, I'm not: I'm arguing that pinching out high-permeability ones it is okay but that removing low-permeability ones is not.

alfbr:
> if Eclipse does it like this

Actually I don't care about how Eclipse does it, I simply want us to comply with the input file format. PINCH is a keyword with a specific documented behaviour. Either we support it, or we don't.

> I'm arguing that pinching out high-permeability cells is okay, but that removing low-permeability ones this way is not.

This sounds like a custom option not currently available in the input file format. If we want to support that functionality I suggest that we either make it a custom option to PINCH or provide it in a separate custom keyword. I guess the closest existing keywords to what you want is PINCHREG and PINCHNUM. That fact may suggest that few people agree with you.

andlaus:
> That fact may suggest that few people agree with you.

Since reality is not a democracy, that does not make my argument wrong, though ;)

alfbr:
True :-)

bska:
We currently modify the `ZCORN` array in order to create the requisite (vertical) non-neighbouring connections across pinched-out cells deactivated due to low pore volumes.  That's wrong and in order to do this correctly we need to change the low-level grid constructor (implementation of function `create_grid_cornerpoint()`).  Moreover, we need to introduce the notion of explicit non-neighbouring connection into the grid structure and to change all solvers to work in terms of a simulation graph rather than a grid.  That is a big job that will affect the entire OPM stack.

On the other hand, a simulation graph is the correct environment for the NNC keyword and has applications in aquifer handling too.  But I just want to reiterate the fact that getting completely correct behaviour in this area will require significant effort.  Our current approach produces a reasonable connection structure but the pore volumes will be off, sometimes significantly, compared to ECL.

> if the pinched out cell represents a blocking fracture,

then the information entered in PINCH controls whether or not to create connections across them and how to compute the derived transmissibility (especially the multipliers).  The issue here is the interaction of keywords MINPV and PINCH, not either of those in isolation.

> I'm arguing that pinching out high-permeability cells is okay, but that removing low-permeability ones this way is not.

Permeability is only tangentially involved.  If there is little volume available to flow the conductivity of the material doesn't really matter.  In particular, the local flow rate may well be high but very little mass will actually be transferred.  ECLIPSE offers several modes for handling the balance between available  volume, connection topology and connection strength.

qilicun:
>That's wrong and in order to do this correctly we need to change the low-level grid constructor (implementation of function create_grid_cornerpoint())

Yes, that's true. 
If I understand correctly, in OPM, PINCH get threshold value from deck, but not directly get in touch with MINPV. MINPV modify ZCORN, and during this process**maybe** some cells‘s value(I'm not sure what value) is under the threshold, so OPM can pinch-out these cells?

atgeirr:
Note that with the default ('GAP') PINCH behaviour, cells will be connected across a MINPV-ed cell irrespective of the size of the cell/gap. This is the behaviour we support.

qilicun:
From the discussion and MINPV implementation in OPM, I have the following impression, if there is anything wrong, I will be appreciated if any of you can point it out.

Let's put the PV formulation first $PV=|V|\cdot\phi\cdot NTG \cdot MULTPV$.

The current processor in OPM compares ZCORN (NOT MPINV)of identify nodes with the tolerance if PINCH is active, in order to use these feature , the MINPV processor in OPM scan the active cell, if it's PV is less than MINPV, the MINPV processor will move this cell's lower face to higher face(set the ZCORN of each nodes of the lower face to the same value of that in higher face) in z direction  and move the cell's(which below this cell) higher face to this cell's higher face, the result of the process is that these cells' geometry volumes is almost(or identity) zero(because the top face and bottom face is the same face), and these geometry volumes are added to the cell which below this cell.

To be more specified, assuming there are cells [c_1,  ...c_i, c_{i+1},...,c_n] , the geometry volume are [v_1, ... v_i,v_{i+1}, ...,v_n], if c_i's pore volume is less than MINPV, after the process, v_i will be 0, and v_{i+1} will become larger(probable will be v_i + v_{i+1}). This wil has problem, since c_i and c_{i+1} can have totally different porosity, NTG and MULTPV, and this can lead to a different pore volume.  So from this point of view, I think the MINPV process in OPM is not correct.

There maybe some ways to update the MINPV process.

- Just As @bska  and @atgeirr  said, rewrite the corner point grid process. But this needs lots of efforts.
- Don't remove the geometry volume added to the below cells during MINPV process when computing pore volume, this maybe leads to another problem, I have no idea.
- Read pove volume from ECLIPSE, this could be the easiest and safe way to correct PV in OPM.

The above is just some of my thoughts, any suggestion and idea are welcome.

atgeirr:
I think you have this correctly @qilicun. I do not quite understand what your second suggestion means ("Don't remove the geometry volume...") but your third suggestion is essentially what I already proposed some comments ago: "To obtain the exact cell volumes eclipse uses I propose instead adding a little loop to the bottom of GridManager::readFromEclipseGrid() instead, writing PORV/PORO to the grid cell volume array: then we have the right topology already, and can get by simply overwriting the volumes."

qilicun:
@atgeirr 

Yes, the third suggest is borrowed from you, sorry for not pointing it out.

In the second suggestion, since one cell's PV is less than MINPV, this cell's geometry volume is added to the cell which below this cell, so I thinking could we don't use the geometry volume added due to MINPV when we compute PV, just use the original geometry cell volume, even we actually modify that values after MINPV process.

qilicun:
Compare to reading PV from ECLIPSE, I prefer the second way, it can lead us to the correct PV and needn't to change the topology and doesn't rely on ECLIPSE on this point.
I did a test in the second way, total PV relative error is **3.42031918306e-08**, and also following is two plots. Do you think this is reasonable? @atgeirr 

pv_re

pv_abs

atgeirr:
I think this would be the third way... using the ERT library's routines. I did not think of this simple way of doing it though. Looks very good, but to get correct connections you would have to use OPM/opm-autodiff#287 together with the existing opm-core master (without this PR), right?

qilicun:
> but to get correct connections you would have to use OPM/opm-autodiff#287 together with the existing opm-core master (without this PR), right?

Yes. I think that PR looks more reasonable,  and we need to move discussion to that one also. I will close this one.

PR #735, opened 2015-01-26
Using Parser to calculate well reference depth.

joakim-hove:
Followup to: https://github.com/OPM/opm-parser/pull/402

Will use the parser Well->getRefDepth() for the well reference depth; the defaulted / not defaulted issues are left with the parser.


atgeirr:
Looks good, @joakim-hove please merge synchronously the opm-parser PR.

PR #737, opened 2015-01-28
Method advance for SimulatorTimers.

dr-robertk:
This is the interface method for the previously used operator++ which cannot be made virtual since the return type differs.


atgeirr:
Thanks! Just thinking about the documentation: I think advance() will always advance time by currentStepLength(), correct? The doc says so for the adaptive timer, but not the others. I think at least documenting this in the interface would be nice, alternatively referring to advance() in the doc for currentStepLength().

dr-robertk:
Fixed.

PR #738, opened 2015-01-28
Constify parallelinformation and workaround Eigen's weirdness

blattms:
* We need to compute the reductions on a const object and therefore this PR makes them const
* Eigen's containers have nor begin()/end() iterator access for whatever reason therefore we need
to rely on operator[] *sigh*

### DIFF
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -276,17 +276,22 @@ private:
         if( container.size() )
         {
             auto& reduceOperator  = std::get<I>(operators);
-            auto newVal = container.begin();
+            // Eigen:Block does not support STL iterators!!!!
+            // Therefore we need to rely on the harder random-access
+            // property of the containers. But this should be save, too.
+            // Just commenting out code in the hope that Eigen might improve
+            // in this regard in the future.
+            //auto newVal = container.begin();
             auto mask   = ownerMask_.begin();
             auto& value = std::get<I>(values);
-            value = reduceOperator.maskValue(*newVal, *mask);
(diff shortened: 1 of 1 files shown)


opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -263,30 +263,31 @@ class ParallelISTLInformation
...
                           std::tuple<BinaryOperators...>& operators,
-                          std::tuple<ReturnValues...>& values)
+                          std::tuple<ReturnValues...>& values) const
     {
         const auto& container = std::get<I>(containers);
         if( container.size() )
         {
             auto& reduceOperator  = std::get<I>(operators);
-            auto newVal = container.begin();
+            // F*#$&ing Eigen:Block does not support STL iterators!!!!
```
atgeirr: Please do not use inappropriate language in the source code.

atgeirr:
Other than the minor issue noted, this looks fine to me. I assume you have tested this on your system with MPI? I agree that it is strange that Eigen does not have begin() and end() members for 1d data structures, but it probably stems from being matrix-centric: should the iterator you get iterate over all elements, rows or columns? We could make a workaround though: create global begin() and end() functions taking the 1d arrays as arguments and returning pointers as iterators. This will also allow use of range-for in the future. I am not sure though if global begin() and end() are defined on gcc 4.4.

opm/core/linalg/ParallelIstlInformation.hpp
blattms: Rebased with improved impartial comment.

atgeirr: Strangely, it still looks identical to me.

PR #739, opened 2015-01-29
BlackoilState: added forgotten rv variable to equal method.

dr-robertk:
Maybe it was intended to not compare the rv states. It seems like a bug though.


bska:
> Maybe it was intended to not compare the `rv` states

`Rv didn't exist.when `BlackoilState::equals()` was created.

atgeirr:
Thanks for noticing!

PR #740, opened 2015-01-29
PVT classes: implement temperature dependent viscosity

andlaus:
This implements a temperature dependency of the oil and water viscosities according to the Eclipse documentation. (gas does not depend on temperature in E100 for some reason and we don't have a gas composition yet, so we also can't do the E300 approach.) After quite a bit of head-banging, I decided that I did not want to change the APIs for the PVT properties (which is only required once an energy conservation equation is implemented), and that only the viscosities are considered for now, i.e. densities are a job for another month...

### DIFF
--- a/opm/core/props/pvt/PvtConstCompr.hpp
+++ b/opm/core/props/pvt/PvtConstCompr.hpp
@@ -23,6 +23,8 @@
 #include <opm/core/props/pvt/PvtInterface.hpp>
 #include <opm/core/utility/ErrorMacros.hpp>
 
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+
 #include <vector>
 #include <algorithm>
 
@@ -98,7 +100,9 @@ namespace Opm
               ref_B_(1, 1.0),
               comp_(1, 0.0),
               viscosity_(1, visc),
-              visc_comp_(1, 0.0)
+              visc_comp_(1, 0.0),
+              oilvisctTables_(0),
+              watvisctTables_(0)
         {
         }
 
@@ -119,6 +123,13 @@ namespace Opm
                 int tableIdx = getTableIndex_(pvtRegionIdx, i);
                 double x = -visc_comp_[tableIdx]*(p[i] - ref_press_[tableIdx]);
                 output_mu[i] = viscosity_[tableIdx]/(1.0 + x + 0.5*x*x);
+
+                if (oilvisctTables_ != 0 || watvisctTables_ != 0) {
+                    // TODO: temperature dependence
+                    OPM_THROW(std::logic_error,
+                              "temperature dependent viscosity as a function of z "
+                              "is not yet implemented!");
+                }
             }
         }
 
@@ -146,7 +157,7 @@ namespace Opm
         virtual void mu(const int n,
                         const int* pvtRegionIdx,
                         const double* p,
-                        const double* /*T*/,
+                        const double* T,
                         const double* /*r*/,
                         const PhasePresence* /*cond*/,
                         double* output_mu,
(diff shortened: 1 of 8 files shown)


andlaus:
note that this PR needs OPM/opm-parser#404 to be merged first. (but OPM/opm-parser#404 can be merged without this.)

atgeirr:
Thank you for your work. After a brief overview I have some questions:
 - What is the semantics for setOilVisctTables() method (and the corresponding gas method) when it comes to the lifetime of the tables? Would it not be safer to take a copy instead of storing a pointer?
 - Why do you have separate (oil/gas) methods for setting the tables, separate data members for their pointers, and separate implementation parts in (for example) PvtConstCompr? Since all of these classes refer to a single phase only it seems more reasonable to have only a setVisctTables() method etc.
 - Why do you extract data from deck keywords repeatedly at runtime instead of doing this at construction time and storing it?
 - There seems to be a few "fill dXdT with 0" statements remaining after the field has been modified by the new code, such as line 216 of PvtConstCompr.
 - What kind of tests have you run with this code? I am thinking about verification that SPE9 still works with no performance regression etc.

andlaus:
> What is the semantics for setOilVisctTables() method (and the corresponding gas method) when it comes to the lifetime of the tables? Would it not be safer to take a copy instead of storing a pointer?

the requirement is that the the table object lives longer than the PVT object. I also considered using a  `shared_ptr` or a copy because of this, but this is not possible because the current table classes explicitly forbid copying. maybe this should be changed, but it requires @joakim-hove to be okay with it...

> Why do you have separate (oil/gas) methods for setting the tables, separate data members for their pointers, and separate implementation parts in (for example) PvtConstCompr? Since all of these classes refer to a single phase only it seems more reasonable to have only a setVisctTables() method etc.

that's because temperature dependence is handled differently by different phases. cf. OILVISCT where it depends on whether you use alive or dead oil...

(also, implementation wise, the table classes for different keywords are completely separate and they must be because of things like GASVISCT.)

> Why do you extract data from deck keywords repeatedly at runtime instead of doing this at construction time and storing it?

because the table classes already offer a nice `evaluate()` method which does the same as the linear interpolation implemented by opm-core which is not at all slower. why repeat efforts if it is not necessary at all? (I'm not opposed if you convert this yourself, but I've not enough hair left for this task. ;))

> There seems to be a few "fill dXdT with 0" statements remaining after the field has been modified by the new code, such as line 216 of PvtConstCompr.

I don't really get what you mean: the place in the code which you mention are derivatives w.r.t. pressure and R_s, which stay 0 under the assumptions of the class. the point why these derivatives are multiplied by alpha in the "new" code above is to keep more communality between the various other PVT classes. If you prefer, I will move these `fill()` statements in front of the loop which calculates the viscosity, though...

> What kind of tests have you run with this code? I am thinking about verification that SPE9 still works with no performance regression etc.

The unmodified SPE9 and SPE1 decks still run and there are no performance regressions (on my machine it was even about 2% faster with the patch, but that is due to measurement noise), and after OILVISCT and VISCREF statements were added to SPE9, it did not crash and the first timestep could be completed successfully. (this was a debug build, so it took quite a while. also, the problem is that I have no idea what "sensible" values for the *VISCT keywords are and thus cannot do a "real" test.)

atgeirr:
> the requirement is that the the table object lives longer than the PVT object. I also considered using a shared_ptr or a copy because of this, but this is not possible because the current table classes explicitly forbid copying. maybe this should be changed, but it requires @joakim-hove to be okay with it...

This makes me nervous though, as the parser may very well be destroyed before simulation starts. If a shared_ptr is possible that would be better.

atgeirr:
> that's because temperature dependence is handled differently by different phases. cf. OILVISCT where it depends on whether you use alive or dead oil...
> 
> (also, implementation wise, the table classes for different keywords are completely separate and they must be because of things like GASVISCT.)

OK, acceptable. It is annoying though that these things have small but important differences all the time (PVTO vs PVTG I am looking at you). At least the interface of these classes are a bit more uniform, so we hide this complexity.

atgeirr:
On the remaining questions: thanks for your answers, I am ok with what you've done. The reason I misread the fill issue is precisely because you multiplied it in the loop above, I'd rather see that statement go (the multiply) since it is needless and confusing to me.

andlaus:
> This makes me nervous though, as the parser may very well be destroyed before simulation starts. If a shared_ptr is possible that would be better.

currently this won't work as EclipseState returns a reference to the tables array. all things considered, I vote for the copyable table objects option. @joakim-hove is this okay with you?

andlaus:
> I'd rather see that statement go (the multiply) since it is needless and confusing to me.

ok, but be aware that this might also confuse people because for some classes there is the multiplication for others there isn't...

atgeirr:
> I vote for the copyable table objects option

So do I in this case, at least if it is fairly easy to add to opm-parser. Are the tables moveable? I think they have to be at least moveable to be put in a vector (used to require copyable before C++11).

andlaus:
> So do I in this case, at least if it is fairly easy to add to opm-parser.

yep, it is fairly easy: the main reason why I made them non-copyable in the first place was to prevent accidentially passing them by value. Probably I should implement copying using an explicit `copy()` method to keep this property...

> Are the tables moveable? I think they have to be at least moveable to be put in a vector (used to require copyable before C++11).

they are not moveable, but as far as I can remember, they are default-constructible. (which is sufficient to put them into std::vectors...)

andlaus:
all your concerns have been adressed in the current version. Note that this PR now requires OPM/opm-parser#409 to be merged...

andlaus:
ok, after the discussion it became clear that OPM/opm-parser#409 won't be merged, i.e., I reset this PR to square 2. (i.e., I removed the 'create copies of the tables' patch again.) The testing done for this PR is still unchanged...

atgeirr:
I am ready to merge this, but are there any upstream PRs I should wait for?

andlaus:
> are there any upstream PRs I should wait for? 

there are none from my side, and I also rebased it on top of the latest master this morning, so there should not be any conflicts. Be sure to also merge the opm-autodiff PR for this, though.

thanks for the review, BTW :)

andlaus:
(autodiff will still compile and run OPM/opm-autdiff#288 is not merged, but it will ignore temperature dependent viscosity in that case.)

andlaus:
(autodiff will still compile and run OPM/opm-autdiff#288 is not merged, but it will ignore temperature dependent viscosity in that case.)

PR #741, opened 2015-01-30
Fixed writing well information to the restart file

joakim-hove:
[ Observe that this requires a screaming latest version of ERT ]

The original version of the code wrote 14 items per connection to the
ICON array, however when loading restart information the ERT well loader
currently expects to find segment information in item 15. Will now write
an additional segment id item for each connection.

In addition the code has been refactored to use operator[] to set
elements instead of push_back() which was used previously.

Have tested on Norne - works on my machine!

### DIFF
--- a/opm/core/io/eclipse/EclipseWriter.cpp
+++ b/opm/core/io/eclipse/EclipseWriter.cpp
@@ -259,18 +259,24 @@ class Restart : private boost::noncopyable
 public:
     static const int NIWELZ = 11; //Number of data elements per well in IWEL array in restart file
     static const int NZWELZ = 3;  //Number of 8-character words per well in ZWEL array restart file
-    static const int NICONZ = 14; //Number of data elements per completion in ICON array restart file
+    static const int NICONZ = 15; //Number of data elements per completion in ICON array restart file
 
     /**
-     * The constants NIWELZ and NZWELZ referes to the number of elements per well that we write to
-     * the IWEL and ZWEL eclipse restart file data arrays. The constant NICONZ refers to the number of
-     * elements per completion in the eclipse restart file ICON data array.These numbers are written
-     * to the INTEHEAD header.
-     * The elements are added in the methods addRestartFileIwelData(...), addRestartFileZwelData(...)
-     * and addRestartFileIconData(...), respectively.
-     * We write as many elements that we need to be able to view the restart file in Resinsight.
-     * The restart file will not be possible to open with Eclipse, we write to little information to
-     * be  able to do this.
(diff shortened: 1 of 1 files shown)


blattms:
Somehow the switch to using ``vector::operator[]``` seems fishy to
me. AFAIK ```vector::reserve``` rely only reserves internal space and
does not alter the size of the vector. As long as you only access via
``vector::operator[]``` this does not bite you as there is no out of
bounds checking (and from the compiler's/debugger's point of view the
memory is there and access is ok). (It should throw with
```vector::at(size_t)```, though)

For clean code you should either use ```vector.resize(size_t)``` to
allocate the memory or go back to ```vector::push_back``` to insert
the values.

joakim-hove:
>Somehow the switch to using ``vector::operator[]``` seems fishy to
me.

Hmmm - ok; the vectors are created with the `std::vector<T> vector(size , value)` initializer so I would assume it was OK? https://github.com/joakim-hove/opm-core/blob/well-write-bug/opm/core/io/eclipse/EclipseWriter.cpp#L1066

atgeirr:
Assuming that the iwel_data argument indeed gets initialised from the line you pointed to (to have size NIWELZ) it looks to me to be correct but very confusing. The reserve() call is superfluous and only serves to:
 - give the vector twice the required size (assuming NIWELZ is the required size)
 - make our C++ brain alarms go off (reserve() + operator[] is a common usage error)

I would rather replace the reserve() with an assertation that the vector has the expected size.

I cannot see how the old code could have worked, it would have push_back-ed all the data behind the initially allocated parts (unless the line you pointed to that initialises the vector is also new).

joakim-hove:
Thank you for the review guys - I must have been quite low on caffeine when finishing the PR. Memory wise the situation is as follows:

__Original code__: Initialize empty vector; use `reserve`[1] and grow vector with `push_back` and `insert`.

__New code__: Initialize with full predefined size, no use of `reserve`, all assignments with `[]`.

As you have pointed out there was a suporfluos `reserve` left in the final code.

[1]: The `reserve` was not really necessary in the the original code either.

blattms:
Sorry for failing to see this. and ringing the bell. If the vector was
previously resized, everything is OK.

andlaus:
observe that `NICONZ` goes from 14 to 15 which could be the real reason why this could work after the PR...

atgeirr:
This looks reasonable, will merge if it solves the problem it was meant to. @totto82 give the word.

totto82:
Thanks. It works perfectly.

PR #742, opened 2015-02-03
Use Value<double> for Connection Factor

joakim-hove:
[ Requires: https://github.com/OPM/opm-parser/pull/410 ]

Will use Value<double> for connection factor; that facilities asking the question `hasValue()`


atgeirr:
Thanks, will merge!

PR #744, opened 2015-02-03
Rename directory opm/core/tof -> opm/core/flowdiagnostics.

atgeirr:
New name is more precise, as there are not only solvers for time-of-flight there, but also other diagnostic-type solvers, in particular the anisotropic Eikonal solver, as well as general utilities for computing diagnostic quantities.


bska:
Looks good and I've verified that everything builds and all tests run on Ubuntu 12.04 LTS, so I'm going to merge into master.  Thanks a lot.

PR #745, opened 2015-02-04
Add default producer BHP control when not given in deck.

atgeirr:
When no BHP limit is given for a producer, a BHP control is added anyway, with a limit of 1 atm.

This should solve issue OPM/opm-parser#441.

### DIFF
--- a/opm/core/wells/WellsManager.cpp
+++ b/opm/core/wells/WellsManager.cpp
@@ -601,10 +601,14 @@ namespace Opm
                                               w_);
                 }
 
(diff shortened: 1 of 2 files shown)


atgeirr:
@totto82 Perhaps you can try this small change, it should eliminate the negative BHP problem.

totto82:
Thanks. I will try it, and report back.

totto82:
This PR did not solve the negative BHP problem as WCONHIST/RESV clears the control. A fix for this is provided in OPM/opm-autodiff#294. 

I don't think this PR will change any of the benchmarks we have, still I think it implements correct and needed behaviour.

PR #746, opened 2015-02-05
Adaptive timestepping improvements. 

dr-robertk:
AdaptiveSimulatorTimer: -improvement in time step adjustment near end of time interval
                            -max time step parameter
    
PIDTimeStepControl --> TimeStepControl:
        - added simple iteration count time step control
        - bug fix in PIDAndIterationCountTimeStepControl
    
AdaptiveTimeStepping: apply the above changes.

New parameter: 
timestep.control=iterationcount (select simple iteration count time stepping)
timestep.max_timestep_in_days (max time step size in days)
timestep.control.decayrate (decrease time step if iteration count is to large, default 0.75)
timestep.control.growthrate (increase time step if iteration count is smaller than target, default 1.25)

### DIFF
--- a/opm/core/simulator/AdaptiveSimulatorTimer.cpp
+++ b/opm/core/simulator/AdaptiveSimulatorTimer.cpp
@@ -30,17 +30,18 @@
 namespace Opm
 {
     AdaptiveSimulatorTimer::
-    AdaptiveSimulatorTimer( const SimulatorTimerInterface& timer, const double lastStepTaken )
+    AdaptiveSimulatorTimer( const SimulatorTimerInterface& timer,
+                            const double lastStepTaken,
+                            const double maxTimeStep )
         : start_date_time_( timer.startDateTime() )
         , start_time_( timer.simulationTimeElapsed() )
         , total_time_( start_time_ + timer.currentStepLength() )
         , report_step_( timer.reportStepNum() )
+        , max_time_step_( maxTimeStep )
         , current_time_( start_time_ )
         , dt_( 0.0 )
         , current_step_( 0 )
         , steps_()
-        , suggestedMax_( 0.0 )
-        , suggestedAverage_( 0.0 )
     {
         // reserve memory for sub steps
         steps_.reserve( 10 );
@@ -61,31 +62,30 @@ namespace Opm
     void AdaptiveSimulatorTimer::
     provideTimeStepEstimate( const double dt_estimate )
     {
-        // store some information about the time steps suggested
-        suggestedMax_      = std::max( dt_estimate, suggestedMax_ );
-        suggestedAverage_ += dt_estimate;
-
         double remaining = (total_time_ - current_time_);
+        // apply max time step if it was set
+        dt_ = std::min( dt_estimate, max_time_step_ );
 
         if( remaining > 0 ) {
 
             // set new time step (depending on remaining time)
-            if( 1.5 * dt_estimate > remaining ) {
+            if( 1.05 * dt_ > remaining ) {
                 dt_ = remaining;
-                return ;
+                // check max time step again and use half remaining if to large
+                if( dt_ > max_time_step_ ) {
+                    dt_ = 0.5 * remaining;
+                }
+                return;
             }
 
             // check for half interval step to avoid very small step at the end
             // remaining *= 0.5;
 
-            if( 2.25 * dt_estimate > remaining ) {
+            if( 1.5 * dt_ > remaining ) {
                 dt_ = 0.5 * remaining;
-                return ;
+                return;
             }
         }
-
-        // otherwise set dt_estimate as is
-        dt_ = dt_estimate;
     }
 
     int AdaptiveSimulatorTimer::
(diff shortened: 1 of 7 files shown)


opm/core/simulator/AdaptiveSimulatorTimer.cpp
```diff
@@ -61,16 +62,14 @@ namespace Opm
...
-
         double remaining = (total_time_ - current_time_);
+        // apply max time step if it was set
+        dt_ = std::min( dt_estimate, max_time_step_ );
 
         if( remaining > 0 ) {
 
             // set new time step (depending on remaining time)
-            if( 1.5 * dt_estimate > remaining ) {
+            if( 1.05 * dt_ > remaining ) {
```
atgeirr: It seems this conditional can cause the timestep to actually become larger than `max_time_step_`, although by no more than 5%. A minor deficiency, but it should be fixed. The change should be before the next conditional to keep avoiding taking tiny steps at the end of a report interval.

opm/core/simulator/TimeStepControl.cpp
```diff
@@ -20,12 +20,66 @@
...
+                                         const double decayrate,
+                                         const double growthrate,
+                                         const bool verbose)
+        : target_iterations_( target_iterations )
+        , decayrate_( decayrate )
+        , growthrate_( growthrate )
+        , verbose_( verbose )
+    {
+        if( growthrate_ <= 0.0 || growthrate_ > 1.0 ) {
+            OPM_THROW(std::runtime_error,"SimpleIterationCountTimeStepControl: growth rate needs to be between 0 and 1, ");// << growthrate_ );
```
atgeirr: Why is growth rate limited to 1?

Also, I think the rates would be more intuitively specified as multipliers, i.e. instead of decay 0.2 and growth 0.3 you would say decay 0.8 and growth 1.3. That way the sanity test would be that decay should be below and growth above 1. Any particular reason it is done like this?

opm/core/simulator/TimeStepControl.hpp
```diff
@@ -90,7 +118,7 @@ namespace Opm
         /// \param target_iterations  number of desired iterations per time step
         /// \param tol        tolerance for the relative changes of the numerical solution to be accepted
         ///                   in one time step (default is 1e-3)
-        //  \param maxgrodth  max growth factor for new time step in relation of old time step (default = 3.0)
+        //  \param maxgrowth  max growth factor for new time step in relation of old time step (default = 3.0)
```
atgeirr: This is a multiplier, correct? That would make this different (+1) from the growth rate discussed above for no (to me) obvious reason.

opm/core/simulator/TimeStepControl.cpp
dr-robertk: I'll fix this.

opm/core/simulator/TimeStepControl.hpp
dr-robertk: Since I'll fix the one above this one remains correct.

opm/core/simulator/AdaptiveSimulatorTimer.cpp
dr-robertk: Tor Harald discussed this yesterday. This checks for what is remaining is due to the problem of matching the report step at the end of the sub steps. If you fix this than you get one time step that is 95% of remaining and one that is 5%. That is even less desirable.

atgeirr: I agree that is undesirable, and that was my point: to check for max before the next conditional. That way, if remaining is 100, dt_estimate is 96 and max is 98, we should get dt == 98 (instead of 100) at this point. Then the next check (dt*1.5 < remaining) will trigger, and give dt == 50.

dr-robertk: I don't understand. If remaining is 100 and you get dt == 98 then the next dt can only be 2, even less desirable.

atgeirr: You get dt == 98 but only until line 81, where it will be changed to 50. Next step will then likely be 50 as well.

atgeirr: Ah sorry, I did not notice the return statements! I think that the return should be removed from line 74, it should be followed by the max thresholding, and the test
```
if( 1.5 * dt_ > remaining ) {
```
be replaced by
```
if( 1.5 * dt_ > remaining && dt_ < remaining ) {
```

dr-robertk: I added a check for max time size after dt was set to remaining. If dt = remaining is larger than max time step it is cut down to 0.5 remaining. 

Also, I wouldn't be surprised if somebody complains about this drastic cut in two month from now.

atgeirr: Thanks!

I would be surprised, if someone complained that we did *not* take a longer-than-max time step, I'd expect the opposite in fact.

atgeirr:
Thanks for the patch, and for fixing all issues, while bearing over with my initial failure to understand the code structure...

PR #747, opened 2015-02-05
Makes BlackoilPropertiesFromDeck copyable.

blattms:
We need this for the parallel sim_fibo_ad_cp, where we will need to point to the same parts in two copies without duplicating data.


atgeirr:
Thanks, especially for taking the time to improve docs as you go!

PR #749, opened 2015-02-06
BUGFIX. Compute well connection factor if it set larger than zero.

totto82:
With this fix SPE1 and SPE3 runs as normal. 
The well connection factor will now also be calculated for negative well connection factor (as in SPE1 and SPE3).


totto82:
Sorry the comment is messed up. I mean compute well connection factor if is is set smaller or equal to zero

joakim-hove:
No - this will fall hard on defaulted connection factors.

totto82:
If I understand the manual, its defaulted to zero. I didn't say what it did with negative values.

GitPaean:
If I remember correctly, OPM could not handle -1 as connection factor in COMPDAT while Eclipse accepts it. I had to correct some of them manually to make the DATA file work with OPM.

atgeirr:
If Eclipse accept -1 and treats it as a defaulted value (and I think it does) the correct way to fix this is to check for negative values in the parser and return false from hasValue() in that case.

joakim-hove:
OK - as I have said many times before, getting the default handling Right &#8482; is really difficult. The balance I am trying to strike is:

1. Correctness - it should be impossible to unconciously simulate with invalid values.
2. Convenience - no need to throw before a missing value is actually used; i.e. a "hole" in the deck which is never probed should be ok.  

I am quite happy with the situation for items which either have a valid default value, or no default value at all. However cases like the connection factor where a numerical value (or range) should be interpreted as: "Do not use this literal value, but rather invoke the default machinery" is not really well handled.

The whole purpose of the little funny class `Value<T>` is to give a possibility to safe guard against the use of uninitialized values. The unconditional `getValue()` call in @totto82 PR would fail hard if the connection factor had been defaulted with `*`.

>the correct way to fix this is to check for negative values in the parser and return false from hasValue() in that case.

I agree - that has been done here: https://github.com/OPM/opm-parser/pull/415

PR #750, opened 2015-02-09
Support minpvf.

qilicun:
PR https://github.com/OPM/opm-parser/pull/416 should be merged first.


atgeirr:
This does not change behaviour, only adapt to changed API. That is fine, but we guess in the long term we would like to have different behaviour depending on this. In itself this PR is fine and can be merged when the opm-parser changes go in, @joakim-hove you have license to co-merge.

PR #751, opened 2015-02-10
AdaptiveTimeStepping: improvements.

dr-robertk:
removed initial_fraction, we use restart factor instead and one parameter less. 

Also, to suggest the time step for the next report interval we use the average of all timesteps used in this report interval. This seems to be a little bit less aggressive.


atgeirr:
Looks reasonable. Thanks!

PR #752, opened 2015-02-12
Improvements for flow diagnostic tools

atgeirr:
This collects the following:
 - a new function, `computeWellPairs()` that computes the pore volumes associated with each (injector, producer) pair,
 - the `compute_tof` example program now always computes both forward (injector) and backward (producer) diagnostics (instead of that being a user option),
 - `compute_tof` also calls `computeWellPairs()` and outputs the results, if tracer computations are asked for.

The new function does not have a unit test, but results have been found to be comparable to equivalent results from MRST.

### DIFF
--- a/examples/compute_tof.cpp
+++ b/examples/compute_tof.cpp
@@ -203,14 +203,6 @@ try
     }
     bool compute_tracer = param.getDefault("compute_tracer", false);
 
-    bool start_from_injectors = true;
-    std::string trace_start = param.getDefault<std::string>("trace_start", "Injectors");
-    if (trace_start == "Producers") {
-        start_from_injectors = false;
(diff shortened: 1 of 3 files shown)


bska:
Looks good and I've done some basic build tests without issue on the tested platforms.

On a general note I personally prefer the term "reverse" rather than "backward" time-of-flight, but I'm not going to insist on a terminology change.

atgeirr:
I chose the backward/forward pair of words instead of reverse/<nothing>, as I wanted to make it clear that you should compute and use both in most cases (i.e. the reverse t.o.f. and tracer is also of general interest). I agree it is not an essential naming thing though.

I'll merge this since otherwise we are out of sync with the already-merged update to the tutorial.

PR #755, opened 2015-02-12
Makes right hand side passed to linear solver consistent.

blattms:
Due to the size of the overlap layer and the discretization scheme
the rhs might not contain correct values for overlap cells. This
commit makes sure they are correct by an additional communication step.

### DIFF
--- a/opm/core/linalg/LinearSolverIstl.cpp
+++ b/opm/core/linalg/LinearSolverIstl.cpp
@@ -196,7 +196,7 @@ namespace Opm
         Vector b(opA.getmat().N());
         std::copy(rhs, rhs+b.size(), b.begin());
         // Make rhs consistent in the parallel case
-        comm.copyOwnerToOverlap(b,b);
+        comm.copyOwnerToAll(b,b);
         // System solution
         Vector x(opA.getmat().M());
         x = 0.0;
--- a/opm/core/linalg/LinearSolverIstl.cpp
+++ b/opm/core/linalg/LinearSolverIstl.cpp
@@ -195,6 +195,8 @@ namespace Opm
                 // System RHS
         Vector b(opA.getmat().N());
         std::copy(rhs, rhs+b.size(), b.begin());
+        // Make rhs consistent in the parallel case
+        comm.copyOwnerToOverlap(b,b);
         // System solution
         Vector x(opA.getmat().M());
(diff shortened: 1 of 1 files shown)


bska:
```
opm/core/linalg/LinearSolverIstl.cpp:199: error: 'const class Dune::Amg::SequentialInformation' has no member named 'copyOwnerToOverlap'
```
This is with GCC 4.4 targeting Dune 2.2.1 in a debug, sequential build.

In a release build targeting Dune 2.3.1 with MPI support I get
```
opm/core/linalg/LinearSolverIstl.cpp:199: error: 'const class Dune::OwnerOverlapCopyCommunication<int, int>' has no member named 'copyOwnerToOverlap'
opm/core/linalg/LinearSolverIstl.cpp:199: error: 'const class Dune::Amg::SequentialInformation' has no member named 'copyOwnerToOverlap'
```

blattms:
@bska Thanks a lot for testing. I actually experienced the same
problem and fixed it local but never commit and pushed. Sorry for
that.

Before merging I would like to fixup the last commit.

atgeirr:
To see if I understood this correctly: the rhs (residual) is computed correctly for cells we own, but not for overlap, so this communication ensures that we have the right rhs also for overlap cells. If that is why, my question is: why do we need the correct rhs for overlap cells ? We do not have the correct jacobian row for overlap cells either, why is not that also needed?

blattms:
On Fri, Feb 13, 2015 at 02:18:43AM -0800, Atgeirr Fl� Rasmussen wrote:
> To see if I understood this correctly: the rhs (residual) is computed correctly for cells we own, but not for overlap, so this communication ensures that we have the right rhs also for overlap cells. If that is why, my question is: why do we need the correct rhs for overlap cells ? We do not have the correct jacobian row for overlap cells either, why is not that also needed?
> 

Beause the degrees of freedom we own still depend on those that are in
the overlap. Or looking at the matrix: The matrix row associated with
a owber cell that we own sometimes has nonzero entries in columns that
are associated with overlap cells. Ergo to compute the values for the
owner cells we do need correct values for (at least some) overlap
cells.

bska:
If you do the fixup you wanted, then this (and the companion PR OPM/opm-autodiff#302) is ready for merging.

PR #756, opened 2015-02-16
OPM-166: Use SimulationConfig instead of Deck to get threshold pressure

chflo:
Using class SimulationConfig (from EclipseState) instead of Deck to get threshold pressure values. 
Tested on Norne.

### DIFF
--- a/opm/core/utility/thresholdPressures.hpp
+++ b/opm/core/utility/thresholdPressures.hpp
@@ -49,14 +49,14 @@ namespace Opm
     std::vector<double> thresholdPressures(EclipseStateConstPtr eclipseState, const Grid& grid)
     {
         SimulationConfigConstPtr simulationConfig = eclipseState->getSimulationConfig();
-        std::vector<double> thresholdPressureTable = simulationConfig->getThresholdPressureTable();
+        const std::vector<double>& thresholdPressureTable = simulationConfig->getThresholdPressureTable();
 
         std::vector<double> thpres_vals;
 
         if (thresholdPressureTable.size() > 0) {
 
-            std::shared_ptr<GridProperty<int>> gridProperty = eclipseState->getIntGridProperty("EQLNUM");
-            auto eqlnumData = gridProperty->getData();
+            std::shared_ptr<GridProperty<int>> eqlnum = eclipseState->getIntGridProperty("EQLNUM");
+            auto eqlnumData = eqlnum->getData();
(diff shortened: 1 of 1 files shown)


opm/core/utility/thresholdPressures.hpp
```diff
@@ -38,70 +42,26 @@ namespace Opm
...
-                    if (item->getString(0) == "THPRES") {
-                        has_thpres_option = true;
-                    } else if (item->getString(0) == "IRREVERS") {
-                        OPM_THROW(std::runtime_error, "Cannot use IRREVERS version of THPRES option, not implemented.");
-                    }
-                }
-            }
-        }
+        SimulationConfigConstPtr simulationConfig = eclipseState->getSimulationConfig();
+        std::vector<double> thresholdPressureTable = simulationConfig->getThresholdPressureTable();
```
atgeirr: This should be a const reference (no need to make a copy).

opm/core/utility/thresholdPressures.hpp
```diff
@@ -38,70 +42,26 @@ namespace Opm
...
-                    if (r1 >= max_eqlnum || r2 >= max_eqlnum) {
-                        OPM_THROW(std::runtime_error, "Too high region numbers in THPRES keyword.");
-                    }
-                    thp_table[r1 + max_eqlnum*r2] = p;
-                    thp_table[r2 + max_eqlnum*r1] = p;
-                } else {
-                    OPM_THROW(std::runtime_error, "Missing data for use of the THPRES keyword.");
-                }
-            }
+            std::shared_ptr<GridProperty<int>> gridProperty = eclipseState->getIntGridProperty("EQLNUM");
```
atgeirr: I think `gridProperty` is a very bad name for the variable, it tells us nothing other than its class (and Hungarian notation is not to be encouraged), use `eclnum`, `eclnum_prop` or something like that.

atgeirr:
This PR removes lots of error checking related to the existence of various keywords etc.: are those checks now part of opm-parser? Also: has this PR been tested with decks that do *not* have THPRES, such as the SPE cases?

chflo:
The checks removed are now part of the opm-parser. Also, the PR is tested with decks that do not have THPRES, see 
https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/SimulationConfig/tests/ThresholdPressureTest.cpp

atgeirr:
Thanks for rapid response, I will merge this.

PR #758, opened 2015-02-17
Fix warnings (and small API change)

atgeirr:
This fixes warnings of various kinds, and removes an unneeded argument from the WellsManager constructor.

There are downstream consequences in opm-autodiff and opm-polymer, so companion PRs in those modules must be merged simultaneously with this one.


atgeirr:
I will merge this today unless anyone objects.

PR #760, opened 2015-02-19
Prevents unsigned-signed-comparison warnings for container with signed size

blattms:
Well, you never know. There are containers that use a signed integer
for storing its size. This results in a warning about comparing signed with
unsigned integers. This commit prevents this by explicitly casting the size
to std::size_t.


atgeirr:
That is correct, I think for example Eigen uses signed ints. I support that in principle, as one often would like to use the sizes in contexts where it is natural to subtract.

PR #761, opened 2015-02-20
Ported initStateEquil to using the GridHelpers.

blattms:
Currently the keyword EQUIL is not supported by the fully implicit blackoil simulator when using CpGrid. This commit is a first step towards this as it makes the implementation generic.

### DIFF
--- a/opm/core/simulator/initStateEquil_impl.hpp
+++ b/opm/core/simulator/initStateEquil_impl.hpp
@@ -541,10 +541,8 @@ namespace Opm
                 // imposes the requirement that cell centroids are all
                 // within this vertical span.  That requirement is not
                 // checked.
-                typename UgGridHelpers::Cell2FacesTraits<Grid>::Type
-                    cell2Faces = UgGridHelpers::cell2Faces(G);
-                typename UgGridHelpers::Face2VerticesTraits<Grid>::Type
-                    faceVertices = UgGridHelpers::face2Vertices(G);
+                auto cell2Faces = UgGridHelpers::cell2Faces(G);
+                auto faceVertices = UgGridHelpers::face2Vertices(G);
 
                 for (typename CellRange::const_iterator
                          ci = cells.begin(), ce = cells.end();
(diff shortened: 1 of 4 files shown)


atgeirr:
Thanks for the contribution, I will review it now. We are definitely going to merge this.

opm/core/simulator/initStateEquil_impl.hpp
```diff
@@ -541,21 +541,24 @@ namespace Opm
                 // imposes the requirement that cell centroids are all
                 // within this vertical span.  That requirement is not
                 // checked.
+                typename UgGridHelpers::Cell2FacesTraits<Grid>::Type
+                    cell2Faces = UgGridHelpers::cell2Faces(G);
```
atgeirr: Just a thought for the future (no change required here): would not auto save us a bit of complexity here? Even to the point of doing away with the Traits?

opm/core/simulator/initStateEquil_impl.hpp
```diff
@@ -541,21 +541,24 @@ namespace Opm
...
-                         fi != fe; ++fi)
+                    for (auto fi=cell2Faces[*ci].begin(),
+                              fe=cell2Faces[*ci].end();
+                         fi != fe;
+                         ++fi)
                     {
-                        for (const int
-                                 *i = & fn[ fnp[*fi + 0] ],
-                                 *e = & fn[ fnp[*fi + 1] ];
+                        for (auto i = faceVertices[*fi].begin(), e=faceVertices[*fi].end();
```
atgeirr: Spaces around assignments are mandatory. (Also recommended for other binary ops.)

atgeirr:
Aside from my usual nitpicking (and that was not much in this PR) this looks fine. Will merge when updated.

blattms:
I have put space around the operator and used auto (still not used to
doing that, yet).

atgeirr:
Thanks, merging!

PR #764, opened 2015-03-02
Moves all functions that do not depend on Eigen to Opm::UgGridHelpers.  [1/3]

blattms:
This PR moves all the grid helper functions that do not depend on Eigen to the module where the
grid is declared and implement.

This is the first out of three PRs that should be merged simultaneously.


bska:

Note: This probably requires a complete rebuild of the entire system.

PR #765, opened 2015-03-02
Implement temperature dependent density

andlaus:
this is the stuff which is required to implement a "pseudo thermal" black-oil simulator where the densities and viscosities are temperature dependent but energy is not conserved. This PR depends on OPM/opm-parser#427 to compile. for the code to have any effect the "plumbing" PR for opm-autodiff is required. (but autodiff compiles even without the latter.)

for what is worth this PR does not introduce any new ctest failures on my machine...

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -356,6 +356,7 @@ list (APPEND PUBLIC_HEADER_FILES
 	opm/core/props/pvt/PvtInterface.hpp
 	opm/core/props/pvt/PvtLiveGas.hpp
 	opm/core/props/pvt/PvtLiveOil.hpp
+	opm/core/props/pvt/ThermalWaterPvtWrapper.hpp
 	opm/core/props/rock/RockBasic.hpp
 	opm/core/props/rock/RockCompressibility.hpp
 	opm/core/props/rock/RockFromDeck.hpp
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -357,6 +357,7 @@ list (APPEND PUBLIC_HEADER_FILES
 	opm/core/props/pvt/PvtLiveGas.hpp
 	opm/core/props/pvt/PvtLiveOil.hpp
 	opm/core/props/pvt/ThermalWaterPvtWrapper.hpp
+	opm/core/props/pvt/ThermalOilPvtWrapper.hpp
 	opm/core/props/rock/RockBasic.hpp
 	opm/core/props/rock/RockCompressibility.hpp
 	opm/core/props/rock/RockFromDeck.hpp
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -358,6 +358,7 @@ list (APPEND PUBLIC_HEADER_FILES
 	opm/core/props/pvt/PvtLiveOil.hpp
 	opm/core/props/pvt/ThermalWaterPvtWrapper.hpp
 	opm/core/props/pvt/ThermalOilPvtWrapper.hpp
+	opm/core/props/pvt/ThermalGasPvtWrapper.hpp
 	opm/core/props/rock/RockBasic.hpp
 	opm/core/props/rock/RockCompressibility.hpp
 	opm/core/props/rock/RockFromDeck.hpp
--- a/opm/core/props/pvt/PvtConstCompr.hpp
+++ b/opm/core/props/pvt/PvtConstCompr.hpp
@@ -23,8 +23,6 @@
 #include <opm/core/props/pvt/PvtInterface.hpp>
 #include <opm/core/utility/ErrorMacros.hpp>
 
-#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
-
 #include <vector>
 #include <algorithm>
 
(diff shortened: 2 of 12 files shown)


bska:
I have verified that this (and the companion PR OPM/opm-autodiff#324) builds and all tests run on GCC 4.4/CentOS 5.11 with the current master of opm-parser.  However, I am not able to judge the technical content of this PR so I'll defer those responsibilities to someone who is.

atgeirr:
This looks mostly good to me, but I have some comments and requests:
 - From the commit messages, it seems to me that it is not exactly well-defined what keywords we should use to implement this (E100 vs E300 etc.), but you have made some choices and those are probably fine for now.
 - You seem to have added both water and oil-specific code to PvtConstCompr. Was this unavoidable? Perhaps using a different subclass for water-with-thermal and oil-with-thermal would be better? This also applies to some extent to the other classes, even though they already are phase-specific due to different table structures etc.: would it be better to make the thermal version a subclass?
 - There seems to be a substantial amount of repeated code. This of course only replicates the unfortunate pattern already present in the underlying classes, where new APIs have been added and functions more or less copied, within and across files. Could you try to make the new code less repeating by introducing some helper functions to call instead of replicating implementation? I think we should simplify and refactor these classes soon, OPM/opm-autodiff#327 is a start from the other end.

andlaus:
> From the commit messages, it seems to me that it is not exactly well-defined what keywords we should use to implement this (E100 vs E300 etc.), but you have made some choices and those are probably fine for now.

your impression correct: I actually did not find anything comprehensive about a thermal black-oil model in the Eclipse documentation, so I just took the values of the compositional E300 keywords which seemed appropriate for the component of the phase (specified via the custom [OG]COMPIDX keywords). did I ever mention that for blackoil runs, thermal effects seem to largly be an afterthought in Eclipse?

> You seem to have added both water and oil-specific code to PvtConstCompr. Was this unavoidable? Perhaps using a different subclass for water-with-thermal and oil-with-thermal would be better? This also applies to some extent to the other classes, even though they already are phase-specific due to different table structures etc.: would it be better to make the thermal version a subclass?

in life nothing is unavoidable except death. Subclassing PvtConstCompr as you propose would certainly be possible, but I don't see how this would bring sufficient benefits as amount of copy-and-paste does not get reduced, and the code only gets distributed along even more files than it already is (making it very easy to forget adapting the thermal code if changing the isothermal one)...

what is probably a better idea than splitting these classes into thermal/non-thermal is to split `PvtConstComp` into a class for water and one for oil and to split `PvtDead` into a class for gas ond one for oil. (and BTW, `PvtDeadSpline` should be removed entirely.)

> There seems to be a substantial amount of repeated code. This of course only replicates the unfortunate pattern already present in the underlying classes, where new APIs have been added and functions more or less copied, within and across files. Could you try to make the new code less repeating by introducing some helper functions to call instead of replicating implementation? I think we should simplify and refactor these classes soon, OPM/opm-autodiff#327 is a start from the other end.

I noticed (and was quite annoyed by) this pattern, too. the problem is that there are tiny differences between these methods, e.g. for the 'B()' method, the 'alpha' scaling factor inverse to that of 'b()', some methods need to set derivatives, some don't, etc. I think this should be dealt by throwing away all but a single method per quantity in The Big Refactoring (TM). this would mean that code proposed by this PR would could pretty much stay the same for the methods which "survive"...

atgeirr:
> in life nothing is unavoidable except death. Subclassing PvtConstCompr as you propose would certainly be possible, but I don't see how this would bring sufficient benefits as amount of copy-and-paste does not get reduced, and the code only gets distributed along even more files than it already is (making it very easy to forget adapting the thermal code if changing the isothermal one)...

A better solution (perhaps): put the implementation of these thermal things (that are different by phase) in separate classes with a common interface, let PvtConstCompr and other classes hold a (smart) pointer to such a class (or null of not needed). What do you think?

And The Big Refactoring is going to happen one small step at a time...

andlaus:
> A better solution (perhaps): put the implementation of these thermal things (that are different by phase) in separate classes with a common interface, let PvtConstCompr and other classes hold a (smart) pointer to such a class (or null of not needed). What do you think?

this could be viable, but it could be that there are differences in how the phases must be handled for the different isothermal classes (I don't remember, though). I'll investigate...

> And The Big Refactoring is going to happen one small step at a time...

true. IMO, step number 1 is to throw away everything which is not used. In this context, it is probably useful to remember the following quote of Edgar Dijkstra: "If we wish to count lines of code, we should not regard them as 'lines produced' but as 'lines spent'." This simple fact seems to be ignored by many projects. We should not make this mistake.

andlaus:
okay, this PR has been updated to use wrappers around `PvtInterface` objects and has also been rebased on top of the latest master. on my machine all ctests which work on the current master also pass with this PR applied and there are no performance or other regressions for isothermal decks as far as I can tell. the same applies to the corresponding autodiff PR (OPM/opm-autodiff#324).

the results for the thermal SPE1 test problem (cf OPM/opm-data#27) are significantly different from the isothermal version, but I still have no idea if they are reasonable or not...

atgeirr:
Thanks for the refactoring. Now the code actually got longer... I think that is mostly the fault of our too-rich interface, since you have implemented the wrappers as PvtInterface subclasses. My original suggestion was actually to do it the other way around, by having the PvtLiveOil etc. classes have a smart pointer (null if isothermal) to a PvtThermalInterface (or similar), but I guess your way is just as good, at least if we manage to cut down on the redundant interfaces. I'll take a longer look at this later today or tomorrow, being pressed for time today.

PR #766, opened 2015-03-04
SimulatorReport: add total number of newton and linear iterations to report struct.

dr-robertk:
This PR adds the number of newton and linear iterations to the report structure.


bska:
Looks good.

While you're here, would you mind replacing
```C++
#include <iostream>
```
with the more targeted
```C++
#include <ostream>
```
in `SimulatorReport.cpp`.  Admittedly, this is a `.cpp` file rather than a header, but we shouldn't bring in `<iostream>` indiscriminately.

atgeirr:
Thanks, will merge.

PR #768, opened 2015-03-05
OPM 157 write RFT data

chflo:
Added writing of eclipse RFT file to EclipseWriter. The writing of RFT data is handled in a separate class, EclipseWriteRFTHandler.

### DIFF
--- a/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
+++ b/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
@@ -80,6 +80,9 @@ namespace EclipseWriterDetails {
         if (rft_nodes.size() > 0) {
             ecl_rft_file_update(filename.c_str(), rft_nodes.data(), rft_nodes.size(), ecl_unit);
         }
+
+        //Cleanup: The ecl_rft_file_update method takes care of freeing the ecl_rft_nodes that it receives.
+        //         Each ecl_rft_node is again responsible for freeing it's cells.
     }
 
 
--- a/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
+++ b/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
@@ -1,0 +1,140 @@
+/*
+  Copyright 2015 Statoil ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <vector>
+
+#include <opm/core/io/eclipse/EclipseWriteRFTHandler.hpp>
+#include <opm/core/simulator/SimulatorState.hpp>
+#include <opm/core/simulator/BlackoilState.hpp>
+#include <opm/core/simulator/SimulatorTimer.hpp>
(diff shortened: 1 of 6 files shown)


opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
```diff
@@ -0,0 +1,140 @@
...
+                                               std::vector<double>& swat,
+                                               std::vector<double>& sgas) {
+
+
+
+        std::vector<ecl_rft_node_type *> rft_nodes;
+        for (std::vector<WellConstPtr>::const_iterator ci = wells.begin(); ci != wells.end(); ++ci) {
+            WellConstPtr well = *ci;
+            if ((well->getRFTActive(simulatorTimer.currentStepNum())) || (well->getPLTActive(simulatorTimer.currentStepNum()))) {
+                ecl_rft_node_type * ecl_node = createEclRFTNode(well,
```
bska: Method `createEclRFTNode()` allocates memory for the `ecl_rft_node_type` pointer, presumably using `malloc()`.  What happens to that memory once `rft_nodes` goes out of scope?

chflo: The ERT library takes care of the cleanup of the ecl_rft_nodes that it receives in the ecl_rft_file_update method.
Cleanup of the ecl_rft_cells within each node are the responsibility of each ecl_rft_node, so this is also taken care of.

But I guess this deserves a comment in the code…I will add a comment on the cleanup of the ecl_rft_nodes

bska: > The ERT library takes care of the cleanup of the `ecl_rft_node`s that it receives in the `ecl_rft_file_update` method.

Aha.  I see.

> Cleanup of the `ecl_rft_cells` within each node are the responsibility of each `ecl_rft_node`, so this is also taken care of.

Good.

> But I guess this deserves a comment in the code…I will add a comment on the cleanup of the `ecl_rft_node`s

Yes please.  That'd be good.

bska:
I take it all requisite upstream (i.e., opm-parser) changes have already been made, or does this depend on any as-yet unmerged PRs?

chflo:
This PR is dependent on opm-parser PR's 431 and 434 which has both been merged

bska:
> This PR is dependent on opm-parser PR's [...] which have both been merged

I see.  I'll do a test build and report back.  In the meantime, would you mind putting the test helper functions (`verifyRFTFile()` and the `create*()` family) into an anonymous namespace in `tests/test_EclipseWriteRFTHandler.cpp`?  That's the C++ equivalent of C's `static` functions and eliminates build diagnostics of the form
```
/home/bska/work/opm/src/opm/opm-core/tests/test_writenumwells.cpp: In function ‘Opm::DeckConstPtr createDeck(const std::string&)’:
/home/bska/work/opm/src/opm/opm-core/tests/test_writenumwells.cpp:136: warning: no previous declaration for ‘Opm::DeckConstPtr createDeck(const std::string&)’
```

joakim-hove:
> or does this depend on any as-yet unmerged PRs?

It also depends on: https://github.com/Ensembles/ert/pull/712

Which was merged ~ two weeks ago.

bska:
I didn't see any obvious changes to the output of opm-autodiff's [`flow`](https://github.com/OPM/opm-autodiff/blob/master/examples/flow.cpp) utility on the SPE9 benchmark.  Even the file sizes are the same and the files themselves compare equal in terms of `cmp(1)`.  Do I need to set some option to enable RFT output or do we need to make additional changes, e.g., to [`class Opm::BlackoilOutputWriter`](https://github.com/OPM/opm-autodiff/blob/master/opm/autodiff/SimulatorFullyImplicitBlackoilOutput.hpp) to benefit from this work?

Note: These are just incidental comments.  I will merge the patch set irrespective of the replies.  Probably on Monday (2015-03-09).

joakim-hove:
You must ask for RFT with the RFT keyword in the SCHEDULE section. Then you
should get a CASE.RFT file with RFT output.
7. mars 2015 13:50 skrev "Bård Skaflestad" <notifications@github.com>
følgende:

bska:
> You must ask for RFT with the `WRFT` keyword in the SCHEDULE section

Aha, that helps a lot.  Thanks.  I'll merge into master.

PR #769, opened 2015-03-06
Eclipsewriter: only write summary in substeps

dr-robertk:
This PR fixes the issue of having "monstrous" amount of data produced when sub stepping is enabled. Now, only the summary is written on sub steps.

### DIFF
--- a/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
+++ b/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
@@ -105,7 +105,7 @@ namespace EclipseWriterDetails {
         ecl_rft_node_type * ecl_rft_node = ecl_rft_node_alloc_new(well_name.c_str(), type.c_str(), recording_date, days);
 
         CompletionSetConstPtr completionsSet = well->getCompletions(timestep);
-        for (int index = 0; index < completionsSet->size(); ++index) {
+        for (size_t index = 0; index < completionsSet->size(); ++index) {
             CompletionConstPtr completion = completionsSet->get(index);
             size_t i = (size_t)completion->getI();
             size_t j = (size_t)completion->getJ();
@@ -131,7 +131,7 @@ namespace EclipseWriterDetails {
 
     void EclipseWriteRFTHandler::initGlobalToActiveIndex(const int * compressedToCartesianCellIdx, size_t numCells, size_t cartesianSize) {
         globalToActiveIndex_.resize(cartesianSize, -1);
-        for (int active_index = 0; active_index < numCells; ++active_index) {
+        for (size_t active_index = 0; active_index < numCells; ++active_index) {
             //If compressedToCartesianCellIdx is NULL, assume no compressed to cartesian mapping, set global equal to active index
             int global_index = (NULL != compressedToCartesianCellIdx) ? compressedToCartesianCellIdx[active_index] : active_index;
             globalToActiveIndex_[global_index] = active_index;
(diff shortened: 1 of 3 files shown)


bska:
Can you rebase this on top of current master to remove the merge commit (f28d0b6)?

dr-robertk:
Rebased.

bska:
> Rebased.

Thanks.  Unless anyone beats me to it, I'll test this once I'm done with #768.

bska:
> I'll test this once I'm done with #768.

With the merging of #768 this can no longer be automatically merged due to relatively small conflicts being compounded by white-space adjustments through the introduction of the
```C++
 if( reportStepIdx_ != timer.reportStepNum() )
```
branch.

Would you mind rebasing this on top of current master?

dr-robertk:
H, finally. I also took the liberty to fix a couple of minor issues introduced by the 768 PR.

dr-robertk:
I just thought about it. I'm not sure what exactly was introduced by this RTF Handler. This is now only reported at report steps. In sub steps only the summary is written. Is this correct?

bska:
> I'm not sure what exactly was introduced by this RTF Handler

It's entirely possible that I misunderstand the documentation, but as far as I can tell the keywords `WRFT` and `WRFTPLT` output data only at the (SCHEDULE) point at which the keyword is entered.  In other words, to have production logging output (pressures, saturations, flow rates &c) for the well bore throughout the simulation, one would need to specify the appropriate keyword (usually `WRFTPLT`) multiple times in the SCHEDULE section.

Maybe the authors @chflo, @joakim-hove, or @iLoop2 could weigh in on the matter?

joakim-hove:
>It's entirely possible that I misunderstand the documentation, but as far as I can tell the keywords WRFT and WRFTPLT output data only at the (SCHEDULE) point at which the keyword is entered.

The `WRFT`and `WRFTPLT` keywords have quite rich behaviour, depending on the items chosen you can get:

- Output at the current timestep.
- Output for every timestep - starting now (or at a specified future timestep).
- Output at the current timestep - or alternatively when the well first opens.

All of the time related fencing is handled by the `EclipseState` object, and the RFT handler just consults a timeStep -> bool mapping to check whether an RFT should be output or not. 

>This is now only reported at report steps. In sub steps only the summary is written. Is this correct?

With a wide margin *good enough*. Depending on how you read the documentation it might be argued that RFT output should be generated for every simulator timestep when the second/third item of the `WRFTPLT` equals `TIMESTEP` - the underlying Parser++ functionality does not support that anyway.

dr-robertk:
When running Norne now, I get the following output when writing a report step: 

PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data
PLT not supported, writing RFT data

Is this intended?

chflo:
Yes, this is intended.  We do not support writing of PLT yet. Since PLT is a superset of RFT, we write RFT instead and write an error message.
In the near future, these messages will be written to a log instead.

dr-robertk:
@chflo: good to know.

dr-robertk:
@bska: I changed the default to std::invalid_argument.

dr-robertk:
Ok, now it's "perfect". It would have been easier, just to provide the method in the first place.

bska:

PR #770, opened 2015-03-06
External cmake

joakim-hove:
This PR is a proof of concept of how we can separate the build system into a seperate repository. The PR should *not* be merged as is - but hopefully it can serve as a starting point for discussions at the OPM meeting.

To try it out:

1. `git clone git@github.com:joakim-hove/opm-cmake` in $RANDOM_PATH
2. Wipe the build directory of `opm-core` and configure with: `cmake -DOPM_CMAKE_ROOT=$RANDOM_PATH/opm-cmake`

Currently the opm-cmake project is a 100% copy of the `cmake/` directory of the master branch in `opm-core` - however if we choose to proceed in this direction it will be natural to make changes/improvements to the `opm-cmake` project.


bska:
Commit  46c3c56 removes the `cmake` directory which also contains the `-prereqs.cmake` files.  Is it your intention that the `opm-cmake` project also be aware of (and maintain) all modules' prerequisite information?

joakim-hove:
>Commit 46c3c56 removes the cmake directory which also contains the -prereqs.cmake files. Is it your intention that the opm-cmake project also be aware of (and maintain) all modules' prerequisite information?

I guess that each project should maintain their own `-prereqs.cmake` files; the current PR is mainly meant as an illustration of a possible approach, in particular I tried to keep the amount of changes as low as possible.

bska:
> I guess that each project should maintain their own `-prereqs.cmake` files

That would be my preferred option as well.

> I tried to keep the amount of changes as low as possible.

Fair enough.  Let's discuss it on those grounds.

joakim-hove:
> I guess that each project should maintain their own -prereqs.cmake files

I have looked a bit more into this; unfortunately the `Findopm-xxx` module requires the `opm-xxx-prereqs.cmake` file, so without a slightly more involved refactoring we must assemble all the `xxx-prereqs.cmake` files in the `opm-cmake` repository.

bska:
> without a slightly more involved refactoring we must assemble all the `xxx-prereqs.cmake` files in the `opm-cmake` repository.

That's unfortunate because it means that `opm-cmake` would become less useful, maybe even close to useless, for modules that aren't developed by the OPM team.  In a sense this property creates a dichotomy between the insiders (members of GitHub group OPM) and independent parties.

My current view is that if the effect can be achieved at all, then the long-term goal should be for per-module specification of that module's prerequisites.

blattms:
This might be provocative, but I will mention this anyway:

I think the way Kitware would like to have the world is that every
package ships a package configuration file that has (litterally) all
the information in it to build a program using the package. That way
people can even use regular Makefiles and other buildsystem can be
used without relying on OPM's build system infrastructure.

The way this would work with OPM is that the configuration files
include all the necessary compile, and link flags of all the packages
used by the specific OPM module. (basically that is the prerequisite
stuff). With cmake one would just do a find_package(opm-module) and
use opm-module_INCLUDE_DIRS, opm-module-LIBRARIES,
opm-module-COMPILE_FLAGS, etc. (whatever the case of opm-module is).

One can even do this on the command line:

cmake \
-DCMAKE_SYSTEM_PREFIX_PATH=~/src/dune/current/dune-common/build-cmake \
--find-package -DNAME=dune-common -DCOMPILER_ID=GNU -DLANGUAGE=CXX \
-DMODE=COMPILE
for the compile options or with -DMODE=LINK for the linker
options. For autotools cmake ships cmake.m4 which makes this a bit
easier.

This approach could even be used as an addition to our normal one if
we deactivate config-mode in the find_package calls under
cmake/Modules.

joakim-hove:
>That's unfortunate because it means that opm-cmake would become less useful, maybe even close to useless, for modules that aren't developed by the OPM team.

Well - I agree that having the `xxx-prereqs.cmake` files in a common repository is unfortunate, however I do feel you are exaggerating the matter slightly here. As I see it assembling the build system in one common location is an organisational prerequisite for improvements for improvements of any kind, however I strongly prefer this approach:

1. Assemble the current build system ~unmodified in `opm-cmake`repository.
2. Improve the build system in various ways.

joakim-hove:
> This might be provocative, but I will mention this anyway:

Well - to be honest my CMake skills are so rudimentary that I do not undertand whether I was awarded a slap in the face, or a pat on the back, after reading your message :-)

The main point which I feel rather strong about is to remove the massive duplication inherent in the current build system, and also that this de-duplication should be completed before we improve/change the build system.

andlaus:
> That's unfortunate because it means that opm-cmake would become less useful, maybe even close to useless, for modules that aren't developed by the OPM team. In a sense this property creates a dichotomy between the insiders (members of GitHub group OPM) and independent parties.
>
> My current view is that if the effect can be achieved at all, then the long-term goal should be for per-module specification of that module's prerequisites.

I agree with all of this, but you  need to also consider that it makes the life of "insiders" easier because the N^2 problem goes away.

on a more general note when it comes to fixing the build system: is there a good argument why the Dune cmake build system cannot be used? I haven't investigated it in-depth, but it seems to already have solved the issues mentioned by @bska. While I don't expect it to be free of ideosyncracies, at least this move should take most of the maintainance burden from us. (and it is my impression that nobody in OPM really likes to work on be build system...)

joakim-hove:
OK; similar PR's made in all modules.

rolk:
> The way this would work with OPM

The way this _already_ works in OPM is that `opm-xxx-config.cmake` (for CMake) and `opm-xxx.pc` (for pkgconfig) is dumped in the build directory. The code in OpmFind.cmake:103--116 exists to pick up these files.

PR #771, opened 2015-03-24
Fix property method calls.

atgeirr:
Cell argument can not be null for BlackoilPropertiesFromDeck class. This fixes a usage in the non-equilibrium initialisation code. Also, it makes it possible to initialise from a deck without RS or RV.


bska:

PR #772, opened 2015-03-26
Use deck units for summary output

andlaus:
And also fix up the start time used for summary output.

Note that this required a small `EclipseWriter` API change, so this PR must be merged synchronously with the corresponding opm-autodiff and opm-polymer PRs...

### DIFF
--- a/opm/core/io/OutputWriter.cpp
+++ b/opm/core/io/OutputWriter.cpp
@@ -47,10 +47,12 @@ private:
 /// Psuedo-constructor, can appear in template
 template <typename Format> unique_ptr <OutputWriter>
 create (const ParameterGroup& params,
+        std::shared_ptr <const Deck> deck,
         std::shared_ptr <const EclipseState> eclipseState,
         const Opm::PhaseUsage &phaseUsage,
         std::shared_ptr <const UnstructuredGrid> grid) {
     return unique_ptr <OutputWriter> (new Format (params,
+                                                  deck,
                                                   eclipseState,
                                                   phaseUsage,
                                                   grid->number_of_cells,
@@ -65,6 +67,7 @@ create (const ParameterGroup& params,
 /// to the list below!
 typedef map <const char*, unique_ptr <OutputWriter> (*)(
         const ParameterGroup&,
+        std::shared_ptr <const Deck> deck,
         std::shared_ptr <const EclipseState> eclipseState,
         const Opm::PhaseUsage &phaseUsage,
         std::shared_ptr <const UnstructuredGrid>)> map_t;
@@ -76,6 +79,7 @@ map_t FORMATS = {
 
 unique_ptr <OutputWriter>
 OutputWriter::create (const ParameterGroup& params,
+                      std::shared_ptr <const Deck> deck,
                       std::shared_ptr <const EclipseState> eclipseState,
                       const Opm::PhaseUsage &phaseUsage,
                       std::shared_ptr <const UnstructuredGrid> grid) {
@@ -93,7 +97,7 @@ OutputWriter::create (const ParameterGroup& params,
         // invoke the constructor for the type if we found the keyword
         // and put the pointer to this writer onto the list
         if (params.getDefault <bool> (name, false)) {
-            list->push_front (it->second (params, eclipseState, phaseUsage, grid));
+            list->push_front (it->second (params, deck, eclipseState, phaseUsage, grid));
         }
     }
 
--- a/opm/core/io/OutputWriter.hpp
+++ b/opm/core/io/OutputWriter.hpp
@@ -21,6 +21,7 @@
 #define OPM_OUTPUT_WRITER_HPP
 
 #include <memory>  // unique_ptr, shared_ptr
+#include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/core/simulator/SimulatorTimerInterface.hpp>
 
 struct UnstructuredGrid;
@@ -107,6 +108,7 @@ public:
      */
     static std::unique_ptr <OutputWriter>
     create (const parameter::ParameterGroup& params,
+            std::shared_ptr <const Deck> deck,
             std::shared_ptr <const EclipseState> eclipseState,
             const Opm::PhaseUsage &phaseUsage,
             std::shared_ptr <const UnstructuredGrid> grid);
--- a/opm/core/io/eclipse/EclipseWriter.cpp
+++ b/opm/core/io/eclipse/EclipseWriter.cpp
@@ -443,7 +443,8 @@ public:
     // on the classes defined in the following.
 
     // add rate variables for each of the well in the input file
-    void addAllWells(Opm::EclipseStateConstPtr eclipseState,
+    void addAllWells(Opm::DeckConstPtr deck,
+                     Opm::EclipseStateConstPtr eclipseState,
                      const PhaseUsage& uses);
     void writeTimeStep(int writeStepIdx,
                        const SimulatorTimerInterface& timer,
@@ -612,6 +613,8 @@ public:
     { return ertHandle_; }
 
 protected:
+
+
     void updateTimeStepWellIndex_(const std::map<std::string, int>& nameToIdxMap)
     {
         const std::string& wellName = well_->name();
@@ -724,7 +727,8 @@ public:
              Opm::WellConstPtr well,
              PhaseUsage uses,
              BlackoilPhases::PhaseIndex phase,
-             WellType type)
+             WellType type,
+             bool useFieldUnits)
         : WellReport(summary,
                      eclipseState,
                      well,
@@ -732,8 +736,9 @@ public:
                      phase,
                      type,
                      'R',
-                     "SM3/DAY" /* surf. cub. m. per day */)
-    { }
+                     handleUnit_(phase, useFieldUnits))
+    {
+    }
 
     virtual double retrieveValue(const int /* writeStepIdx */,
                                  const SimulatorTimerInterface& timer,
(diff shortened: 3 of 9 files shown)


atgeirr:
Thanks for the patch, it will eventually go in. I do have one question I'd like answered: would it not be better to avoid the
```
bool useFieldUnits = !deck->hasKeyword("METRIC");
```
in this class and rather put it (or an enum version that covers the other unit variants too) in the EclipseState? I think that would be a more proper place for it.

andlaus:
I'm not sure if EclipseState is the place for such low-level stuff, but if @joakim-hove is okay with it, I'll implement it. Regardless of that, I'd prefer if the deck will continued to be passed to the output writers because you never know what it will be needed for...

atgeirr:
> I'd prefer if the deck will continued to be passed to the output writers because you never know what it will be needed for...

I think that @joakim-hove may disagree with that as a principle, but I am fine with it for now.

atgeirr:
The code looks ok, and I have verified that it works as advertised on SPE9, which uses FIELD units. Will merge. Thanks!

joakim-hove:
I guess this works, and I understand that there was a desire to get this functionality in before the release, but I am still *very* disappointed with the implementation which lets the use of `Deck` proliferate - at the expense of the higher level `EclipseState` class. 

The `Deck` instance is a verbatim internalization of the keywords in the input deck, with no keyword-keyword interactions taken into account. By using the `Deck` instance directly we purposely spread Eclipse logic throughout the higher level modules like opm-core and opm-autodiff - I consider that unfortunate. The following (admittedly *quite contrived*) deck input will fail in the current example:

```txt
FIELD -- Use field units.

METRIC -- No - use metric units after all.

FIELD -- Eventually - we decided to use FIELD.
```
If we instead used the `EclipseState` to keep track of the unit system used in the input deck, it would not be necessary to duplicate the ordering checks throughout the code[1]. Furthermore this will obviously fail spectacularly when we want to implement support for Lab scale[2] - why embed such a booby trap in the first place?

> @andlaus: I'm not sure if EclipseState is the place for such low-level stuff

Well - now the low level stuff has escaped from opm-parser/EclipseState and appeared in opm-core, `EclipseState` is certainly the right place for this kind of code - and guess what - it is already there[3]:

```C++
EclipsteState state(deck);
auto unitSystem = state.getDeckUnitSystem();
auto unitType = unitSystem->getType();
```
After this the `unitType`variable will be an enum: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Units/UnitSystem.hpp#L33

>I think that @joakim-hove may disagree with that as a principle, but I am fine with it for now.

As already mentioned I disagree strongly with the "side-effects" of this PR. Not only because I am a jerk with principles, but because I think higher level code like opm-core and opm-autodiff should as much as possible utilize the `EclipseState` object where the idiosyncrasies of the Eclipse input format have been normalized away. If you disagree with my views of `EclipseState` versus `Deck` that is fair enough, and we should discuss that - but I really think this was merged prematurely.


>[...] I'd prefer if the deck will continued to be passed to the output writers because you never know what it will be needed for...

You know very well my opinion of `EclipseState` versus `Deck`; I realize there might be situations where low level `Deck` access is required, but those should well thought out exceptions, and this is certainly not one of them. We do not move the code in the wrong direction - just because *we never know what it will be needed for...*


image


[1]: The current `EclipseState` code would also fail on this contrived example, but that is only one place to fix it.

[2]: It is true that opm-parser is also lacking support for Lab scale - but it is quite easily added, and ref. discussion in Trondheim earlier this month opm based simulations on core plugs is certainly a viable task.

[3]. OK - I am the first to admit that the code pointed to in the example is not very clear; improvements are welcome.

atgeirr:
> As already mentioned I disagree strongly with the "side-effects" of this PR. Not only because I am a jerk with principles, but because I think higher level code like opm-core and opm-autodiff should as much as possible utilize the EclipseState object where the idiosyncrasies of the Eclipse input format have been normalized away. If you disagree with my views of EclipseState versus Deck that is fair enough, and we should discuss that - but I really think this was merged prematurely.

I was not 100%  sure if you thought the unit issue had a place in EclipseState or not (now I know...). As can be seen from my first comment on this PR I think so as well. However, I thought this issue was important enough to get a fix in before we release that I accepted this fix. I was not aware that the required APIs were already in EclipseState (I did not really look hard for it though), and assumed real extra work in opm-parser would be required.

Obviously, we'll change this to the correct, not-passing-the-deck implementation, now that we know the right API exists in EclipseState.

andlaus:
> The following (admittedly quite contrived) deck input will fail in the current example:

I'm pretty sure that the low-level parser code will also fail at this ;)

(to be sure: It should fail, and it should do so loudly. About the last point I'm not so certain...)

> Well - now the low level stuff has escaped from opm-parser/EclipseState and appeared in opm-core, EclipseState is certainly the right place for this kind of code - and guess what - it is already there[3]:

I'm not opposed to doing this using EclipseState, but I think until the deck object can be replaced entirely by EclipseState everywhere, both objects should always be available. (they already are available in most classes and IMO that's not too problematic because most of the opm-{core,autodiff,porsol,...} modules are so ECL specific that using the deck object directly does not matter too much anymore.)

> I think higher level code like opm-core and opm-autodiff should as much as possible utilize the EclipseState object where the idiosyncrasies of the Eclipse input format have been normalized away.

I don't like this input format either, but anyone who pretends that most of the code in the opm-{core,autodif,...} modules was not designed around these ideosyncrasies from day one is either not knowing what she talks about or flat-out lying ;)

Anyway, tomorrow I'll do a PR which changes the determination of the unit system to the approach which you outlined above.

> We do not move the code in the wrong direction - just because we never know what it will be needed for...

I guess that few of us will object to your "EclipseState only" crusade. (if you are prepared to do most of the work implied.)

atgeirr:
> I guess that few of us will object to your "EclipseState only" crusade. (if you are prepared to do most of the work implied.)

I for one do *not* object to this, assuming we are talk about as much *as possible* and no more.

For example, consider the properties (PVT etc.), that are currently implemented by classes ending in `...FromDeck`, they do currently directly access the Deck. We have the following EclipseState-using future alternatives:
 - having an API for evaluating properties in the EclipseState
 - having an API for getting tabular data from the EclipseState

I think neither alternative improves on the current situation: the first is really a step towards moving the entire simulator inside the EclipseState, while the second one does not provide a good abstract interface since we still need to separate PVDO from PVCDO for example (while it admittedly could make it a little easier to support SOF3-family saturation functions).

(End of example.)

I think that deciding where the EclipseState stops, when to use the lower-level Deck and such questions can only be done on a case-by-case basis, as I think my example above shows. For the current unit-issue, I agree that there is no reason not to use EclipseState.

atgeirr:
> Anyway, tomorrow I'll do a PR which changes the determination of the unit system to the approach which you outlined above.

Thanks for that @andlaus, I will try to get it in asap.

joakim-hove:
>I think that deciding where the EclipseState stops, when to use the lower-level Deck and such questions can only be done on a case-by-case basis, 

Good - I agree that this can probably only be handled on a case-by-case basis; my very loose idea is that information which can uniquely read from one keyword can naturally be read from the Deck, whereas information which must/can be combined from several different keywords should come from EclipseState; admittedly very much comes in the second category.

>For example, consider the properties (PVT etc.), [...]

Yes  - those are important; and in my opinion not obvious examples of this challenge.

> * having an API for evaluating properties in the EclipseState
* having an API for getting tabular data from the EclipseState

>I think neither alternative improves on the current situation: the first is really a step towards moving the entire simulator inside the EclipseState, while the second one does not provide a good abstract [...]

> the first is really a step towards moving the entire simulator inside the EclipseState

Here I disagree - as I see it this would be a step towards moving the entire *input processing* into the EclipseState - and that is actually where it should be in my opinion. 

**Disclaimer: "quite clueless about property tables"** 
Actually I am currently looking into supporting the `SOF3` keyword (https://github.com/OPM/opm-parser/pull/440); my current (limited) understanding of the situation is that relperm and capillary pressure can be entered using different families of keywords.  

In the current implementation the EclipseState contains quite simple table objects which map directly to the input keywords, and the logic of getting relperm from this-or-that keyword is handled in opm-core, or not handled at all. In my opinion the EclipseState object should handle that normalization, so that calling scope can ask: "give me the oil-water relperm", and not care which keywords (if any???) were used in the input deck. A consequence of this is that the relative importance of opm-core diminishes; opm-parser handles all input processing and another module, i.e. opm-autodiff in the case of flow, handles the actual simulation.
**/Disclaimer**

atgeirr:
I think that the saturation properties (SGOF, SWOF, now also SOF3 etc.) *may* be possible to give a reasonable abstract interface in EclipseState, since they all (to my knowledge at least) build tables for interpolation (unlike the PVT situation). However, there are three reasons it may not be a good idea:
 - that interface may turn out too complicated to be useful, especially considering the large amount of keywords and behaviours relating to end-point scaling
 - the implementation of these behaviours are so closely related to the actual keywords that keeping that knowledge in a single spot (the SaturationPropsFromDeck class) may be better than abstracting it
 - until we try to support a new input/deck format (for example for Sensor or CDG simulators) we do not really know what is a good abstraction for these properties

Right now there is not much abstraction for this: the SWOF data are passed through the EclipseState more or less as read. A good solution could be to make the new family of keywords end up creating the same (old) data in EclipseState, if possible. In this way, the calling code need not know which family of keywords was used to specify the data.

joakim-hove:
>A good solution could be to make the new family of keywords end up creating the same (old) data in EclipseState, if possible. 

A bit uncertain if I consider this a good solution?

>In this way, the calling code need not know which family of keywords was used to specify the data.

But yes - here we fully agree - that is my main point!

andlaus:
> I for one do _not_ object to this, assuming we are talk about as much as possible and no more.

Me neither and that's what I wrote, right? I suppose we have the usual problem here: everyone agrees but nobody considers it important enough to devote sufficient time to it. (that applies to me, maybe others have other priorities...)

PR #773, opened 2015-03-27
FindCXX11Features.cmake: check for the c++-2011 "final" specifier

andlaus:
this keyword is quite useful to avoid mistakes during development because the compiler will complain if it sees a method marked as 'final' which has not been declared 'virtual' in a base class...

### DIFF
--- a/cmake/Modules/FindCXX11Features.cmake
+++ b/cmake/Modules/FindCXX11Features.cmake
@@ -3,6 +3,7 @@
 #
 # Sets the follwing variable:
 #
+# HAVE_FINAL                       True if the compiler supports the "final" quantifier
 # HAVE_TYPE_TRAITS                 True if the <type_traits> header is available and implements sufficient functionality
 # HAVE_SHARED_PTR                  True if std::shared_ptr is available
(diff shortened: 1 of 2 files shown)


atgeirr:
Since opm-material is a prerequisite for opm-porsol and therefore also opm-upscaling, I do not think you can make it a required property (until we abandon gcc 4.4 at least), but if I understand this PR correctly it only adds the test and the config.h variable which should be ok.

andlaus:
> Since opm-material is a prerequisite for opm-porsol and therefore also opm-upscaling, I do not think you can make it a required property (until we abandon gcc 4.4 at least), but if I understand this PR correctly it only adds the test and the config.h variable which should be ok.

`final` is completely optional and that's actually the point of this PR, because `opm/material/OpmFinal.hpp` contains the following:

```c++
#if HAVE_FINAL
#define OPM_FINAL final
#else
#define OPM_FINAL /* nothing; compiler does not recognize "final" */
#endif
```

i.e., on compilers which do not support the `final` keyword, it also won't be used. this PR only ensures that the `OPM_FINAL` macro becomes non-trivial on compilers which support the keyword...

PR #774, opened 2015-03-30
AdaptiveTimeStepping: also catch ISTLError caused in AMG.

dr-robertk:
This happens when the time step is too large. A restart with smaller time step solves the problem. This is needed when Norne is run with AMG.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -148,6 +148,10 @@ namespace Opm {
                 std::cerr << e.what() << std::endl;
                 // also catch linear solver not converged
             }
+            catch (const Dune::ISTLError& e) {
+                std::cerr << e.what() << std::endl;
+                // also catch errors in ISTL AMG that occur when time step is too large
+            }
 
             // (linearIterations < 0 means no convergence in solver)
             if( linearIterations >= 0 )


atgeirr:
Looks fine as it just extends existing treatment of exceptions in this context. Two questions:
 - Could we catch std::exception instead since all caught exceptions derive from it? (I now notice that for some reason ISTLError does not, in fact derive from it (why?), so we cannot do this.)
 - Why do we get an exception in the first place? I would think that this could be considered regular and not exceptional behaviour (and therefore we should perhaps be using a different mechanism to handle it).

dr-robertk:
Initially I only wanted to catch NumericalProblem. Then in order to make it work, I also had to catchstd::runtime_error. I can investigate if we can remove this. Although this kind of late time change decrease the stability of the code. 
ISTLError is a different story and we should investigate what the problem is. 
So in the end we want to keep only one exception and that is NumericalProblem.

dr-robertk:
I double checked: runtime_error is thrown at two places, in the Newton and for the linear solver. 
There are various options, all of them are undesirable. 

1) Make new exception with the name NoSolverConvergence (or something like it). That and only that is caught by AdaptiveTimeStepping. 
2) Re-factor the interface of the Newton and LinearSolver such that convergence is return and the solution because a parameter. 
3) Hack around  and abuse the internally stored number of iterations (-1) to indicate convergence failure.

atgeirr:
I think this option is the best in the long run:
> 2) Re-factor the interface of the Newton and LinearSolver such that convergence is return and the solution because a parameter.

And I agree that this is unfortunate (and we do not want to mess it up now):
>  Although this kind of late time change decrease the stability of the code. 

I assume that the cause of this:
> ISTLError is a different story and we should investigate what the problem is. 

would be something we could avoid by careful programming (that is, we made a subtle mistake). If it is a mathematical impossibility, I may disagree with the choice of throwing in istl, but we'd have to just accept that for the time being.

Anyway, thanks for answering my questions, and thanks for the patch!

blattms:
@dr-robertk Please file a bug report to keep track of the bogus AMG behaviour.

PR #776, opened 2015-03-30
summary output: use the EclipseState to determine the deck unit system

andlaus:
seems like the guy who added EclipseState::getDeckUnitSystem() suffers
from dementia...

note that the ugly `handleUnit_()` method is still required as long as a correct "unit name" ought to be supported. (the conversion itself could equally well be done using opm-parser's unit handling code...)

### DIFF
--- a/opm/core/io/OutputWriter.cpp
+++ b/opm/core/io/OutputWriter.cpp
@@ -47,12 +47,10 @@ private:
 /// Psuedo-constructor, can appear in template
 template <typename Format> unique_ptr <OutputWriter>
 create (const ParameterGroup& params,
-        std::shared_ptr <const Deck> deck,
         std::shared_ptr <const EclipseState> eclipseState,
         const Opm::PhaseUsage &phaseUsage,
         std::shared_ptr <const UnstructuredGrid> grid) {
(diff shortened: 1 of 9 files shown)


atgeirr:
Sorry for being dense, but this does not remove the extra deck arguments that turned out to be unnecessary, right? I would like those removed as well...

andlaus:
> I would like those removed as well...

I'd prefer to keep it around because it will require to modify about 20 classes and three modules to remove it again. If you stick to your opinion, expect this PR to be amended this afternoon...

atgeirr:
> If you stick to your opinion, expect this PR to be amended this afternoon...

I do. If it is easier, we could reverse the PRs? Anyway thanks for rewriting!

andlaus:
> I do.

ok, I'll fix this.

> If it is easier, we could reverse the PRs? Anyway thanks for rewriting!

at least for opm-core it would not be easier since about 60% of the original patch is concerned with the actual functionality. the other two modules can simply be reversed.

andlaus:
ok, done. merge it synchronously with the PRs for autodiff and polymer, though..

joakim-hove:
Cool :-)

atgeirr:
Thanks!

PR #778, opened 2015-04-08
fix screw-up in the unit handling for the summary output

andlaus:
basically, the unit system was reversed for rates and a
UnitSystem::UnitType object was implicitly casted to bool for the
bottom hole pressure monitor. (if the BHP monitor worked, it was only
by accident...)


atgeirr:
Thanks, will merge.

PR #779, opened 2015-04-09
reduce the default optimization level from -O3 to -O2

andlaus:
-O3 is sometimes declared "unsafe"
(cf. https://wiki.gentoo.org/wiki/GCC_optimization ) and it seems like
it bit us in https://github.com/OPM/opm-material/issues/23 if GCC 4.8
was used. so let's play safe and use -O2 for now...

### DIFF
--- a/cmake/Modules/UseOptimization.cmake
+++ b/cmake/Modules/UseOptimization.cmake
@@ -38,7 +38,7 @@ if (CXX_COMPAT_GCC)
 
   # default optimization flags, if not set by user
   set_default_option (CXX _opt_dbg "-O0" "(^|\ )-O")
-  set_default_option (CXX _opt_rel "-O3" "(^|\ )-O")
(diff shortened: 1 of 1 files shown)


bska:
I have a started a complete build that does effectively this.  If that pans out (i.e., fixes OPM/opm-material#23 on Linux Mint 17.1) I'll merge this and synchronise the changes to the other repositories.

atgeirr:
Good!

> synchronise the changes to the other repositories.

It struck me that there are several recent changes to the cmake/ things that should be synchronised.

bska:
I said:
> If that [build] pans out (i.e., fixes OPM/opm-material#23 on Linux Mint 17.1) I'll merge this 

It did (after a couple of false starts), so I will merge this.

> and synchronise the changes to the other repositories.

That will have to wait until tomorrow though, because it's getting late and I have other issues to attend to.

Issue #726, opened 2015-01-20
PETSc support does not compile on Debian wheezy with g++4.4


blattms:
I did the mistake and made a clean build which detected PetSc (version 3.2) on my system. Unfortunately, the PETSc supports seems broken due to my local version and compiler:

```sh
[  2%] Building CXX object CMakeFiles/opmcore.dir/opm/core/linalg/LinearSolverPetsc.cpp.o
/usr/bin/g++-4.4   -DHAVE_CONFIG_H=1 -DDUNE_COMMON_FIELDVECTOR_SIZE_IS_METHOD=1 -g -O0 -Wall -Wextra -pipe -Wno-unknown-pragmas -std=c++0x -DNDEBUG -mtune=native -I/home/mblatt/src/dune/opm/opm-core/opm-parallel -I/home/mblatt/src/dune/opm/opm-core -I/usr/include/suitesparse -I/usr/include/superlu -I/usr/lib/openmpi/include -I/usr/lib/openmpi/include/openmpi -I/usr/include/petsc -I/home/mblatt/src/dune/3rdParty/ert/include -I/home/mblatt/src/dune/opm/dune-common -I/home/mblatt/src/dune/opm/dune-istl -I/home/mblatt/src/dune/opm/opm-parser    -o CMakeFiles/opmcore.dir/opm/core/linalg/LinearSolverPetsc.cpp.o -c /home/mblatt/src/dune/opm/opm-core/opm/core/linalg/LinearSolverPetsc.cpp
/home/mblatt/src/dune/opm/opm-core/opm/core/linalg/LinearSolverPetsc.cpp: In constructor ‘Opm::<unnamed>::KSPTypeMap::KSPTypeMap(const std::string&)’:
/home/mblatt/src/dune/opm/opm-core/opm/core/linalg/LinearSolverPetsc.cpp:41: error: ‘KSPCHEBYSHEV’ was not declared in this scope
/home/mblatt/src/dune/opm/opm-core/opm/core/linalg/LinearSolverPetsc.cpp: At global scope:
/home/mblatt/src/dune/opm/opm-core/opm/core/linalg/LinearSolverPetsc.cpp:178: warning: unused parameter ‘nonzeros’
[...]

PR #784, opened 2015-04-13
Fixex compile issues with PETSc on Debian wheezy

blattms:
This PR fixes two issues:
* Compilation with g++-4.4 because of missing conversion from const char* to char* triggered by converting a pair of const char* to a pair of std::string.
* Missing suppport for KSPCHEBYSHEV in PETSc 3.2

And thus fixes issue #726


atgeirr:
Thanks! Will merge.

PR #785, opened 2015-04-14
Correct documentation of why we use operator[] to initialize map. (Completes #784)

blattms:
This commit updates the source code comment about using operator[] to
initialize the unordered map. Thanks to Bard's persistence we found
out that the cause is not the construction of the key value of type
std::string from const char* but the mapped type being a (mutable)
char* (due to C?).

This completes the PR #784.

### DIFF
--- a/opm/core/linalg/LinearSolverPetsc.cpp
+++ b/opm/core/linalg/LinearSolverPetsc.cpp
@@ -38,8 +38,10 @@ namespace{
             : default_type_(default_type)
         {
             // g++-4.4 has problems converting const char* to char*
-            // which it thinks is needed for constructing std::string.
-            // Using operator[] circumvents this problem.
+            // The problem is caused by the mapped type being PCType
+            // which (at least in PETSc 3.2) is char* because of C
+            // (in the header there is "#define PCType character*(80)").
+            // and the KSP...  defines being const char* (because of C++).
             type_map_["richardson"] = KSPRICHARDSON;
             // Not available in PETSC 3.2 on Debian
             //type_map_["chebyshev"] = KSPCHEBYSHEV;


blattms:
See also the discussion in #726 with @bska

bska:
Thanks.  I do have one comment, not on the content of this PR but on the overall string-to-type mapping.  In light of the information that `KSPType` and `PCType` are `#define`d aliases of `const char*`, using
```C++
std::unordered_map<std::string, KSPType>
```
becomes a little troubling.  As the table in [the user manual](http://www.mcs.anl.gov/petsc/petsc-3.5/docs/manualpages/KSP/KSPType.html#KSPType) shows, we are essentially implementing the identity mapping, but more worrying is the fact that the statements
```C++
type_map_["richardson"] = KSPRICHARDSON;
// ...
```
put random pointers into the container--pointers that exist only in the body of the constructor.

We may want to alter the implementation of `KSPTypeMap` to something along the lines of
```C++
class KSPTypeMap {
public:
    explicit
    KSPTypeMap(const std::string& default_type = "gmres")
        : default_type_(default_type)
    {
        solvers_.insert(KSPRICHARDSON);
        // Not available in PETSC 3.2 on Debian
        //solvers_.insert(KSPCHEBYSHEV);
        solvers_.insert(KSPCG);
        solvers_.insert(KSPBICG);
        solvers_.insert(KSPGMRES);
        solvers_.insert(KSPFGMRES);
        solvers_.insert(KSPDGMRES);
        solvers_.insert(KSPGCR);
        solvers_.insert(KSPBCGS);
        solvers_.insert(KSPCGS);
        solvers_.insert(KSPTFQMR);
        solvers_.insert(KSPTCQMR);
        solvers_.insert(KSPCR);
        solvers_.insert(KSPPREONLY);
    }

    KSPType
    find(const std::string& type) const
    {
        PetscKrylovSolvers::const_iterator it = solvers_.find(type);

        if (it == solvers_.end()) {
            it = solvers_.find(default_type_);
        }

        if (it == type_map_.end()) {
            OPM_THROW(std::runtime_error, "Unknown KSPType: '" << type << "'");
        }

        return KSPType(type.c_str());
    }

private:
    typedef std::set<std::string> PetscKrylovSolvers;

    std::string default_type_;
    PetscKrylovSolvers solvers_;
};
```

atgeirr:
> put random pointers into the container--pointers that exist only in the body of the constructor.

Why is this? Is it the pointers to the string literals you are thinking of? If so, I would assume that their content would be copied to the std::strings used as keys?

bska:
>> put random pointers into the container--pointers that exist only in the body of the constructor.
>
> Why is this? Is it the pointers to the string literals you are thinking of?

No, I'm thinking of the `KSP*` defined constants.  `std::unordered_map<std::string, KSPType>` is just another way of spelling
```
std::unordered_map<std::string, const char*>
```
so while the **keys** are fine, the values (pointers) are not.

blattms:
@bska I do not see how this can work. What you call "random" pointers are actually globally defined pointers (C style strings). I guess these are needed by PETSc internally (for pointer equality comparison?). For your example you would need ```std::set<PCType>```. But I am no way a PETSc expert. Therefore @qilicun or @jorgekva can probably better comment on the design.

BTW: There is a much easier patch than the one that I proposed:

Instead of using ```std::unordered_map<string, PCType>``` we could have used ```std::unordered_map<string, const char*>```, left the rest like it was and hope that PETSc never switches PCType from char* to anything else.

jorgekva:
Personally I think this entire design should be removed and reworked.

But, if we for some reason should stick with this design, I think a better idea is to let this be handled through the option database rather than introduce another type-unsafe argument layer on top of the bindings.

bska:
> I do not see how this can work.

Why not?  The revised `KSPTypeMap::find()` checks if the input `type` is among the known solvers and, if so, just returns the `.c_str()` which we know will be the `KSPType`.  Otherwise we error out.

> What you call "random" pointers are actually globally defined pointers (C style strings)

They are not (globally defined).  The `KSP*` constants are at best C-style string literals but there's no reason to assume that any memory is set aside for them--let alone that you can form a unique pointer to it.  If you're **really** lucky, the compiler chooses to coalesce all string literals into read-only memory in which case you will have pointer identity, but PETSc itself uses `PetscStrcmp`--which essentially boils down to `std::strcmp(s1,s2)==0`--to query the list of registered functions (cf. `KSPRegister()` and `PCRegister()`).

Given this I'm inclined to echo @jorgekva's sentiment
> Personally I think this entire design should be removed and reworked.

blattms:
I share this opinion, as long as the code compiles with g++-4.4 on Debian (which wasn't the case with PETSc since at least 2.5 month).

atgeirr:
Is this a showstopper? Please tell me it is not so...

bska:
> Is this a showstopper?

For those who use PETSc, yes it is.  Otherwise no.  We may want to exclude PETSc support from 2015.04.

jorgekva:
PETSc support isn't ready for 2015.04. I didn't want this merged in the first place, so I vote for removal.

ymmv.

atgeirr:
> PETSc support isn't ready for 2015.04. I didn't want this merged in the first place, so I vote for removal.

We already merged some PETSc stuff in the fall, we cannot easily remove it again, just for the release. I propose the following: we leave the (incomplete, experimental) support in, but do not build binary packages with PETSc. Would that be satisfactory for all?

Also: this PR should still be merged since it is just a clarification of what's already in there. Agree?

jorgekva:
By "this" I meant the PETSc support merged in autumn, which this PR now updates. Does any code at all depend on it or use it yet?

This PR is of course fine if PETSc support stays.

atgeirr:
OK, merging this now. Feel free to submit improved versions of the PETSc support, but unless fixing critical build issues I'd rather see them after the release.

jorgekva:
I'll clean up and submit the missing details of the first iteration sometime this month.

PR #786, opened 2015-04-14
Store unhandled command-line arguments

atgeirr:
The ParameterGroup class will after this no longer ignore command-line arguments it does not handle itself, but store and make the unhandled arguments available through the unhandledArguments() method.

This should not affect existing usage. The reason for the change is a need for more command-line flexibility for applications such as flow.

### DIFF
--- a/opm/core/utility/parameters/ParameterGroup.cpp
+++ b/opm/core/utility/parameters/ParameterGroup.cpp
@@ -329,7 +329,7 @@ namespace Opm {
 	    }
 	}
 
-        const std::vector<std::string> ParameterGroup::unhandledArguments() const
+        const std::vector<std::string>& ParameterGroup::unhandledArguments() const
         {
             return unhandled_arguments_;
         }
--- a/opm/core/utility/parameters/ParameterGroup.cpp
+++ b/opm/core/utility/parameters/ParameterGroup.cpp
@@ -329,5 +329,10 @@ namespace Opm {
 	    }
 	}
 
+        const std::vector<std::string> ParameterGroup::unhandledArguments() const
+        {
+            return unhandled_arguments_;
+        }
+
     } // namespace parameter
 } // namespace Opm
--- a/opm/core/utility/parameters/ParameterGroup.hpp
+++ b/opm/core/utility/parameters/ParameterGroup.hpp
@@ -268,7 +268,7 @@ namespace Opm {
 	    void insertParameter(const std::string& name, const std::string& value);
 
             /// Unhandled arguments from command line parsing.
-            const std::vector<std::string> unhandledArguments() const;
+            const std::vector<std::string>& unhandledArguments() const;
 
 	private:
 	    typedef std::shared_ptr<ParameterMapItem> data_type;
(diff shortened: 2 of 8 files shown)


opm/core/utility/parameters/ParameterGroup.cpp
```diff
@@ -329,5 +329,10 @@ namespace Opm {
 	    }
 	}
 
+        const std::vector<std::string> ParameterGroup::unhandledArguments() const
```
bska: Did you intend to return a copy here?

opm/core/utility/parameters/ParameterGroup.hpp
```diff
@@ -276,6 +279,7 @@ namespace Opm {
 	    map_type map_;
 	    const ParameterGroup* parent_;
 	    bool output_is_enabled_;
+            std::vector<std::string> unhandled_arguments_;
```
bska: There looks to be some space-vs-tab issue here.

opm/core/utility/parameters/ParameterGroup.cpp
atgeirr: No I did not. Thanks for pointing out.

opm/core/utility/parameters/ParameterGroup.hpp
atgeirr: Yes, that file turns out to still be using tabs. Will do so as well, and leave fixing it all for later.

bska: Okay, cool.

bska:
> This should not affect existing usage

I suppose not, but on the other hand, I used to get
```
$ ./compute_eikonal_from_files x
WARNING: Ignoring file 'x' with unknown extension.
Valid filename extensions are 'xml' and 'param'.
ERROR: The group '' does not contain an element named 'grid_filename'.
```
whereas I now get
```
$ ./compute_eikonal_from_files x
ERROR: The group '' does not contain an element named 'grid_filename'.
```
In other words, we lose a little bit of diagnostic ability for faulty usage in utilities that rely on the default behaviour of the `ParameterGroup`.  On the other hand, I guess you could argue that the original diagnostic was not too informative anyway.

atgeirr:
I will change this so that the verify_syntax constructor argument will control how this is treated. If true, encountering such an unknown item will throw, and I will make it print a useful message.

A separate problem is that some programs (like compute_eikonal_from_files) pass false for that argument, when they should not. That is a separate problem though, but I'll fix it here since it is related to the behaviour of the flag.

atgeirr:
Updated to warn and throw in strict mode, use strict mode for some programs that should have used it all the time.

atgeirr:
POSIX compliant...

atgeirr:
The new multi-PR build test feature on Jenkins fails to complete building opm-autodiff, but judging from the error message I think there is a bug in the test setup rather than the PRs themselves.

bska:
> The new multi-PR build test feature on Jenkins fails to complete building opm-autodiff,

I have completed a build test locally and am currently performing a run-test of `flow`.  I have no further comments and will merge once the run-test completes (unless there are issues, of course).

bska:
> [I] will merge once the run-test completes

It is still running, but looks to be following a well-known path.  I'll merge this into master and follow up with the companion PR (OPM/opm-autodiff#356) once this change makes its way to the end of all downstream modules on CI.

PR #787, opened 2015-04-15
[cmake] Add CMake module for finding PT-Scotch in dune-cornerpoint

blattms:
There already is a module for finding Zoltan, which supports
both ParMETIS and PT-Scotch underneath. While the former is
only free for academic use the latter is true open source. Therefore
with this commit we add a module for finding PT-Scotch and add the
necessary libraries and include paths for ZOLTAN.

This PR will be needed for the new load balancing in dune-cornerpoint.


blattms:
Should I propagate this PR to the other modules, too? To all of them or just the ones that use dune-cornerpoint? Is there a script?

blattms:
Is @rolk's description at https://github.com/rolk/rolk.github.io/blob/master/_posts/2013-09-23-build-system-sync.md still up to date?

bska:
I'll take care of propagating the module to the other repositories (including opm-cmake).

blattms:
@bsak Great! Thanks a lot.

atgeirr:
I do not think this has been incorporated in any way, and since the build-system no longer resides in opm-core but in opm-cmake I will close the issue, and must ask you to open a new PR in opm-cmake. Sorry for the inconvenience. At least the "propagate build system changes" part is no longer needed!

PR #790, opened 2015-04-21
Modify default parameters to be suitable for the Norne case.

atgeirr:
I would appreciate if @dr-robertk or @totto82 could review and check this PR and OPM/opm-autodiff#362.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -35,28 +35,28 @@ namespace Opm {
 
     AdaptiveTimeStepping::AdaptiveTimeStepping( const parameter::ParameterGroup& param )
         : timeStepControl_()
-        , restart_factor_( param.getDefault("solver.restartfactor", double(0.1) ) )
+        , restart_factor_( param.getDefault("solver.restartfactor", double(0.3) ) )
         , growth_factor_( param.getDefault("solver.growthfactor", double(1.25) ) )
           // default is 1 year, convert to seconds
         , max_time_step_( unit::convert::from(param.getDefault("timestep.max_timestep_in_days", 365.0 ), unit::day) )
-        , solver_restart_max_( param.getDefault("solver.restart", int(3) ) )
-        , solver_verbose_( param.getDefault("solver.verbose", bool(false) ) )
-        , timestep_verbose_( param.getDefault("timestep.verbose", bool(false) ) )
+        , solver_restart_max_( param.getDefault("solver.restart", int(10) ) )
+        , solver_verbose_( param.getDefault("solver.verbose", bool(true) ) )
+        , timestep_verbose_( param.getDefault("timestep.verbose", bool(true) ) )
         , last_timestep_( -1.0 )
     {
         // valid are "pid" and "pid+iteration"
         std::string control = param.getDefault("timestep.control", std::string("pid+iteration") );
         // iterations is the accumulation of all linear iterations over all newton steops per time step
-        const int defaultTargetIterations = 30;
+        const int defaultTargetIterations = 8;
 
-        const double tol = param.getDefault("timestep.control.tol", double(1e-3) );
+        const double tol = param.getDefault("timestep.control.tol", double(4e-5) );
         if( control == "pid" ) {
             timeStepControl_ = TimeStepControlType( new PIDTimeStepControl( tol ) );
         }
         else if ( control == "pid+iteration" )
         {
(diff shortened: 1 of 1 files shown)


dr-robertk:
@atgeirr: Maybe we only need the parameter change in autodiff. I have to double check. When only using the parameter check in opm-autodiff Norne was running considerably faster.

atgeirr:
The setup that has been tested the most is the one with both sets of changes, unless I have messed up something. So if we decide to merge only the one in opm-autodiff we should at least rerun those tests.

dr-robertk:
I'm on it to figure out if it's worth it. I'll let you know by tomorrow. Otherwise I'll merge tomorrow.

bska:
# Execution Time Experiments #

Note: The computer system on which I conducted these experiments was not completely idle during the tests.  In other words, the results are not obtained in a fully controlled environment but are indicative nonetheless.  The preliminary conclusion is that, at least based on benchmarks SPE 1 and SPE 9, PR #790 makes no difference and that OPM/opm-autodiff#362 is detrimental to computational performance.  Whether or not to merge these then hinges entirely on whether or not they represent improvements for the Norne case.

The tables use the following legend keys
```
r=b
```
with `r` denoting repositories (`c` is opm-core, `ad` is opm-autodiff) and `b` denoting branches of that repository (`m` is master and `pr` is the pull request branch that modifies the default solver parameters).  Table values are total execution time in seconds on Linux Mint 17.1 (x86_64), with 2x6 Intel(R) Xeon(R) CPU E5-2630 v2 @ 2.60GHz (SMT) and sufficient RAM.

All of the timing results were obtained using the following command sequence
```sh
{ for i in $(seq 10); do
    ${exe} ${case} 2>&1
  done; } | tee run.log
awk '/Total time taken/{print $NF}' run.log
```
in which `${exe}` is the sequential release-build version of `flow.cpp` targeting Dune 2.3.1 and `${case}` is the name of the benchmark case file (from repository opm-data, viz [`spe1/SPE1DECK.DATA`](https://github.com/OPM/opm-data/blob/master/spe1/SPE1DECK.DATA) and [`spe9/SPE9.DATA`](https://github.com/OPM/opm-data/blob/master/spe9/SPE9.DATA)).

## SPE 1 Benchmark Case ##

c=m,ad=m|c=m,ad=pr|c=pr,ad=m|c=pr,ad=pr
------- | ---------- | ---------- | ----------
3.42251 | 7.55955 | 3.43328 | 8.48473
3.42759 | 7.25026 | 3.45201 | 8.99475
3.40756 | 8.4041 | 3.44203 | 9.67313
3.4224 | 7.89388 | 3.45833 | 8.47918
3.42526 | 5.86377 | 3.41907 | 8.18856
3.47243 | 6.20732 | 3.45938 | 8.48905
3.53012 | 7.53053 | 3.45488 | 8.38815
3.5266 | 7.25281 | 3.44243 | 8.27435
3.44589 | 7.4887 | 3.46298 | 8.56066
3.53021 | 6.94427 | 3.46285 | 9.1404

## SPE 9 Benchmark Case ##

c=m,ad=m|c=m,ad=pr|c=pr,ad=m|c=pr,ad=pr
------- | ---------- | ---------- | ----------
61.2334 | 86.8762 | 62.9717 | 113.879
60.46 | 87.6672 | 72.5066 | 111.771
60.6098 | 84.416 | 62.0095 | 114.594
61.1462 | 85.729 | 61.5114 | 117.145
61.0691 | 86.2536 | 70.9323 | 113.657
60.4907 | 87.3771 | 61.8871 | 114.05
60.5411 | 88.5942 | 62.6627 | 113.27
60.0212 | 84.7393 | 60.1303 | 113.976
60.6725 | 88.6066 | 60.3574 | 112.748
60.2238 | 87.4781 | 62.9112 | 116.178

atgeirr:
Thanks for the experimental runs. The cases with ad=m should be equvialent, since then we do not activate the adaptive timestepping and the opm-core changes are all in the AdaptiveTimestepping class.

I would expect the runs to be slower than before when using adaptive timestepping, since these decks have report steps that are short enough to converge directly. It should be possible to make quicker runs with the adaptive algorithm given some tuning and fewer report steps specified in the deck, you could try with the SPE9_DEFAULT.DATA deck which was made for that purpose.

In any case, I am not too worried about the SPE1/3/9 performance using the default parameters, as long as the cases run and we are not increasing by an order of magnitude. From my point of view, the results shown are acceptable.

atgeirr:
Have adjusted parameters, any more?

BTW, your comments were on the commit, not the PR's "Files changed" tab. That makes it a little harder to get the context for the comments: if using the Files changed view the github interface will show context in the Conversation view.

PR #793, opened 2015-04-24
fix tests

akva2:
- cannot use BOOST_CHECK_CLOSE around 0 due to a relative test
- slacken a tolerance significantly

this fixes tests on i386. i'm a bit concerned about that tolerance i had to slack quite a lot. still, it's correct to about 4 digits (it's a percentage not a factor).


akva2:
with this, the entire opm suite builds and passes tests on i386.

PR #797, opened 2015-04-28
Finddune-istl.cmake: Add UMFPACK as an optional dependency

andlaus:
this makes the cmake module usable by downstream modules which do not
explicitly specify an UMFPACK dependency.

### DIFF
--- a/cmake/Modules/Finddune-istl.cmake
+++ b/cmake/Modules/Finddune-istl.cmake
@@ -17,7 +17,8 @@ find_opm_package (
 
   # required dependencies
   "dune-common REQUIRED;
-  SuperLU
+  SuperLU;
+  SuiteSparse COMPONENTS umfpack
   "
   # header to search for
   "dune/istl/bcrsmatrix.hh"


atgeirr:
While I agree this should be fixed, I would like build-system changes to hold it a little unless critical, until the opm-cmake change has been made. @joakim-hove I leave it to you if you want to merge this now or close or postpone it if it makes your work with opm-cmake harder.

andlaus:
ok. this is also not really an urgent issue because this patch is only required for Dune 2.3 or later. Since Dune 2.2 does not support UMFPack, it be perfectly happy without the patch...

blattms:
While this patch does not hurt, it does not seem to be needed either:
Umfpack is a prerequisite of opm-core anyway.

Or am I missing something (like opm modules that do not depend on
opm-core but on dune-istl)?

joakim-hove:
Whether this patch makes sense from a technical point of view is unfortunately beyond me; however build system changes should not in any way stand in the way of merging - build system update is unfortunately not &epsilon; away.

andlaus:
> While this patch does not hurt, it does not seem to be needed either: Umfpack is a prerequisite of opm-core anyway.

this is actually the point: after #798 there are modules (i.e., eWoms) which use dune-istl but do not depend on opm-core. without this patch, this causes dune-istl to bail out if UMFPack is installed...

PR #798, opened 2015-04-28
Remove some unused code

andlaus:
This is mostly infrastructural code (Opm::Spline,
Opm::TriDiagonalMatrix, property system and the cubic polynomial
inversion code) that is only used by opm-material and eWoms. The
original intention of bringing this code into opm-core was to allow
other modules to start to use this easily. Since nothing in this
direction happened during the last one and a half years, the code only
represents baggage in the opm-core context and should thus be moved to
their consumer modules to make the life of everyone involved simpler.

also, removing this code from opm-core makes it considerably easier to refactor the dependency hierarchy of the OPM modules.


atgeirr:
> also, removing this code from opm-core makes it considerably easier to refactor the dependency hierarchy of the OPM modules.

This is a good argument, and I am therefore fine with the change. However, I do like the idea of reusing common things like this, and moving it out of opm-core means that we will risk reimplementing things instead of reusing, because it is no longer in the logical location for such. We may revisit this after reorganising the modules however.

I assume that this must be merged simultaneously with PRs in opm-material and/or ewoms, so I leave it to @andlaus to merge this when convenient.

andlaus:
> we will risk reimplementing things instead of reusing

that's true. for this reason I'd like to see a "very fundamental" module for which depending on it does not come with any complexity cost. (in particular, it should not produce a library and it should also not introduce any dependencies.) possibly, the core of the build system should also be moved there...

> I assume that this must be merged simultaneously with PRs in opm-material and/or ewoms, so I leave it to @andlaus to merge this when convenient.

no, it is orthogonal the opm-mater/ewoms changes. But since you're okay with it I'll merge it now...

PR #799, opened 2015-05-08
Adds possibility of having a well not stored on a grid partitioning.

blattms:
In a parallel run each process only knows a part of the grid. Nevertheless
it does hold the complete well information. To resolve this the WellsManager
must be able to handle this case.

With this commit its constructor gets a flag indicating whether this is
a parallel run. If it is, then it does not throw if a well has cells that
are not present on the local part of the grid. Nevertheless it will check
that either all or none of the cells of a well are stored in the local part
of the grid.

Wells with no perforated cells on the local will still be present but set to SHUT.

### DIFF
--- a/opm/core/wells/WellsManager.cpp
+++ b/opm/core/wells/WellsManager.cpp
@@ -308,13 +308,13 @@ namespace Opm
 
     /// Default constructor.
     WellsManager::WellsManager()
-        : w_(0), is_parallel_run_()
+        : w_(0), is_parallel_run_(false)
     {
     }
 
     /// Construct from existing wells object.
     WellsManager::WellsManager(struct Wells* W)
-        : w_(clone_wells(W)), is_parallel_run_()
+        : w_(clone_wells(W)), is_parallel_run_(false)
     {
     }
 
@@ -323,7 +323,7 @@ namespace Opm
                                const size_t timeStep,
                                const UnstructuredGrid& grid,
                                const double* permeability)
-        : w_(0), is_parallel_run_()
+        : w_(0), is_parallel_run_(false)
     {
         init(eclipseState, timeStep, UgGridHelpers::numCells(grid),
              UgGridHelpers::globalCell(grid), UgGridHelpers::cartDims(grid), 
--- a/opm/core/wells/WellsManager.cpp
+++ b/opm/core/wells/WellsManager.cpp
@@ -308,13 +308,13 @@ namespace Opm
 
     /// Default constructor.
     WellsManager::WellsManager()
-        : w_(0)
+        : w_(0), is_parallel_run_()
     {
     }
 
     /// Construct from existing wells object.
     WellsManager::WellsManager(struct Wells* W)
-        : w_(clone_wells(W))
+        : w_(clone_wells(W)), is_parallel_run_()
     {
     }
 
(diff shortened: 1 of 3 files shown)


opm/core/wells/WellsManager.cpp
```diff
@@ -308,13 +308,13 @@ namespace Opm
 
     /// Default constructor.
     WellsManager::WellsManager()
-        : w_(0)
+        : w_(0), is_parallel_run_()
```
bska: I can never remember the exact details so I'm always wary of default-constructed primitive types.  I think you should stick an explicit `false` in the initialisation of `is_parallel_run` just to be on the safe side.

blattms: As we requested we now false explicitly.

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -195,7 +193,38 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
...
+            }
         }
+        {   // WELSPECS handling
+            well_names_to_index[well->name()] = well_index;
+            well_names.push_back(well->name());
+            {
+                WellData wd;
+                wd.reference_bhp_depth = well->getRefDepth();
+                wd.welspecsline = -1;
+                if (well->isInjector( timeStep ))
```
atgeirr: Please use explicit braces.

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -195,7 +193,38 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
...
             }
+            if ( is_parallel_run_ )
+            {
+                // Set wells that are on other processor to SHUT.
+                std::size_t sum_completions_on_proc = std::accumulate(completion_on_proc.begin(),
+                                                                      completion_on_proc.end(),0);
+                if ( sum_completions_on_proc == 0 )
+                {
+                    const_cast<Well&>(*well).setStatus(timeStep, WellCommon::SHUT);
+                    continue;
```
atgeirr: This continue statement is so far away from the for loop statement that it is not immediately obvious which loop is continued. Would it not be better to drop that line, and make the below `if` an `else if`? I *think* that would be equivalent.

atgeirr:
Looks fine apart from minor quibbles noted.

opm/core/wells/WellsManager_impl.hpp
blattms: On Sun, May 10, 2015 at 11:37:44PM -0700, Atgeirr Flø Rasmussen wrote:
```
> >          }
> > +        {   // WELSPECS handling
> > +            well_names_to_index[well->name()] = well_index;
> > +            well_names.push_back(well->name());
> > +            {
> > +                WellData wd;
> > +                wd.reference_bhp_depth = well->getRefDepth();
> > +                wd.welspecsline = -1;
> > +                if (well->isInjector( timeStep ))
> 
> Please use explicit braces.
>
```

Sorry, but I do not understand your comment. Can you please clarify?
I did not change the code here, just the indentation level.

opm/core/wells/WellsManager_impl.hpp
blattms: You are right.
The ```continue;```is gone

opm/core/wells/WellsManager_impl.hpp
atgeirr: I meant if () { ... }, but as it was lacking braces already, we'll fix that at the next opportunity instead...

PR #800, opened 2015-05-08
 Allow to prevent printing to std::cout.

blattms:
This allows for deactivating output to std:cout. This is handy for parallel runs where we only
want to print statistics on one process.


bska:
Looks good from a cursory glance.  I'll take it for a spin later today and merge if no issues turn up.

Issue #801, opened 2015-05-08
Issue with building opm-core from source


rinauoy:
Dear Friends, 

I have been trying to build opm-core module from source. I have followed instructions and encountered the attached error. 
![screenshot from 2015-05-08 08 25 20](https://cloud.githubusercontent.com/assets/9147160/7539589/e20e6288-f55c-11e4-8a1c-c68d3c6cd7de.png)

I would really appreciate if anyone could help. 

Regards, 

Rina

bska:
Hi Rina,

From an initial glance, this **looks** like a version mismatch between the opm-core and the ERT libraries.  Did you use the latest Git sources of all components, some of them, or none of them?

Or maybe I should ask the question differently: What instructions (link) did you follow that led to the build failure?

rinauoy:
Hi Bard, 

Thanks for your reply.

I just got the opm-core-master zip today and other prerequisite components as stated in ReadMe file in opm-core-master folder. I referred the steps in ReadMe file as instructions. 

Cheers, 

Rina

bska:
Hi Rina,

Thanks for the update and sorry about the late reply. When you say

> I referred the steps in ReadMe file as instructions.

I just want to be clear that you mean the [`README`](https://github.com/OPM/opm-core/blob/master/README) in opm-core's top-level directory.

From your screenshot you appear to be running Ubuntu 14.04 LTS (Trusty Tahr), but please correct me if I'm wrong.  One of the steps in the instructions mentioned earlier is to install the ERT libraries using the command
```sh
sudo apt-get install ert.ecl
```
On Ubuntu 14.04 that will get you the 2015.04 release of the ERT libraries. I'm afraid I don't remember the exact details, but there was a recent breaking change (Ensembles/ert@613366c) in the ERT libraries that might not have made it into the 2015.04 release.  I'll ask the ERT (and OPM) developer @joakim-hove to set the records straight.

In any case, the current master sources of opm-core depend on the breaking change in ERT being in place.  If the ert.ecl package does not include that change, then at the very least the instructions in the README file are inaccurate (or, more bluntly, wrong).  If you are going to continue using the master sources of opm-core, then I recommend that you remove the pre-pacakged version of the ERT (`sudo apt-get remove ert.ecl`) and also switch to using the master (Git) [sources](https://github.com/Ensembles/ert) of the ERT.  Otherwise, you should probably switch to using the 2015.04 release branch.

I'm sorry that I can't be more helpful at the moment.

[...]

PR #802, opened 2015-05-11
Update to notify user of known problem.

atgeirr:
This minor change:
 - updates the list of components (remove deck input, add flow diagnostics),
 - updates some deps to no longer be optional,
 - warns the user of the need to use synced versions of ERT and OPM.

The last change is in response to #801.

The whole README should be checked for accuracy, especially that the install instructions work on SUSE and Red Hat, but that should not delay the current fix.

### DIFF
--- a/README
+++ b/README
@@ -118,7 +118,10 @@ sudo yum install boost-devel suitesparse-devel blas-devel lapack-devel
 sudo yum install tinyxml-devel
 sudo yum-config-manager --add-repo \
     http://www.opm-project.org/packages/current/redhat/6/opm.repo
-sudo yum install libsuperlu3 ert.ecl-devel dune-istl-devel
+sudo yum install libsuperlu3 ert.ecl-devel
+
+# optional
+sudo yum install dune-istl-devel
 
 
 DEPENDENCIES FOR MACOS X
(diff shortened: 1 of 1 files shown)


joakim-hove:
>The last change is in response to #801.

I am really sorry about that one; my motivation was to not make changes just before the release, and the breaking changes were held back to right after the release. However - the end result was that the compatibility between opm-source and ert-binary was broken very fast. That was bad judgement on my part.

atgeirr:
I agree, and have changed to your suggested wording.

Just an aside: please do not make comments in commits, but in the "Files changed" tab.

README
```diff
@@ -111,8 +116,6 @@ sudo yum install boost-devel suitesparse-devel blas-devel lapack-devel
 
 # libraries necessary for OPM
 sudo yum install tinyxml-devel
-
-# optional libraries
 sudo yum-config-manager --add-repo \
     http://www.opm-project.org/packages/current/redhat/6/opm.repo
 sudo yum install libsuperlu3 ert.ecl-devel dune-istl-devel
```
bska: I would have thought that  library dune-istl-devel would still be optional for opm-core.

atgeirr: True, but below noise level I think. Since I am not sure that the SUSE instructions are 100% correct I didn't intend to spend much effort on them now (but hopefully someone with SUSE can check them: @totto82 I seem to remember that you run that distro, is that right?).

bska: >> I would have thought that library dune-istl-devel would still be optional for opm-core.
>
> below noise level I think

I don't agree.  Elevating the ISTL from an optional to a required prerequisite is an Interface change.

PR #803, opened 2015-05-19
Allow to create the correct communication information for several unknowns.

blattms:
In this case the parallel index set might represent N entries (this might be the number of
cells of grid). Nevertheless, there are several (n) equations/unknowns attached to each index.
In this case we construct a larger index set representing N*n unknows, where each unknown
is attached to an index.

This change only affects parallel runs.

### DIFF
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -115,19 +115,20 @@ public:
     void copyValuesTo(ParallelIndexSet& indexSet, RemoteIndices& remoteIndices,
                       std::size_t local_component_size = 0, std::size_t num_components = 1) const
     {
-        ParallelIndexSet::GlobalIndex max_gi  = local_component_size;
+        ParallelIndexSet::GlobalIndex global_component_size  = local_component_size;
         if ( num_components > 1 )
         {
+            ParallelIndexSet::GlobalIndex max_gi = 0;
             // component the max global index
             for( auto i = indexSet_->begin(), end = indexSet_->end(); i != end; ++i )
             {
                 max_gi = std::max(max_gi, i->global());
             }
-            ++max_gi;
-            max_gi = communicator_.max(max_gi);
+            global_component_size = max_gi+1;
+            global_component_size = communicator_.max(global_component_size);
         }
         indexSet.beginResize();
-        IndexSetInserter<ParallelIndexSet> inserter(indexSet, max_gi,
+        IndexSetInserter<ParallelIndexSet> inserter(indexSet, global_component_size,
                                                     local_component_size, num_components);
         std::for_each(indexSet_->begin(), indexSet_->end(), inserter);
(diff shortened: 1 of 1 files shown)


atgeirr:
Looks like an important fix! Thanks for catching, I'll merge this. There are a few minor issues, I will note in the diff.

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -112,10 +112,21 @@ class ParallelISTLInformation
     /// \brief Copy the information stored to the specified objects.
     /// \param[out] indexSet The object to store the index set in.
     /// \param[out] remoteIndices The object to store the remote indices information in.
-    void copyValuesTo(ParallelIndexSet& indexSet, RemoteIndices& remoteIndices) const
+    void copyValuesTo(ParallelIndexSet& indexSet, RemoteIndices& remoteIndices,
+                      std::size_t local_component_size=0, std::size_t no_components=1) const
```
atgeirr: Use spaces around binary ops, especially assignments.

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -112,10 +112,21 @@ class ParallelISTLInformation
...
     {
+        ParallelIndexSet::GlobalIndex max_gi  = local_component_size;
+        if (no_components>1)
```
atgeirr: I was a little confused at first, as I thought this was a boolean thing (NO components...). Could you use num_components or something similar instead?

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -112,10 +112,21 @@ class ParallelISTLInformation
...
+        {
+            // component the max global index
+            for( auto i = indexSet_->begin(), end = indexSet_->end(); i != end; ++i )
```
atgeirr: Braces around loop body.

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -112,10 +112,21 @@ class ParallelISTLInformation
...
+                max_gi = std::max(max_gi, i->global());
+            ++max_gi;
```
atgeirr: I do not understand why it is incremented here?

bska: Maximum index to number of entities?

blattms: the name is misleading. First it is the maximum global index. But we are actually interested in an upper bound of the number of global components. Hence the icrement. I will add a commit that introduces an additional variable to make this clear.

blattms:
Hopefully, all the nitpicks  are now fixed :)

atgeirr:
> Hopefully, all the nitpicks are now fixed

Looks like it, thanks for your persistence!

blattms:
Thanks for the very fast merging.

PR #804, opened 2015-05-21
Added suport for converting of Metric grid values

iLoop2:
Depending on : https://github.com/OPM/opm-parser/pull/468

### DIFF
--- a/opm/core/io/eclipse/EclipseWriter.cpp
+++ b/opm/core/io/eclipse/EclipseWriter.cpp
@@ -547,8 +547,13 @@ public:
 
         eclGrid->resetACTNUM(&actnumData[0]);
 
+
         // finally, write the grid to disk
-        eclGrid->fwriteEGRID(egridFileName_.ertHandle());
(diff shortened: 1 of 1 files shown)


atgeirr:
Looks correct, but a question is: how do we (in general) handle upstream library changes (ERT) that break the API and forces updates? If lots of more things are coming in the same vein, I would prefer to merge all such PRs at the same time (and force everyone to update ERT only once) instead of one by one. Comments @iLoop2 or @joakim-hove?

joakim-hove:
>If lots of more things are coming in the same vein, I would prefer to merge all such PRs at the same time (and force everyone to update ERT only once) instead of one by one

That is a fair comment - we of course try to minimize breakages of this kind; and if we are *aware* of more problems we certainly chunk them together - but that is seldom the situation. I suggest we look into slack: https://slack.com/ as a very low overhead channel to communicate things like: "Breaking api changes - update ERT".

joakim-hove:
Merging: See: https://github.com/OPM/opm-cmake/pull/8

PR #805, opened 2015-05-22
Make adaptive time stepping parallel

blattms:
The only stage where parallelism changes the adaptive time stepping is when some inner products on the saturation andpressure are computed.

This PR makes this part parallel by adding an additonal boost::any parameter to the constuctor of the adaptive time stepping and the controller classes. Per default this is empty. In a parallel run it contains a ParallelIstlInformation object encapsulating the information about the parallelisation. This then used
to compute the parallel inner product.

There will be a small upcoming PR in opm-autodiff that uses this one.

### DIFF
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -404,6 +404,12 @@ private:
     template<typename BinaryOperator>
     struct MaskToMinOperator
     {
+        // This is a real nice one: numeric limits has to a type without const
+        // or reference. Otherwise we get complete nonesense.
+        typedef typename std::remove_reference<
+            typename std::remove_const<typename BinaryOperator::result_type>::type
+            >::type Result;
+
         MaskToMinOperator(BinaryOperator b)
         : b_(b)
         {}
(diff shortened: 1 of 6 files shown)


blattms:
BTW with this have a parallel simulation of SP1 with adaptive time stepping working. Will try SPE9 next.

atgeirr:
Looks good as far as I can tell. Merging.

opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -427,17 +471,21 @@ class ParallelISTLInformation
...
+            }
+        }
+        Result getInitialValue()
+        {
+            //g++-4.4 does not support std::numeric_limits<T>::lowest();
+            // we rely on IEE 754 for floating point values and use min()
+            // for integral types.
+            if( std::is_integral<Result>::value )
+            {
+                return -std::numeric_limits<Result>::min();
```
bska: This is undefined behaviour.  As `Result` is an integral type, you have no way of knowing if `-min()` is representable as a `Result`.

bska: Did you mean
```C++
return std::numeric_limits<Result>::min();
```
i.e., without the negative sign?

blattms: Thanks. That is exactly what it should be. Sorry! Will post another PR.

atgeirr:
Good catch Bård, I did not notice that.

PR #806, opened 2015-05-26
[bugfix,parallel] Deactivate non-existing wells in manager.

blattms:
Previously, we used the setStatus method to set wells that do not
exist on the local grid to SHUT. Or at least this is what I thought
that ```well.setStatus(timestep, SHUT)```. Unfortunately, my
assumption was wrong. This was revealed while testing a parallel run
with SPE9 that threw an expeption about "Elements must be added in
weakly increasing order" in Opm::DynamicState::add(int, T). Seems like
the method name is a bit misleading.

As it turns out the WellManager has its own complete list of active
wells (shut wells are simply left out). Therefore we can use this
behaviour to our advantage: With this commit we not only exclude shut
wells from the list, but also the ones that do not exist on the local
grid. We even get rid of an ugly const_cast.

With this PR I was able to run SPE9 in parallel.

### DIFF
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -234,7 +234,6 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
 
         well_index++;
     }
-    std::cout<<"well_index="<<well_index<<" no wells="<<wells.size()<<std::endl;
     // Set up reference depths that were defaulted. Count perfs.
 
     const int num_wells = well_data.size();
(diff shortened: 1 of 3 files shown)


opm/core/wells/WellsManager_impl.hpp
```diff
@@ -233,7 +234,7 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
 
         well_index++;
     }
-
+    std::cout<<"well_index="<<well_index<<" no wells="<<wells.size()<<std::endl;
```
atgeirr: I guess this debug output should be removed or commented out?

atgeirr:
Looks simple and good! I'll merge when the debug output has been removed.

blattms:
Debugging output removed.

PR #807, opened 2015-05-27
changed: locate opm-cmake using config mode

akva2:
- requires https://github.com/OPM/opm-cmake/pull/12 for a fully working build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,14 +18,18 @@ cmake_minimum_required (VERSION 2.8)
 
 # additional search modules
 set( OPM_CMAKE_ROOT "" CACHE PATH "Root directory containing OPM related cmake modules")
+option(SIBLING_SEARCH "Search for other modules in sibling directories?" ON)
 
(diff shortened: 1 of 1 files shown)


bska:
Do other modules need similar treatment or do they get the location of opm-cmake transitively from opm-core?

akva2:
other modules need similiar treatment, i'll get to it as i go through the modules.

bska:
> other modules need similar treatment

Right.  That's what I thought.

> i'll get to it as i go through the modules.

Thanks!

akva2:
All pulls sent. With the referenced opm-cmake pull, this and the equivalents for the other modules,  builds on my box pass both as sibling and system installs.

bska:
> @atgeirr merged commit 3f99037

Was this merged a tad too soon?  The introduction stipulates that this depends on OPM/opm-cmake#12 which is still not in opm-cmake's master branch.

atgeirr:
Oops, I interpreted the comment "all pulls sent" as an all clear.

PR #808, opened 2015-05-27
Suppress a warning in serial mode.

atgeirr:
A function argument is only used in the HAVE_MPI block.


bska:
Looks good.  I personally prefer the word "sequential" rather than "serial" here, but that's not a reason to update.  I'll merge into master.

PR #811, opened 2015-05-28
Silence multiple warnings.

atgeirr:
Also add more warnings to the disabling list of disable_warnings.h.

### DIFF
--- a/opm/core/utility/platform_dependent/disable_warnings.h
+++ b/opm/core/utility/platform_dependent/disable_warnings.h
@@ -61,6 +61,11 @@
 #pragma GCC diagnostic ignored "-Wshadow"
 #pragma GCC diagnostic ignored "-Wsign-compare"
 #pragma GCC diagnostic ignored "-Wunused-parameter"
+#pragma GCC diagnostic ignored "-Wtautological-compare"
+#pragma GCC diagnostic ignored "-Wtype-limits"
+#pragma GCC diagnostic ignored "-Wunused-function"
+#pragma GCC diagnostic ignored "-Wunneeded-internal-declaration"
+#pragma GCC diagnostic ignored "-Wunused-private-field"
 #endif // COMPATIBLE_COMPILER
 
 #endif // SILENCE_EXTERNAL_WARNINGS
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -28,11 +28,14 @@
 #include <exception>
 
 #if HAVE_MPI && HAVE_DUNE_ISTL
-#include "mpi.h"
+#include <mpi.h>
+
+#include <opm/core/utility/platform_dependent/disable_warnings.h>
 #include <dune/istl/owneroverlapcopy.hh>
 #include <dune/common/parallel/interface.hh>
 #include <dune/common/parallel/communicator.hh>
 #include <dune/common/enumset.hh>
+#include <opm/core/utility/platform_dependent/reenable_warnings.h>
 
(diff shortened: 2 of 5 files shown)


blattms:
Looks fine to me.

opm/core/utility/platform_dependent/disable_warnings.h
```diff
@@ -61,6 +61,11 @@
 #pragma GCC diagnostic ignored "-Wshadow"
 #pragma GCC diagnostic ignored "-Wsign-compare"
 #pragma GCC diagnostic ignored "-Wunused-parameter"
+#pragma GCC diagnostic ignored "-Wtautological-compare"
+#pragma GCC diagnostic ignored "-Wtype-limits"
+#pragma GCC diagnostic ignored "-Wunused-function"
```
bska: I'd prefer **not** to silence this (`-Wunused-function`).  What upstream source begets that treatment?

atgeirr: Something from dune. I do not remember right now, but if you compile without the suppression active you'll see it of course (at least if we are on the same dune, 2.3 git release branch).

I do not want any warnings from third-party code to come through, so I want this in, too. The whole point is to see the warnings we can do something about.

andlaus: have you tried to get the patches for these warnings into the Dune master? I know that this can be a painful process, but I think it is better than hacking around these warnings via pragmas...

atgeirr: We had a discussion on this a while ago, and some of the warnings would take a lot of effort to eliminate. Some of the types in the list above are only triggered in a few places, and could be fixed (and thereafter removed from our suppression list).

The basic problem however, is that every time we updated a third party library, the warnings would appear again, because they do not mandate that the code is warning-free, at least not at the warning level I compile at. So the long-term way to solve this is to convince the developers of third-party libraries (and I count dune as one in this context, even though it is more closely related) to adopt such mandates. I do not think this is feasible, I have not even been able to convince most of the OPM developers of this! Personally, I was convinced by @bska to increase my warning level some years ago, and have not regretted it.

bska: > I was convinced by @bska to increase my warning level some years ago, and have not regretted it.

I personally "just" filter away the third-party warnings before I look at them.

atgeirr:
We'll just have to agree to disagree about the pragma mechanism. Since this is a modification of existing systems and nothing new, I'll self-merge this now.

andlaus:
> We'll just have to agree to disagree about the pragma mechanism. Since this is a modification of existing systems and nothing new, I'll self-merge this now.

I guess agreeing to disagree does not prevent you to try to get real fixes into dune? (you may fail, but I'm not aware that you've even tried. Also, I suppose that at least @alfbr would also like to see this.)

alfbr:
Yes, you are right about that. If we take in a dependency, we should at least report issues upstream. On vacation right now, but I can have a look on this one when I return.

atgeirr:
> I guess agreeing to disagree does not prevent you to try to get real fixes into dune? (you may fail, but I'm not aware that you've even tried. Also, I suppose that at least @alfbr would also like to see this.)

I did notify Markus of some dune-istl warnings I felt were a bit suspicious, the others I ignored. As I discussed above I do not think it is worth the while to patch third-party libraries just to have the warnings reappear as they continue their development. Of course that means there is a risk that we ignore bugs in third-party code that we might have discovered with warnings on *if we always studied them diligently*. I am not aware of anyone that does that (for third-party code) however.

PR #812, opened 2015-06-02
Opm 206 Use IOConfig from EclipseState to decide which Eclipse output files to write (and when)

chflo:
Use the IOConfig object from EclipseState to decide:
* Whether restart file should be written for a specified report step
* whether restart files are to be unified or not (UNIFOUT keyword, default if keyword not present is multiple)
* whether an EGRID file should be written (GRIDFILE, NOGGF keywords)
* whether an INIT file should be written (INIT keyword)
* whether the above files should be formatted or not (FMTOUT keyword, default if keyword not present is unformatted)

* Removed former setting for interval writes to disk (from params) 

* Changed a test due to changes in behaviour when using IOConfig (ie if UNIFOUT keyword not set, default is multiple restart files, not unified)

### DIFF
--- a/opm/core/io/eclipse/EclipseWriter.cpp
+++ b/opm/core/io/eclipse/EclipseWriter.cpp
@@ -237,12 +237,13 @@ public:
     FileName(const std::string& outputDir,
              const std::string& baseName,
              ecl_file_enum type,
-             int writeStepIdx)
+             int writeStepIdx,
+             bool formatted)
     {
         ertHandle_ = ecl_util_alloc_filename(outputDir.c_str(),
                                              baseName.c_str(),
                                              type,
-                                             false, // formatted?
+                                             formatted,
                                              writeStepIdx);
     }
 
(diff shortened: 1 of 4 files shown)


atgeirr:
> Removed former setting for interval writes to disk (from params)

I am a little fond of this possibility, especially when writing ascii (matlab-format) output, so can I activate this option from the deck? It is better to activate it from the parameters though, so it can be done on the command line and not require deck editing. For running tests it can be a nice option to have, to reduce the amount of disk and time used for output while still having a representative number of steps to compare to a reference.

Why has it been removed?

chflo:
Hi, interval output of restart files is now supported through the keyword RPTRST, with the mnemonics BASIC and FREQ:

Documentation for Basic:

“If set to 3, restart files are created every nth report time. The
frequency of the restarts is governed by the mnemonic "FREQ=n".”

My suggestion is that we implement the interval from command line functionality by modifying the IOConfig object, in terms of BASIC=3 and FREQ = num from command line.



Christine

atgeirr:
> My suggestion is that we implement the interval from command line functionality by modifying the IOConfig object, in terms of BASIC=3 and FREQ = num from command line.

If I understand you correctly this would allow us to still give the parameter of the command line, and that the simulator class would then query the IOConfig object instead of the outputInterval_ member? If so, then I am in favour of the change, and would merge it.

chflo:
Yes, we agree☺ I will update the pull request with this change

Christine

chflo:
Override of IOConfig's restartfile interval write setting is now implemented in terms of the following PRs:
https://github.com/OPM/opm-parser/pull/495
https://github.com/OPM/opm-autodiff/pull/395

bska:
> Override of `IOConfig`'s restartfile interval write setting is now implemented in terms of [two pull requests]

Good.  But we still need this PR (i.e., #812) in order to enable that code, right?  If so, would you mind rebasing this on top of current master?  The PR cannot currently be automatically merged (no green button).

chflo:
Hi, I have now rebased this on top of current master

atgeirr:
Thanks! This looks fine to me. Has it been tested with the SPE and Norne testcases?

bska:
> This looks fine to me.

Agreed.  We will merge this once the prerequisite PRs are available in their respective master branches.

chflo:
Hi, it has now been tested on SPE1 from opm-data, this deck has RPTSCHED integer mnemonics (RESTART = 4), this generated as espected RESTART output (unified) for every report step. 
Verified that the cmd override output_Interval=0 generated no RESTART file output, and that output_interval=3 generated restart output every third timestep.

PR #813, opened 2015-06-03
Add WPIMULT keyword support

iLoop2:
This is depending on : https://github.com/OPM/opm-parser/pull/488

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ install_manifest.txt
 CTestTestfile.cmake
 DartConfiguration.tcl
 Testing/
+.DS_Store
+.idea
+build
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -170,7 +170,7 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                         pd.cell = cell;
                         {
                             const Value<double>& transmissibilityFactor = completion->getConnectionTransmissibilityFactorAsValueObject();
-
+                            const double wellPi = completion ->getWellPi();
                             if (transmissibilityFactor.hasValue()) {
                                 pd.well_index = transmissibilityFactor.getValue();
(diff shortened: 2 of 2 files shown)


bska:
> This is depending on : OPM/opm-parser#488

Just to be clear: This requires said change to opm-parser, but is not necessitated by that change.  Is that correct?  In other words, this can be merged at any time once the requisite change to opm-parser is in place?

joakim-hove:
> In other words, this can be merged at any time once the requisite change to opm-parser is in place?

That is correct.

atgeirr:
I have a little trouble understanding the naming of things: what is wellPi? Is it a multiplier? If so, it should be named to make that obvious. Also: does it always exist? The connection transmissibility factor does not always exist, and so the Value<> template is used, but that is not done for this new quantity.

(Digression: perhaps Value should be called MaybeValue<> instead? That would make it more obvious when to use it.)

joakim-hove:
> Is it a multiplier? If so, it should be named to make that obvious

Yes - it is a multiplier - and should be named accordingly.

> Also: does it always exist?

Well - the sound default value is 1.0 - so yes it always exists.

> perhaps Value should be called MaybeValue<> instead?

When we have reached sufficiently new C++ versions `Value<T>` should be implemented with `std::optional<T>`

joakim-hove:
OK - now the upstream: https://github.com/OPM/opm-parser/pull/488 has been merged.

atgeirr:
> Yes - it is a multiplier - and should be named accordingly.

I assume that you mean it will be changed upstream in opm-parser at some point? When you do, you'll need to change this file as well, for the API change, and I hope you also rename the local variables here then.

I'll merge this now.

GitPaean:
This keyword should be removed from 

https://github.com/OPM/opm-simulators/blob/6f18f7115f8a4d091ad058d24b70bfe5b454d232/opm/autodiff/MissingFeatures.cpp#L187


It also show this list is difficult to maintain..

PR #814, opened 2015-06-05
Opm 187: Added resize method to WellState

edbru:
Added resize method to WellState.hpp. 
The overall trigger for this is to be able to support the RESTART and SKIPREST keywords, 
We want to be able to populate WellState object from a RESTART file 

The intention for the resize method is to be able to implement the following: 

1) One instance of the wellstate object for a simulation (all reportsteps). Call the resize method for each reportstep.
2) With only one instance of WellState, calling resize for each reportstep, WellsManager, WellsManager.cwells, and the call to wellstate.init (for each reportstep) can be removed.

This PR depends on opm-parser PR: https://github.com/OPM/opm-parser/pull/490

### DIFF
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -22,11 +22,30 @@
 
 #include <opm/core/wells.h>
 #include <opm/core/well_controls.h>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <vector>
 #include <cassert>
+#include <utility>
+#include <iostream>
+#include <string>
 
 namespace Opm
 {
+    ///
+    /// \brief The WellIndex struct
+    ///     wellNumber
+    ///         The index of the well as it occurs in the Schedule.OpenWells object.
+    ///     wellCompletionOffset
+    ///         Absolute offset, i.e. from the start, for the completions for the well.In effect, it is the index where this
+    ///         well's completions will start. Used for the perfrates_ and perfpress_ vectors.
+    ///     completionMap
+    ///         Offset relative to wellCompletionOffset, for this completion. Used for the perfrates_ and perfpress_ vectors.
+    struct WellIndex
+    {
+        size_t wellNumber;
+        size_t wellCompletionOffset;
+        std::map<std::size_t, size_t> completionMap;
+    };
 
     /// The state of a set of wells.
     class WellState
@@ -107,6 +126,82 @@ namespace Opm
             }
         }
 
+        ///
+        /// \brief resize
+        ///     This method resizes the vectors bhp_, temperature_, wellrates_, perfrates_ and perfpress,
+        ///     to the sizes requiered for the report step that is about to be simulated.The values that
+        ///     are already in these vectors, as a result of the previous report step that was simulated,
+        ///     are repositioned to their new indices - taking into account that both wells and completions
+        ///     can be both opened and closed.
+        /// \param state
+        ///     The EclipseState object.
+        /// \param reportStep
+        ///     The number of the report step that is about to be simulated.
+        void resize(const EclipseState& state, const size_t reportStep)
+        {
+            std::vector<WellPtr> wells = state.getSchedule()->getOpenWells(reportStep);
+            const size_t numPhases = state.getNumPhases();
+            const int numCompl = getNumCompletions(wells, reportStep);
+
+            const size_t numWells = wells.size();
+
+            const std::vector<double> bhpOld = bhp_;
+            const std::vector<double> temperatureOld = temperature_;
+            const std::vector<double> wellRatesOld = wellrates_;
+            const std::vector<double> perfratesOld = perfrates_;
+            const std::vector<double> perfpressOld = perfpress_;
+
+            std::map<std::string, WellIndex> newMap = createWellIndexMap(state, wells, reportStep);
+
+            const double default_temp = 273.15 + 20.0;
+            const double default_perfpress = -1e100;
+            const double default_rates = 0.0;
+
+            bhp_.resize(numWells);
+            temperature_.resize(numWells, default_temp);
+            wellrates_.resize(numWells * numPhases, default_rates);
+            perfrates_.resize(numCompl, default_rates);
+            perfpress_.resize(numCompl, default_perfpress);
+
+            for (auto oldIter = wellIndexMap_.begin(); oldIter != wellIndexMap_.end(); ++oldIter) {
+                const std::string& wellName = (*oldIter).first;
+                auto newIter = newMap.find(wellName);
+
+                if (newIter != newMap.end()) {
+                    const WellIndex& oldWi = (*oldIter).second;
+                    const size_t oldIndex = oldWi.wellNumber;
+
+                    const WellIndex& newWi = (*newIter).second;
+                    const size_t newIndex = newWi.wellNumber;
+
+                    bhp_[newIndex] = bhpOld[oldIndex];
+                    temperature_[newIndex] = temperatureOld[oldIndex];
+
+                    for (int i = 0; i < numPhases; ++i) {
+                        wellrates_[newIndex * numPhases + i] = wellRatesOld[oldIndex * numPhases + i];
+                    }
+
+                    for (auto oldComplIter = oldWi.completionMap.begin(); oldComplIter != oldWi.completionMap.end(); ++oldComplIter) {
+                        size_t completionId = (*oldComplIter).first;
+                        auto newComplIter = newWi.completionMap.find(completionId);
+
+                        if (newComplIter != newWi.completionMap.end()) {
+                            const size_t oldComplIndex = (*oldComplIter).second;
+                            const size_t newComplIndex = (*newComplIter).second;
+
+                            const size_t oldWellCompletionOffset = oldWi.wellCompletionOffset;
+                            const size_t newWellCompletionOffset = newWi.wellCompletionOffset;
+
+                            perfrates_[newWellCompletionOffset + newComplIndex] = perfratesOld[oldWellCompletionOffset + oldComplIndex];
+                            perfpress_[newWellCompletionOffset + newComplIndex] = perfpressOld[oldWellCompletionOffset + oldComplIndex];
+                        }
+                    }
+                }
+            }
+
+            wellIndexMap_ = newMap;
+        }
+
         /// One bhp pressure per well.
         std::vector<double>& bhp() { return bhp_; }
         const std::vector<double>& bhp() const { return bhp_; }
@@ -133,6 +228,94 @@ namespace Opm
         std::vector<double> wellrates_;
         std::vector<double> perfrates_;
         std::vector<double> perfpress_;
+
+        ///
+        /// \brief wellIndexMap_
+        ///     This map is used in the resize method. It holds information between report steps.
+        std::map<std::string, WellIndex> wellIndexMap_;
+
+        int getWellMapIndex(const std::string wellName) const
+        {
+            int retVal = -1;
+            auto iter = wellIndexMap_.find(wellName);
+
+            if (iter != wellIndexMap_.end()) {
+                const WellIndex& wi = (*iter).second;
+                retVal = wi.wellNumber;
+            }
+
+            return retVal;
+        }
+
+        int getNumCompletions(const std::vector<WellPtr>& wells, const int reportStep) const
+        {
+            int numCompl = 0;
+
+            for (size_t i = 0; i < wells.size(); ++i) {
+                int numComplForWell = getNumCompletions(wells[i], reportStep);
+                numCompl += numComplForWell;
+            }
+
+            return numCompl;
+        }
+
+        int getNumCompletions(const WellPtr& well, const int reportStep) const
+        {
+            int numComplForWell = 0;
+            CompletionSetConstPtr complSet = well->getCompletions(reportStep);
+
+            for (size_t i = 0; i < complSet->size(); ++i) {
+                WellCompletion::StateEnum complState = complSet->get(i)->getState();
+
+                if (complState == WellCompletion::StateEnum::OPEN) {
+                    numComplForWell++;
+                }
+            }
+
+            return numComplForWell;
+        }   
+
+        std::map<std::string, WellIndex> createWellIndexMap(const EclipseState& state, const std::vector<WellPtr>& wells, const int reportStep)
+        {
+            EclipseGridConstPtr grid = state.getEclipseGrid();
+            size_t numWells = wells.size();
+            std::map<std::string, WellIndex> newMap;
+
+            size_t totalWellCompletionOffset = 0;
+            for (size_t i = 0; i < numWells; ++i) {
+                WellIndex wi;
+                wi.wellNumber = i;
+                wi.wellCompletionOffset = totalWellCompletionOffset;
+
+                totalWellCompletionOffset += getNumCompletions(wells[i], reportStep);
+                fillCompletionsMap(grid, wells[i], wi, reportStep);
+                std::pair <std::string, WellIndex> p(wells[i]->name(), wi);
+                newMap.insert(p);
+            }
+
+            return newMap;
+        }
+
+        void fillCompletionsMap(const EclipseGridConstPtr& grid, const WellPtr& well, WellIndex& wi, const int reportStep) const
+        {
+            CompletionSetConstPtr complSet = well->getCompletions(reportStep);
+            size_t open_completions_count = 0;
+
+            for (size_t idx = 0; idx < complSet->size(); ++idx) {
+                CompletionConstPtr completion = complSet->get(idx);
+
+                if (completion->getState() == WellCompletion::StateEnum::OPEN) {
+                    int i = completion->getI();
+                    int j = completion->getJ();
+                    int k = completion->getK();
+
+                    size_t globalIndex = grid->getGlobalIndex(i, j, k);
+                    std::pair <size_t, size_t> p(globalIndex, open_completions_count);
+                    wi.completionMap.insert(p);
+                    ++open_completions_count;
+                }
+            }
+        }
     };
 
 } // namespace Opm
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -288,7 +288,7 @@ namespace Opm
                 wi.wellCompletionOffset = totalWellCompletionOffset;
 
                 totalWellCompletionOffset += getNumCompletions(wells[i], reportStep);
-                fillCompletionsMap(grid, wells[i], wi, reportStep);
+                fillCompletionsMap(grid, wells[i], reportStep, wi);
                 std::pair <std::string, WellIndex> p(wells[i]->name(), wi);
                 newMap.insert(p);
             }
@@ -296,7 +296,7 @@ namespace Opm
             return newMap;
         }
 
-        void fillCompletionsMap(const EclipseGridConstPtr& grid, const WellPtr& well, WellIndex& wi, const int reportStep) const
+        void fillCompletionsMap(const EclipseGridConstPtr& grid, const WellPtr& well, const int reportStep, WellIndex& wi) const
         {
             CompletionSetConstPtr complSet = well->getCompletions(reportStep);
             size_t open_completions_count = 0;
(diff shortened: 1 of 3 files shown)


opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,254 @@ namespace Opm
...
+            size_t numWells = wells.size();
+            std::vector<double> bhpTemp(bhp_.begin(), bhp_.end());
+            std::vector<double> temperatureTemp(temperature_.begin(), temperature_.end());
+            std::vector<double> wellRatesTemp(wellrates_.begin(), wellrates_.end());
+            std::vector<double> perfratesTemp(perfrates_.begin(), perfrates_.end());
+            std::vector<double> perfpressTemp(perfpress_.begin(), perfpress_.end());
+
+            std::map<std::string, WellIndex> newMap;
+
+            size_t totalWellCompletionOffset = 0;
```
joakim-hove: The whole 'create map' functionality could go to a separate method:
```C++
std::map<std::string, WellIndex> newMap = createWellIndexMap( state , reportStep );
```

joakim-hove:
Some of us have briefly discussed it before; and I guess this is a time to revisit the topic: While iterating through the Schedule section we could record events which require reinitializations by the simulator; i.e. `NEW_WELL`, `NEW_COMPLETION`, `CONTROL_MODE_CHANGE` and so on. One day we really feel adventurous we might even have event for geological property changes in the Schedule section.

atgeirr:
Hi @edbru, thank you for the PR! I will review it now (not done it earlier as I have been away until today), just be warned that I will review this at a very high "nitpick level" since you are fairly new to contributing to opm-core. This is not meant to discourage you, but to make it easier for me to review your next contribution!

> 1) One instance of the wellstate object for a simulation (all reportsteps). Call the resize method for each reportstep.
2) With only one instance of WellState, calling resize for each reportstep, WellsManager, WellsManager.cwells, and the call to wellstate.init (for each reportstep) can be removed.

I do not think that this change to the WellState is sufficient to remove those parts, but I see that it can be a step towards that.

> Some of us have briefly discussed it before; and I guess this is a time to revisit the topic: While iterating through the Schedule section we could record events which require reinitializations by the simulator; i.e. NEW_WELL, NEW_COMPLETION, CONTROL_MODE_CHANGE and so on. One day we really feel adventurous we might even have event for geological property changes in the Schedule section.

@joakim-hove: I think this could be quite useful for many different purposes, for example we might use such information to provide better starting points for our Newton iterations or guide timestepping strategy.

opm/core/simulator/WellState.hpp
```diff
@@ -22,11 +22,20 @@
...
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <vector>
 #include <cassert>
+#include <utility>
+#include <iostream>
+#include <string>
 
 namespace Opm
 {
+    struct WellIndex{
```
atgeirr: I *prefer* to have classes (also namespaces and functions) have their opening brace on the next line (as you did for the resize() method), but I will accept it on the same line (which I prefer is used for `while`, `if`, `for` etc.). If you do keep it on the same line, there must be a space before the opening brace.

atgeirr: This page can be of help for seeing the coding standard I expect:

http://www.opm-project.org/wiki/index.php/Suggested_coding_standard

opm/core/simulator/WellState.hpp
```diff
@@ -22,11 +22,20 @@
...
+        size_t wellNumber;
+        size_t wellCompletionOffset;
+        std::map<std::size_t, size_t> completionMap;
+    };
```
atgeirr: The previous struct has no documentation, either of its members or for the struct itself. Please add comments explaining what it is for, especially since "well index" is a term that has been misused before to mean the "connection transmissibility factor". The term "well productivity index" traditionally mean the relation between flow and drawdown (pressure differential).

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
             }
         }
 
+        void resize(const EclipseState& state, size_t reportStep)
```
atgeirr: Every public method in a class should be documented using the doxygen format. It should have at least a rudimentary explanation of what it does. This method is definitely nontrivial so it must be explained. (A vector::resize() taking an integral number is self-explaining, but this is not).

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+            std::vector<double> bhpTemp(bhp_.begin(), bhp_.end());
+            std::vector<double> temperatureTemp(temperature_.begin(), temperature_.end());
+            std::vector<double> wellRatesTemp(wellrates_.begin(), wellrates_.end());
+            std::vector<double> perfratesTemp(perfrates_.begin(), perfrates_.end());
+            std::vector<double> perfpressTemp(perfpress_.begin(), perfpress_.end());
+
+            std::map<std::string, WellIndex> newMap = createWellIndexMap(state, wells, reportStep);
+
+            bhp_.resize(numWells);
+            temperature_.resize(numWells, 273.15 + 20);
```
atgeirr: It is better to use a constant for such magic numbers (for example `const double default_temp = 273.15 + 20.0;`.
(I do see that the existing code does the same though...)

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+            wellrates_.resize(numWells * numPhases, 0.0);
+            perfrates_.resize(numCompl, 0.0);
+            perfpress_.resize(numCompl, -1e100);
+
+            for (auto oldIter = wellIndexMap_.begin(); oldIter != wellIndexMap_.end(); ++oldIter){
```
atgeirr: Please add a single space before the opening brace.

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+                const std::string& wellName = (*oldIter).first;
+                auto newIter = newMap.find(wellName);
+
+                if(newIter != newMap.end()){
```
atgeirr: Add a space after the `if`and before the opening brace. In general, there are many functions and loops in this patch that lack this.

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+                    const WellIndex& oldWi = (*oldIter).second;
+                    size_t oldIndex = oldWi.wellNumber;
+
+                    const WellIndex& newWi = (*newIter).second;
+                    size_t newIndex = newWi.wellNumber;
+
+                    bhp_[newIndex] = bhpTemp[oldIndex];
+                    temperature_[newIndex] = temperatureTemp[oldIndex];
+
+                    for (int i = 0; i < numPhases; i++){
```
atgeirr: This is a very minor thing, but in C++ the common way to write this loop (and therefore the one you should use unless you have explicit reasons not to) is with the preincrement operator (`++i`) and not the postincrement operator (`i++`). Part of the reason for this is: for general objects, the postincrement can be less efficient, and it is harder to reason about (you cannot implement it without a temporary copy). While there will be no loss of efficiency for integers, it is still a good idea to use preincrement by default.

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+                        wellrates_[newIndex * numPhases + i] = wellRatesTemp[oldIndex * numPhases + i];
+                    }
+
+                    for (auto oldComplIter = oldWi.completionMap.begin(); oldComplIter != oldWi.completionMap.end(); ++oldComplIter){
```
atgeirr: Add space before opening brace (I will not make further comments about this). Also I notice that here, the preincrement operator is used (which is preferred), good!

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
         }
 
+        void resize(const EclipseState& state, size_t reportStep)
+        {
+            std::vector<WellPtr> wells = state.getSchedule()->getOpenWells(reportStep);
+            size_t numPhases = state.getNumPhases();
+            int numCompl = getNumCompletions(wells, reportStep);
+
+            size_t numWells = wells.size();
+            std::vector<double> bhpTemp(bhp_.begin(), bhp_.end());
```
atgeirr: I think that the somethingTemp names here would be better replaced with somethingOld or oldSomething (like oldIndex further down). Also, you can make copies with a lot simpler syntax, either
```
std::vector<double> oldBhp = bhp_;
```
(which I think is most readable) or
```
std::vector<double> oldBhp(bhp_);
```

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+                            size_t newWellCompletionOffset = newWi.wellCompletionOffset;
+
+                            perfrates_[newWellCompletionOffset + newComplIndex] = perfratesTemp[oldWellCompletionOffset + oldComplIndex];
+                            perfpress_[newWellCompletionOffset + newComplIndex] = perfpressTemp[oldWellCompletionOffset + oldComplIndex];
+                        }
+                    }
+                }
+            }
+
+            wellIndexMap_ = newMap;
```
atgeirr: This seems to me to be the only place wellIndexMap_ is updated, unless I missed something. So I see that this function is expected to be used every report step, in order to transfer state from one step's configuration to the next. However this does not interact at all with the init() method, which (if used) will have the values it provided overwritten by resize(). Also, there is already functionality for this in the special init() method of opm-autodiff's WellStateFullyImplicitBlackoil (WSFIB), that takes a previous state as an argument (and does a similar job of updating where possible). I think the implementation provided here is fine, and can probably eventually replace the one in WSFIB, but there are several issues with that:
 - How to provide this function in a way that allows more specialised classes to add more fields, and automatically have them updated (or at least avoiding copying the entire method).
 - How to allow initialisation depending on a reservoir state and controls (as the init() methods currently allow, for example initialising bhp of rate controlled wells depending on reservoir pressures).

The entire resize() method looks a little like a partial re-implementation of init(), made in such a manner that we do not require a new WellState object for each new report step (good!) but it is only partial, and therefore cannot be used to replace the current methods.

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+        }
+
+        void setBhpValue(size_t index, double value){
```
atgeirr: I do not see the point of adding this method, for the following reasons:
 - Doing `state.bhp()[index] = value` serves this purpose well enough.
 - It adds overhead by checking the index, but fails to make it an error if the index is out-of-bounds!

A better implementation would `assert(index < bhp_.size())` instead of the if.

opm/core/simulator/WellState.hpp
```diff
@@ -107,6 +116,146 @@ namespace Opm
...
+            if (index < bhp_.size()){
+                bhp_[index] = value;
+            }
+        }
+
+        double getBhpValue(size_t index) const{
```
atgeirr: This method (and all like it) should also be removed. The mistaken checking of the setter function is even worse here, because if the index is out of bounds we return an uninitialised value! Again, an `assert` would be a lot better.

opm/core/simulator/WellState.hpp
```diff
@@ -133,6 +282,86 @@ namespace Opm
         std::vector<double> wellrates_;
         std::vector<double> perfrates_;
         std::vector<double> perfpress_;
+
+        std::map<std::string, WellIndex> wellIndexMap_;
+
+        int getWellMapIndex(std::string wellName) const{
```
atgeirr: In general, pass function arguments as const references.

opm/core/simulator/WellState.hpp
```diff
@@ -133,6 +282,86 @@ namespace Opm
...
+
+            if(iter != wellIndexMap_.end()){
+                const WellIndex& wi = (*iter).second;
+                retVal = wi.wellNumber;
+            }
+
+            return retVal;
+        }
+
+        int getNumCompletions(std::vector<WellPtr>& wells, int reportStep) const{
```
atgeirr: Again, pass `wells` as a const reference. Passing it as a reference implies that you will change it inside the function.

opm/core/simulator/WellState.hpp
```diff
@@ -133,6 +282,86 @@ namespace Opm
...
+
+            for(size_t i = 0; i < wells.size(); i++){
+                int numComplForWell = getNumCompletions(wells[i], reportStep);
+                numCompl += numComplForWell;
+            }
+
+            return numCompl;
+        }
+
+        int getNumCompletions(WellPtr& well, int reportStep) const{
```
atgeirr: Pass well as a const reference. I also prefer that reportStep is passed as a const int (prevents it from being changed accidentally in the function body), but I will not require it.

opm/core/simulator/WellState.hpp
```diff
@@ -133,6 +282,86 @@ namespace Opm
...
+
+                if(complState == WellCompletion::StateEnum::OPEN){
+                    numComplForWell++;
+                }
+            }
+
+            return numComplForWell;
+        }   
+
+        std::map<std::string, WellIndex> createWellIndexMap(const EclipseState& state, std::vector<WellPtr>& wells, size_t reportStep){
```
atgeirr: Again, use const reference arguments.

opm/core/simulator/WellState.hpp
```diff
@@ -133,6 +282,86 @@ namespace Opm
...
+                totalWellCompletionOffset += getNumCompletions(wells[i], reportStep);
+                fillCompletionsMap(grid, wells[i], wi, reportStep);
+                std::pair <std::string, WellIndex> p(wells[i]->name(), wi);
+                newMap.insert(p);
+            }
+
+            return newMap;
+        }
+
+        void fillCompletionsMap(EclipseGridConstPtr& grid, WellPtr& well, WellIndex& wi, int reportStep) const{
```
atgeirr: Only `wi` is modified, all other arguments should be const. Also, output arguments like `wi` should be placed last in the argument list.

atgeirr:
Having finished reviewing this, I must say that even if all the nitpicks were fixed, it still would not be ready for merging, for the deeper reasons I discussed above: it aims to replace init(), but is at this point only partially complete.

There are also two issues I have not touched upon yet:
 - This makes the WellState depend on the parser classes, thereby moving the boundary between the code that is independent of the parser and the code that depends on it. With this (assuming it evolves into a replacement for the current init() methods), it will no longer be possible to use the WellState or its extensions without an eclipse deck.
 - Care must be taken for the code to work in the parallel setting. This can be seen in the treatment in the WellsManager class. The parser is for the foreseeable future not parallel-aware, and expects to be in command of the entire domain, while in a parallel run the WellState of each process only contains data for the wells owned by that domain (for now we require that each well is owned by a single domain and that all its completions lie in that domain, but that might be loosened eventually).

joakim-hove:
> This makes the WellState depend on the parser classes

Well; the current way to make a `WellState` object is from `struct wells` which is instantiated through `WellsManager` - which again uses the `EclipseState` to get going, so practically no change there - but of course `struct Wells` can in principle be instantiated without the use of `Wellsmanager` - so I agree this is formally a new dependency.

But; as this task has progressed I have had this slightly iffy feeling - which I really should have clarified better with e.g. @atgeirr - is this really necessary at all? Many report steps the schedule configuration does not change at all - and the `WellState` can just carry over to the next report step unchanged; if we then combined this with an event like system which would trigger the need for recalculations when that applies - everything could be significantly simplified?

bska:
> the current way to make a `WellState` object is from `struct Wells` which is instantiated through `WellsManager` - which again uses the `EclipseState` to get going

Yes.  Technically, it is **possible** to instantiate the `WellsManager` without the `EclipseState` (from a `struct Wells` that is otherwise created--typically by the C interface of [`<opm/core/wells.h>`](https://github.com/OPM/opm-core/blob/master/opm/core/wells.h)), but the resulting context is so limited as to be virtually useless.

I have said so previously, and I'll repeat it now for completeness, that I think the `WellsManager` has served its purpose and that our goal should be to deprecate and remove it.  At this time it is the moral equivalent of
```C++
struct Delete {
    void operator()(Wells* W) {
        destroy_wells(W);
    }
};
std::unique_ptr<Wells, Delete> W;
```
and we don't really need it.  Parallel capabilities ought to be hosted elsewhere.

atgeirr:
> Yes. Technically, it is possible to instantiate the WellsManager without the EclipseState (from a struct Wells that is otherwise created--typically by the C interface of <opm/core/wells.h>), but the resulting context is so limited as to be virtually useless.

The init() functionality of the WellState does not depend on the WellsManager, but on Wells. I also disagree that the WellsManager is essentially just a smart pointer, its primary contents is the initialisation of Wells from an EclipseState (and in the parallel case: creating node-specific Wells objects even though the EclipseState is identical across nodes).

I think we should let this rest a little. I understand that this is primarily motivated from a wish to fully support the restart mechanism, and perhaps we need to discuss how to do that before we decide on this.  (So let's not merge or close this PR yet, but keep it alive until we have decided.)

bska:
> its primary contents is the initialisation of `Wells` from an `EclipseState`

And that is essentially busy-work with no real purpose.  We're just copying/extracting some data from the `EclipseState` and putting it into another container.  If we have an `EclipseState` (like in the opm-autodiff simulators), there is no reason we cannot query that object directly.  While the topology section of `struct Wells` may serve a purpose, the constraints portion is obsolete.

atgeirr:
> We're just copying/extracting some data from the EclipseState and putting it into another container.

In the parallel case, we are also essentially filtering/partitioning the data.

bska:
>> We're just copying/extracting some data from the `EclipseState` and putting it into another container.
>
> In the parallel case, we are also essentially filtering/partitioning the data.

Different data.  The parallel case deals with the well topology, not the well constraints.  There is possibly still some reason to representing the topology outside the `EclipseState`.

atgeirr:
> Different data. The parallel case deals with the well topology, not the well constraints. There is possibly still some reason to represent the topology outside the EclipseState.

I agree that if we separate constraints/controls from well topology we could query the (same on every node) Schedule and its associated classes for the controls, as long as the topology is distributed. I would like to use an abstract interface for the constraints part then, but it could be more or less a subset of the current Schedule::Well interface (I do not think getHeadI() or addWELSPECS() belong in such an interface for example, nor would they be needed).

joakim-hove:
>I agree that if we separate constraints/controls from well topology we could query the (same on every node) Schedule and its associated classes for the controls,

I think that is a very good idea - and essentially what I have been nagging about every now and then:-)

edbru:
Have now fixed up the small and easy issues that was commented. As you commented, let the PR await further discussions.

atgeirr:
I cannot see any fixup commits, it looks to me like you rebased the PR, correct?

While that is fine to do before we merge (although I do not think it is necessary) it can often be convenient to have the fixups as a separate commit (or commits), especially if there are many of them, to make it easier to review. As for this PR, I'll assume that you fixed it all as I wanted it to be :-)

Apart from that, we need to discuss how to take this further (towards reducing busywork as @bska put it), and I'd like some agreement on that before merging.

opm/core/simulator/WellState.hpp
```diff
@@ -133,6 +228,94 @@ namespace Opm
...
+                totalWellCompletionOffset += getNumCompletions(wells[i], reportStep);
+                fillCompletionsMap(grid, wells[i], wi, reportStep);
+                std::pair <std::string, WellIndex> p(wells[i]->name(), wi);
+                newMap.insert(p);
+            }
+
+            return newMap;
+        }
+
+        void fillCompletionsMap(const EclipseGridConstPtr& grid, const WellPtr& well, WellIndex& wi, const int reportStep) const
```
atgeirr: Having scanned through the revised version, I can only say thanks for doing a thorough job with the revision! The line above is the only change I miss: please move the output argument last.

edbru:
Changed order of method-parameters, output parameter last.

atgeirr:
Thanks for the fixups. Now for the hard questions: is the best way forward to expand this function to include the init()-functionality? In that case, it might be good to rename it (to init perhaps...?), also we should look at how it would be used in the simulator class.

chflo:
Closing this pull request from Edbru - will continue his work and submit a new PR when ready.

PR #815, opened 2015-06-05
Initial PR to add NNC support. 

totto82:
This PR provide a NNC class and a test case to read NNC from deck and to add NNCs 
No effort has currently been made to add support for NNC in the simulator itself. 

I hope for input if you miss important functionality.


bska:
Just from a *very* brief look I have one question on the design.  Why do you have separate `nnc1_` and `nnc2_` arrays replete with read-only linearised access?  Do you foresee ever using only one of those arrays?  Generally, merging those arrays into a single object that mirrors the `UnstructuredGrid`'s `->face_cells` array is more useful from the point of view of the simulator.

opm/core/grid/nnc.cpp
```diff
@@ -0,0 +1,64 @@
...
+*/
+#include "config.h"
+#include "nnc.hpp"
+#include <array>
+
+namespace Opm
+{
+    NNC::NNC(Opm::DeckConstPtr deck, EclipseGridConstPtr eclipseGrid)
+    {
+        Opm::DeckKeywordConstPtr nnc = deck->getKeyword("NNC");
```
joakim-hove: To accomodate for the case of several `NNC`keywords you should have an outer loop based on the `Deck::getKeywordList()`method.

joakim-hove:
I guess this is a valuable contribution which OPM should have; apart from the comments I have made I think it can be merged. However I would *really like* this patch to be targetting opm-parser instead of opm-core; this is the type "slightly-more-usable-than-the-deck" objects which the `EclipseState` class is full of; in my opinion the `EclipseState::EclipseGrid` class.would be a good home for the `NNC` class - that will also be a very convenient place to have it when adding `NNC` information to the `EGRID` output file.

Apart from the changes required to bolt it on to the `EclipseGrid` I think this can target opm-parser without any changes.

totto82:
>Just from a very brief look I have one question on the design. Why do you have separate  nnc1_  and  nnc2_  arrays replete with read-only linearised access? Do you foresee ever using only one of those arrays? Generally, merging those arrays into a single object that mirrors the  UnstructuredGrid 's  ->face_cells  array is more useful from the point of view of the simulator.

@bska  The reason for using two seperate arrays is to mimic Eclipse outputs (NNC1, NNC2), but I agree that it may be more convinient from the simulator side to just have one. But for now I leave it the way it is. 

The other comments are adressed in OPM/opm-parser#491

PR #818, opened 2015-06-11
Added missing free( ) to malloc() memory.

joakim-hove:
As discovered by @andlaus and @bska


bska:

PR #819, opened 2015-06-11
Opm 188 Read and write of RESTART wellstate data

edbru:
Added read and write of RESTART wellstate data.
This task is a step towards supporting RESTART functionality in Flow. 

The task is "self-contained", not dependent on other tasks.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -158,6 +158,7 @@ list (APPEND MAIN_SOURCE_FILES
 # find tests -name '*.cpp' -a ! -wholename '*/not-unit/*' -printf '\t%p\n' | sort
 list (APPEND TEST_SOURCE_FILES
   tests/test_writenumwells.cpp
+        tests/test_readWriteWellStateData.cpp
 	tests/test_EclipseWriter.cpp
 	tests/test_EclipseWriteRFTHandler.cpp
 	tests/test_compressedpropertyaccess.cpp
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -41,6 +41,7 @@ list (APPEND MAIN_SOURCE_FILES
 	opm/core/grid/cpgpreprocess/uniquepoints.c
 	opm/core/io/eclipse/EclipseGridInspector.cpp
 	opm/core/io/eclipse/EclipseWriter.cpp
+        opm/core/io/eclipse/EclipseReader.cpp
 	opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
 	opm/core/io/eclipse/writeECLData.cpp
 	opm/core/io/OutputWriter.cpp
@@ -289,6 +290,7 @@ list (APPEND PUBLIC_HEADER_FILES
 	opm/core/io/eclipse/EclipseGridInspector.hpp
 	opm/core/io/eclipse/EclipseUnits.hpp
 	opm/core/io/eclipse/EclipseWriter.hpp
+        opm/core/io/eclipse/EclipseReader.hpp
 	opm/core/io/eclipse/EclipseWriteRFTHandler.hpp
 	opm/core/io/eclipse/writeECLData.hpp
 	opm/core/io/OutputWriter.hpp
--- a/opm/core/io/eclipse/EclipseReader.cpp
+++ b/opm/core/io/eclipse/EclipseReader.cpp
@@ -1,0 +1,62 @@
+/*
+  Copyright 2015 Statoil ASA.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "EclipseReader.hpp"
+#include <iostream>
+#include <opm/core/simulator/WellState.hpp>
+
+#include <ert/ecl/ecl_file.h>
+
+namespace Opm
+{
+    void restoreOPM_XWELKeyword(const std::string restart_filename, int reportstep, WellState& wellstate)
+    {
+        const char * keyword = "OPM_XWEL";
+        const char* filename = restart_filename.c_str();
+
+        ecl_file_type* file_type = ecl_file_open(filename, 0);
+        bool block_selected = ecl_file_select_rstblock_report_step(file_type , reportstep);
+
+        if (block_selected) {
+            ecl_kw_type* xwel = ecl_file_iget_named_kw(file_type , keyword, 0);
+            const double* xwel_data = ecl_kw_get_double_ptr(xwel);
+            size_t offset = 0;
+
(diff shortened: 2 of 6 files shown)


bska:
Overall I think this is a useful feature that should go in, but I also have a number of questions and comments on the specifics of the implementation so I won't be merging the code as-is.

I will add comments inline in the relevant sections of the code.

CMakeLists_files.cmake
```diff
@@ -41,6 +41,7 @@ list (APPEND MAIN_SOURCE_FILES
 	opm/core/grid/cpgpreprocess/uniquepoints.c
 	opm/core/io/eclipse/EclipseGridInspector.cpp
 	opm/core/io/eclipse/EclipseWriter.cpp
+        opm/core/io/eclipse/EclipseReader.cpp
```
bska: Note: For historical reasons, we use `<tab>` as the indentation character in `CMakeLists_files.cmake`.  I think you should too.

opm/core/io/eclipse/EclipseReader.cpp
```diff
@@ -0,0 +1,62 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "EclipseReader.hpp"
+#include <iostream>
```
bska: This is a `.cpp` file, so it's not **that** critical, but I still don't see the reason to `#include <iostream>` here.  You don't appear to be using any of those facilities.  Is `<iostream>` needed as a prerequisite for other headers?

opm/core/io/eclipse/EclipseReader.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <ert/ecl/ecl_file.h>
+
+namespace Opm
+{
+    void restoreOPM_XWELKeyword(const std::string restart_filename, int reportstep, WellState& wellstate)
+    {
+        const char * keyword = "OPM_XWEL";
+        const char* filename = restart_filename.c_str();
+
+        ecl_file_type* file_type = ecl_file_open(filename, 0);
```
bska: Does `ecl_file_open()` acquire any resources?  If so, do we have to dispose of those to prevent resource leaks?

opm/core/io/eclipse/EclipseReader.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "EclipseReader.hpp"
+#include <iostream>
+#include <opm/core/simulator/WellState.hpp>
+
+#include <ert/ecl/ecl_file.h>
+
+namespace Opm
+{
+    void restoreOPM_XWELKeyword(const std::string restart_filename, int reportstep, WellState& wellstate)
```
bska: Please use reference-to-const for input arguments (`const std::string& restart_filename` in this case).

opm/core/io/eclipse/EclipseReader.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+        ecl_file_type* file_type = ecl_file_open(filename, 0);
+        bool block_selected = ecl_file_select_rstblock_report_step(file_type , reportstep);
+
+        if (block_selected) {
+            ecl_kw_type* xwel = ecl_file_iget_named_kw(file_type , keyword, 0);
+            const double* xwel_data = ecl_kw_get_double_ptr(xwel);
+            size_t offset = 0;
+
+            for (size_t i = 0; i < wellstate.bhp().size(); ++i) {
```
bska: Is there a reason you're using explicit loops rather than `std::copy()` or `std::vector<>::assign()`?

opm/core/io/eclipse/EclipseReader.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef ECLIPSEREADER_HPP
+#define ECLIPSEREADER_HPP
+
+#include <iostream>
```
bska: No.

Don't `#include <iostream>` in headers unless the header cannot be autonomous without it (**very** rare).

opm/core/io/eclipse/EclipseWriter.cpp
```diff
@@ -1221,7 +1243,11 @@ void EclipseWriter::writeTimeStep(const SimulatorTimerInterface& timer,
         std::vector<int>         iwell_data( numWells * Opm::EclipseWriterDetails::Restart::NIWELZ , 0 );
         std::vector<int>         icon_data( numWells * ncwmax * Opm::EclipseWriterDetails::Restart::NICONZ , 0 );
 
+        int sz = wellState.bhp().size() + wellState.perfPress().size() + wellState.perfRates().size() + wellState.temperature().size() + wellState.wellRates().size();
```
bska: `std::vector<double>::size_type` or `std::size_t`, please.

opm/core/io/eclipse/EclipseWriter.cpp
```diff
@@ -322,8 +323,29 @@ class Restart : private boost::noncopyable
         iwel_data[ offset + IWEL_STATUS_ITEM ] = EclipseWriter::eclipseWellStatusMask(well->getStatus(currentStep));
     }
 
+    void addRestartFileXwelData(std::vector<double>& xwel_data, const WellState& wellState) const {
```
bska: Output parameters (`xwel_data`) should be placed last.

bska:
> I will add comments inline in the relevant sections of the code.

In addition to the inline comments, I also think the serialisation code is a little too brittle.  When maintaining this at a later stage one has to be very careful not to break the ordering of the vectors (e.g., when adding a new vector or removing an old) in the function pair `addRestartFileXwelData()` and `restoreOPM_XWELKeyword()`.  I would prefer a little more encapsulation of that aspect of the code.

opm/core/io/eclipse/EclipseReader.cpp
joakim-hove: > Does ecl_file_open() acquire any resources?

Yes indeed - every `ecl_file_open()` should be matched with a `ecl_file_close()`.

bska: >> Does `ecl_file_open()` acquire any resources?
>
> Yes indeed

In that case, can it fail (and return 0)?

joakim-hove: > In that case, can it fail (and return 0)?

Yes indeed

bska: Thanks a lot for that information.  In light of this we clearly need to be a lot more careful in the implementation of this function.

edbru:
Fixed review comments, and updated test due to merged IOConfig functionality.

bska:
> Fixed review comments,

Thanks.  I still notice that `EclipseReader.hpp` `#include`s the `<iostream>` header and I'd really prefer that it did not.  Why is that needed?

edbru:
Changed include of iostream to include of string.

chflo:
The EclipseReader will be furter extended with read of the solution state (ie SGAS, SWAT, PRESSURE).
The current PR can be merged, or you can wait until the reader has been updated with the solution fields.

bska:
> The current PR can be merged, or you can wait until the reader has been updated with the solution fields.

I think I'd prefer to get this in now (subject to edbru/opm-core#1).  This is sufficiently useful in its own right that having it in master is beneficial.

chflo:
Good:)

atgeirr:
@chflo, please merge edbru/opm-core#1 so that we can merge this PR.

chflo:
Have merged as requested

atgeirr:
Thanks, will merge!

PR #820, opened 2015-06-17
Adding two more warnings to disable_warnings.h.

GitPaean:
* -Wpragmas 

This is used to suppress the warnings [-Wpragmas] when the compilers do not recognize the warning options. like 
```
opm-core/opm/core/utility/platform_dependent/disable_warnings.h:59:32: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
 #pragma GCC diagnostic ignored "-Wdeprecated-register"
```
* -Wunused-but-set-variable

This is used to suppress the warnings for unused but set variables.


atgeirr:
Thanks, this should not impact anything other than warnings (and only for third-parties and only for users with SILENCE_EXTERNAL_WARNINGS turned on). Merging.

PR #823, opened 2015-06-26
Use fluidstates for satfuncs

andlaus:
the intention of this patch series is to avoid having to explicitly pass all quantities on upon which a "saturation" function can possibly depend on and replace it by objects that expose them in a "collected" manner, i.e., which implement a subset of opm-material's fluid state API.

currently there are two fluid states shipped as an example: `BlackoilStateToFluidState` which takes a blackoil state object and exposes it as a fluid state and `ExplicitArraysFluidState` which -- surprise -- takes global arrays of quantities and exposes them through the (local) fluid state API.

performance-wise I could not observe any difference beyond the measurement noise of this patch series compared to the current master version.

still to be done is to get rid of the `eval*Deriv()` methods in the saturation function classes and also of the variants which take parameters for end-point-scaling and/or hysteresis in order to avoid the 2^#numFeatures problem of the current API: The first thing can be accomplished using a "localized AD" approach, similar to the one used in opm-material, the latter can be achieved by folding the required parameters into the saturation function objects directly and use pointers to the quantities which are not cell specific. this is a large job, though, and will thus be done another day.


atgeirr:
Thanks for the effort, I will now start reviewing this. Have you run through the complete Norne benchmark with this? I assume that this change *should* produce identical runs as before, but have you verified that?

opm/core/simulator/ExplicitArraysFluidState.hpp
```diff
@@ -0,0 +1,95 @@
...
+     * index.
+     */
+    void setIndex(unsigned arrayIdx)
+    { arrayIdx_ = arrayIdx; }
+
+    /*!
+     * \brief Set the array containing the phase saturations.
+     *
+     * This array is supposed to be of size numPhases*size and is not allowed to be
+     * deleted while the ExplicitArraysFluidState object is alive.
```
atgeirr: From the implementation of saturation() below I can tell that the saturation array is expected to have phases running fastest (not cells), this fact should also be stated in the documentation here.

opm/core/simulator/BlackoilStateToFluidState.hpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+class BlackoilStateToFluidState
+{
+public:
+    typedef double Scalar;
+
+    enum { numPhases = 3 };
+    enum { numComponents = 3 };
+
+    BlackoilStateToFluidState(const BlackoilState& blackoilState)
```
atgeirr: Should document that this does not take a copy of the BlackoilState but only references it.

opm/core/props/satfunc/SatFuncMultiplexer.hpp
```diff
@@ -0,0 +1,268 @@
...
+ *        runtime.
+ */
+class SatFuncMultiplexer
+{
+public:
+    typedef NonuniformTableLinear<double> TableType;
+
+    enum SatFuncType {
+        Gwseg, // <- supposed to be the default
+        Stone2,
```
atgeirr: We also need Stone1 soon, but I guess that would not be very hard to add.

opm/core/props/satfunc/SatFuncMultiplexer.hpp
```diff
@@ -0,0 +1,268 @@
...
+        setSatFuncType(satFuncType);
+
+        OPM_MULTIPLEXER_SATFUNC_CALL(satFunc.init(eclState,
+                                                  tableIdx,
+                                                  phaseUsage,
+                                                  /*samples=*/0));
+    }
+
+    /*!
+     * \brief Returns which saturation function is actually selected.
```
atgeirr: Wrong doc.

opm/core/props/satfunc/SatFuncMultiplexer.hpp
```diff
@@ -0,0 +1,268 @@
...
+    };
+
+// this is a helper macro which helps to save us from RSI in the following
+#define OPM_MULTIPLEXER_SATFUNC_CALL(codeToCall)    \
+    switch (satFuncType_) {                         \
+    case Gwseg: {                                   \
+        typedef SatFuncGwseg<TableType> SatFunc;    \
+        if (false) SatFunc dummyForWarning;         \
+        auto& satFunc = getGwseg();                 \
+        codeToCall;                                 \
```
atgeirr: Being always nervous around macros, is it not a good idea to wrap the codeToCall in a do-loop?

opm/core/props/satfunc/SatFuncMultiplexer.hpp
```diff
@@ -0,0 +1,268 @@
...
+#undef OPM_MULTIPLEXER_SATFUNC_CALL
+
+private:
+    bool wasInitialized_;
+    SatFuncType satFuncType_;
+    union {
+        char gwseg[sizeof(SatFuncGwseg<TableType>)];
+        char simple[sizeof(SatFuncSimple<TableType>)];
+        char stone2[sizeof(SatFuncStone2<TableType>)];
+    } satFuncsData_  __attribute__((aligned(sizeof(double))));
```
atgeirr: Why have you used this scary union instead of a base class pointer? Also, is the __attribute__ standard C++ or just something that happens to work on gcc/clang (would it work for example with visual studio)?

opm/core/props/satfunc/SatFuncMultiplexer.hpp
```diff
@@ -0,0 +1,268 @@
...
+
+    SatFuncBase<TableType>& getSatFuncBase()
+    {
+        return const_cast<SatFuncBase<TableType>&>(static_cast<const SatFuncMultiplexer*>(this)->getSatFuncBase());
+    }
+
+// the strict aliasing compiler warnings need to be ignored here. This is safe because we
+// deal with the alignment of the data ourselfs.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
```
atgeirr: Are the warnings caused by the reinterpret_cast usage?

opm/core/props/satfunc/SatFuncMultiplexer.hpp
bska: > Also, is the `attribute` standard C++

It is not.

> would it work for example with visual studio

No.  Visual C++ uses different syntax to achieve the same (or similar) effect--e.g.,
```C++
__declspec(align(32)) struct Str1{
   int a, b, c, d, e;
};
```

CMakeLists_files.cmake
```diff
@@ -365,6 +362,7 @@ list (APPEND PUBLIC_HEADER_FILES
 	opm/core/props/rock/RockFromDeck.hpp
 	opm/core/props/satfunc/SatFuncBase.hpp
 	opm/core/props/satfunc/SatFuncGwseg.hpp
+	opm/core/props/satfunc/SatFuncMultiplexer.hpp
 	opm/core/props/satfunc/SatFuncSimple.hpp
```
atgeirr: Seems to be missing BlackoilStateToFluidState and ExplicitArraysFluidState files.

atgeirr:
I have now done an initial reading of the PR. First, I think that transitioning to an API which is more opm-material-like is ok, and something that should be considered. However this PR does quite a bit more than that, and I do not understand why:
 - The use of a template parameter to indicate which three-phase model to use has been replaced. Why remove the template? 
 - The replacement uses runtime switches supplied (mostly) by a macro. Why use macros instead of common inheritance-based polymorphism?

The (in my opinion) biggest problem with the existing saturation functions (the classes SatFuncGwseg, SatFuncStone2 and SatFuncSimple) is that they started out as duplicates of each other with just a small difference, ripe for refactoring, but instead hysteresis and end-point scaling was added to the gwseg version (and perhaps some was added to the others), and now they are too different to easily refactor. 

One way out of this might be to just delete the Simple and Stone2 versions, I do not think they are used in any serious manner. Then we would create new ones (at least for Stone2 and Stone1 which we will need soon) from scratch while refactoring to avoid copy-and-paste code. For this to be a good approach the end-point-scaling and hysteresis treatment must be orthogonal to the three-phase treatment. I am assuming that is the case, @osae please correct me if that is wrong! I am hoping that the choice of three-phase model could be implemented with a few classes, each of them with only a single function, that did this each in their own way.

@andlaus, you point out that we should get rid of the combinatorial explosion of variants due to endpoint scaling etc. I agree completely, should we not do that before we implement this API change? Are there any immediate benefits from doing it the other way round?

opm/core/props/satfunc/SatFuncMultiplexer.hpp
andlaus: > > would it work for example with visual studio
>
> No. Visual C++ uses different syntax to achieve the same (or similar) effect--e.g.,

True. it does not work in VC++, but there if I remember correctly there may be a c++-2011 way to do it. The problem is that the C++-11 style attributes are only supported in GCC 4.8 and later.

when/if VC is really supposed to be supported, we should deal with it when we pushing for it (i.e., hopefully never ;)

also, be aware that this alignment stuff is just a safety/performance measure: at least on X86 machines it will also work without it, but it may be slower...

opm/core/props/satfunc/SatFuncMultiplexer.hpp
andlaus: yep. if memory is supposed to be used efficiently, I unfortunately could not find any way around `union` plus `reinterpret_cast` approach...

andlaus: BTW: note that because of the alignment of the union, the GCC warnings are bogus...

opm/core/props/satfunc/SatFuncMultiplexer.hpp
andlaus: probably more like "amiguous". Fixed, thanks.

opm/core/props/satfunc/SatFuncMultiplexer.hpp
andlaus: I'd say it is a good idea if this macro was supposed to be used by external code, but it is internal and gets undefined at the end of this file...

opm/core/props/satfunc/SatFuncMultiplexer.hpp
andlaus: yep. the situation is different if end-point scaling and/or hysteresis also needs to be supported for `Stone1` (seems like EPS is required while hysteresis isn't). The `SatFuncMultiplexer` class does not need to care, though...

CMakeLists_files.cmake
andlaus: right. fixed.

opm/core/simulator/ExplicitArraysFluidState.hpp
andlaus: ok, changed. thanks.

opm/core/simulator/BlackoilStateToFluidState.hpp
andlaus: ok, I added a note which says that the `BlackoilState` object must live longer than the `BlackoilStateToFluidState` one.

opm/core/props/satfunc/SatFuncMultiplexer.hpp
andlaus: > Why have you used this scary union instead of a base class pointer? 

this would probably also work, but would add one level of indirection and thus potentially worse performance and it would not get rid of this pesky `reinterpret_cast` as far as I can see. Probably the performance hit is not measureable, though, so I'll try it and report back...

atgeirr:
> The use of a template parameter to indicate which three-phase model to use has been replaced. Why remove the template?

I now noticed that you did indeed answer this question in a commit message, and I am OK with your reasoning. However the second issue (using explicit switching for the replacement) is still critical I think.

opm/core/props/satfunc/SatFuncMultiplexer.hpp
andlaus: ok, I've implemented it the way you suggested and it turned out to be quite a bit nicer. I'll still have to run it on Norne, but I don't expect it to change anything. Thanks for your suggestion.

andlaus:
> The replacement uses runtime switches supplied (mostly) by a macro. Why use macros instead of common inheritance-based polymorphism?

because inheritance based polymorphism cannot be used to call templated methods of derived classes when only having a base object.

> The (in my opinion) biggest problem with the existing saturation functions (the classes SatFuncGwseg, SatFuncStone2 and SatFuncSimple) is that they started out as duplicates of each other with just a small difference, ripe for refactoring, but instead hysteresis and end-point scaling was added to the gwseg version (and perhaps some was added to the others), and now they are too different to easily refactor. 

agreed. I worked on this as well, but I decided that the PR was becoming quite big even without this, so I wanted to get the fluid state stuff in first. the problem with end point scaling is that it applies to the twophase functions and these cannot be easily nested (if at all) in the current opm-core framework.

> One way out of this might be to just delete the Simple and Stone2 versions, I do not think they are used in any serious manner. Then we would create new ones (at least for Stone2 and Stone1 which we will need soon) from scratch while refactoring to avoid copy-and-paste code.

that's basically what I indend to do next.

> @andlaus, you point out that we should get rid of the combinatorial explosion of variants due to endpoint scaling etc. I agree completely, should we not do that before we implement this API change? Are there any immediate benefits from doing it the other way round?

I think it does not really matter which order is used. first going to fluid states makes it possible to swap in the opm-material implementation (which already has means of dealing with this) without needing big changes, though...

atgeirr:
> because inheritance based polymorphism cannot be used to call templated methods of derived classes when only having a base object.

Of course, I see it now. I am not quite happy with this though, and will think about it a little to see if I can find a better way.

andlaus:
>> because inheritance based polymorphism cannot be used to call templated methods of derived classes when only having a base object.
>
> Of course, I see it now. I am not quite happy with this though, and will think about it a little to see if I can find a better way.

I'd prefer dynamic polymorphism here too, but making the methods of the fluid states and those of the satfuncs virtual probably has a performance impact  and -- more importantly -- makes it harder to use local automatic differentiation instead of plain scalar values. (It's possible to work around this by introducing a template argument for the type scalar values to the class, but this requires quite a bit of code and the result is going to be more inflexible in terms of which kinds of derivatives can be calculated.)

atgeirr:
I have decided to merge this as is. Thanks for the effort, and for updating where requested. I do not see a better way to avoid the multiplexer stuff than you have in the code now, but I think that after refactoring we may see more opportunities.

PR #825, opened 2015-07-03
Revise SimulatorState and BlackoilState.

dr-robertk:
This PR changes the implementation of the SimulatorState and BlackoilState such that all derived classes can add data via the registerCellData or registerFaceData. 

Application to simulator data such as I/O or communication can then be implemented on top of the SimulatorState class only. 

The current behavior of both classes should be as before.

### DIFF
--- a/opm/core/simulator/SimulatorState.cpp
+++ b/opm/core/simulator/SimulatorState.cpp
@@ -50,8 +50,9 @@ SimulatorState::init(int number_of_cells, int number_of_faces, int num_phases)
 {
     num_phases_ = num_phases;
 
-    cellData_.resize( 3 );
-    faceData_.resize( 2 );
+    // clear memory
+    cellData_ = std::vector< std::vector<double> > ( 3 );
+    faceData_ = std::vector< std::vector<double> > ( 2 );
 
     pressure().resize(number_of_cells, 0.0);
     temperature().resize(number_of_cells, 273.15 + 20);
--- a/opm/core/simulator/SimulatorState.cpp
+++ b/opm/core/simulator/SimulatorState.cpp
@@ -2,6 +2,7 @@
 #include <opm/core/grid.h>
 
 #include <cmath>
+#include <cassert>
 
 using namespace Opm;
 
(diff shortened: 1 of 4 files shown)


bska:
I haven't looked at this in any kind of detail--neither the design nor the implementation--so I have not formed any opinion on the PR yet.  I do have one small question though: Is there a reason that the `*Id_` properties (e.g., `pressureId_`) are marked `public:` in class `SimulatorState`?

dr-robertk:
One should probably make those protected. I also think that we should probably introduce a basis class that stores all the data and then derive from that.

dr-robertk:
@bska: The ids are now protected.

opm/core/simulator/SimulatorState.cpp
```diff
@@ -48,19 +49,62 @@ void
...
+size_t
+SimulatorState::registerCellData( const std::string& name, const int components, const double initialValue )
+{
+    // check if init has been called
+    assert( cellData_.size() > 0 );
+    const size_t pos = cellData_.size();
+    cellDataNames_.push_back( name );
+    const size_t nCells = pressure().size();
+    assert( nCells > 0 );
+    cellData_.push_back( std::vector<double>( nCells * components, initialValue ) );
```
atgeirr: Just created a small test to see if this is as efficient as using emplace_back(), and indeed it is just as good, as I would assume. (It would be a little cleaner though, but perhaps not clearer to the reader, so just let it stay as it is.)

atgeirr:
I like this quite a bit, but have a few comments:
 - Why did you not use the registration mechanism for SimulatorState as well? To me it looks like it is unnecessary to hardcode (with static indices) any of the IDs.
 - I'd prefer if there was no default argument for the register...() methods, otherwise it is easy to mistake the component argument for a default value.
 - The IDs of BlackoilState are private while those for SimulatorState are protected. It would be better if all were protected I think.

dr-robertk:
@atgeirr: I notice that we should also use the register for SimulatorState. Then the question is, how should we name the base class, which then becomes the general interface. 
The intialvalue as default parameter seems to be ok with me as it mimics the std::vector behavior. 
The ids are all protected now.

atgeirr:
Two more comments:
 - It is not possible to use the field names for anything. Are you planning on adding a get-field-by-name method?
 - The equals() method could be made *a lot* simpler now.

atgeirr:
> Then the question is, how should we name the base class, which then becomes the general interface.

I assume that you intend to move the basic storage and registration to a base class (which creates no fields), and make SimulatorState just add its fields? If so, it could be called SimulatorStateBase for example (perhaps not SimulatorStateInterface since it provides more than just the interface). Otherwise I think keeping SimulatorState as the base class is ok, it could still set up its fields using register...() methods.

> The intialvalue as default parameter seems to be ok with me as it mimics the std::vector behavior

That I did not quite understand. Are you referring to std::vector<T>(n) initialising n elements equal to T(), and T() being 0.0 for doubles?

dr-robertk:
@atgeirr: The names should be whatever is used in the deck so that you could use them in EclipseWriter etc. 
I think it's (easily) possible to add a get by name method, although that is not all that fast.

atgeirr:
> I think it's (easily) possible to add a get by name method, although that is not all that fast.

I agree that it would not be fast, but it could be used for example in an output code to ask "do you have the field XYZ", and so the output writer could simply receive the base class reference, thereby making the coupling weaker and more runtime-flexible.

(I do not ask for that in this PR, just asking to understand your plans.)

atgeirr:
I think this can be merged as is, just tell me if you want to make any changes first (such as the equals() method or using the registration mechanism for all fields).

dr-robertk:
My idea was the the output writer simply goes through all cellData and faceData and writes them. 
Of course when writing you can check whether the data you are about to write is contained in the initial data set, e.g. 
for all i in cellData 
  if cellDataName( i ) contained in fields 
    write( cellData( i ) 
 endif 
endfor

I hope it's somewhat clear.

dr-robertk:
@atgeirr: Now also SimulatorState is using register{Cell/Face}Data. Form my side this is now good to merge.

atgeirr:
Thanks, I think it is better now, will merge.

PR #828, opened 2015-07-08
Do not use pressure() until after init() in BlackoilState.

atgeirr:
This fixes a test failure after the merging of #825.


bska:

PR #834, opened 2015-07-28
use opm-material for the saturation functions

andlaus:
this PR has been quite a while in the making (mainly due to extensive testing). it replaces the current "saturation function" code by the equivalents in opm-material. note the following points:

bugs in the current opm-core satfuncs
===

the results of `flow` for this branch are not identical the the current `master` version. there are several reasons for this:
- the current saturation functions have a bug in the end point scaling implementation which causes the `scaleSat()` and `scaleSatInv()` methods to not to be their mutual inverses. cf https://github.com/OPM/opm-core/blob/master/opm/core/props/satfunc/SatFuncSimple.cpp#L53 and https://github.com/OPM/opm-core/blob/master/opm/core/props/satfunc/SatFuncSimple.cpp#L70 
- the current hysteresis implementation has a severe bug in that the "saturation shifts" from drainage to imbibition would always be zero if the `S != scaleSatInv(scaleSat(S))` issue would not bite. cf. https://github.com/OPM/opm-core/blob/master/opm/core/props/satfunc/SatFuncBase.hpp#L273 and note that this code would go once in a circle if the bug above was not present
- the saturation-shift-for-imbibition-is-zero problem of the previous point causes a discontinuity in the relperms, and I'll bet a glass of beer with the first guy to accept it that this was the root cause of our stability issues when testing various compilers/flags for the last release
- the saturation scaling of the capillary pressure in the `master` version seems to cause a valgrind complaint about an uninitialized variable used in an if condition. Note that you have to run `flow` in debug mode for the Norne deck to see this, so it might take a while.
- the current opm-core saturation functions only implement relperm hysteresis for the non-wetting phase (is there a reason for this?) while the opm-material code also has relperm hysteresis for the wetting phase. The reason for this is that even though the Eclipse documentation always shows examples using the non-wetting phase, the stuff which it says equally applies to the wetting phase and I have not found any it mentioning that Eclipse only does non-wetting phase relperm hysteresis.
- the hysteresis code in opm-core has another issue via the fact that it uses a different value for the saturation to update the deviation point on the main-drainage curve versus the saturation which it actually uses to calculate the relperms (cf. https://github.com/OPM/opm-core/blob/master/opm/core/props/satfunc/SatFuncGwseg.hpp#L218 vs https://github.com/OPM/opm-core/blob/master/opm/core/props/satfunc/SatFuncBase.hpp#L261 ). this causes another discontinuity. I suspect the reason why this did not totally trash convergence for the Norne case is that the saturation used for updating the point of deflection from the MDC is usually larger than the saturation used to calculate the relperms, i.e. for sufficiently small time steps the drainage curve was always chosen.

comparison of the results
===

the above nonwithstanding, there are branches in my personal opm-material and opm-core repositories which implement a "bug-for-bug" compatibility and compare the results: https://github.com/andlaus/opm-material/tree/satfunc_instrumentation and https://github.com/andlaus/opm-core/tree/satfunc_instrumentation . (note that they must be synchronously as they depend on each other.) using these branches, one gets an exact match (modulo numerical noise) between opm-core and opm-material of the relperm and capillary pressure values as well as their derivatives for SPE1, SPE9 and Norne.

performance
===

- for the SPE1 and SPE9 problems I could not measure any performance difference in `flow` beyond the measurement noise: Some runs are 1-3% faster in the current master version, some are faster by the same margin using the version of this branch
- Norne is about 15% slower with this branch, but that is very likely because hysteresis has not really being used in the current `master` version (see above)


andlaus:
I forgot some minor points:

- with this branch, end point scaling and hysteresis are also available for the `Stone1` and `Stone2` three-phase laws in addition to the default threephase law of Eclipse.
- still to be done is initialization using keywords other than `SWOF` and `SGOF`, but this should not be very hard anymore because there are not many places which are affected by this.
- we should consider to move the `EpsScalingInfo` class to the opm-parser module in order to reduce the number of places in the code which need to access the underlying table keywords. this would make `opm-parser` a hard dependency of `opm-material` if the `Ecl*MaterialLaw`s should be used and it might thus be undesireable.

alfbr:
Thanks for the thorough work!  I am afraid this work needs a lot of effort to be reviewed. I have couple of questions of a more general nature. Why did you make opm-material a dependency instead of including the work directly in opm-core? The performance hit is unfortunate, so I would really like to see that investigated further. Can you provide a more conclusive analysis on that?

dr-robertk:
I think it's a good idea to keep a modular structure. Especially opm-core is already bloated already. 
It the typical "you don't want to buy a cow just because you want to drink a glass of milk" problem. 
For external purpose, such as advertisement, we can still create a super module which should allow us to combine all opm modules needed for flow into one module such that a reservoir engineer only has to install one debian package instead of 10 plus dependencies. For maintenance, however, a modular structure is crucial. And yes, you are right that the dependencies in opm are broken in the sense that a module called core depends on several other modules.

alfbr:
Actually I am getting the cow when I only asked for the milk. I only asked for an improved relperm implementation in opm-core and got a new low level dependency on production code which includes the kitchen sink. Moreover, I fail to see how splitting complex code into repositories makes code more modular or less bloated.

dr-robertk:
Well, splitting code into modules should make it modular, literally. The general idea be to make dependencies optional such that the production code, which is likely to not use the latest and greatest features, is not destabilized by changes like the above.

alfbr:
I agree, but repositories and modules are two different things, so your argument seems misplaced. In this case the opposite happens, the things that can break just increased, as Kai was the first to discover. It may be for a good reason, but then I would like to hear the reason.

joakim-hove:
First of all; impressive work by @andlaus - unfortunately I must say I do not feel competent on reviewing this - however I sincerely hope someone does.

As to the module organization problem - we have had at least two recurring debates in the OPM community:

1. Should we be strict or relaxed/pragmatic when reading input?
2. Should we split the code in many modules/repositories[1] - or should we coalesce to few/one(??) large?

For 1) I hope: https://github.com/OPM/opm-parser/pull/527 will be start of a more fine grained solution - however for 2) that is obviously more difficult. My take on the situation is:

1. The dependencies are currently broken - with `opm-core` as the main culprit; maybe the content of `opm-core` should be split among the remaining modules - and `opm-core` removed completely? Or at the very least  `opm-core` should be renamed. There is no law of nature that the project should have a `core` module; as long `flow` is our main product the module organisation is more like a string of beads than a hub with a core.

2. The death throes in a piece of software sets in when people fear making changes - if you want your software to stay healthy and relevant it must be able to sustain continuous change. For the module structure in opm I feel we should be slightly more agile - willing to experiment a bit. 

   It is impossible for anyone suggesting new dependency structure, like @andlaus is doing here, to guarantee a perfect organisation - and If that is expectation from the rest of the community development will hurt. In my opinion we should not be so scared of changes like this - if it turns out to be a bad idea we can change it again - like we do with the code itself.

Anyway - important and difficult topic!

> [1] @alfbr: but repositories and modules are two different things

Well from a everyday-developer point of view I would say the situation in Opm is that one module feels like one repository and visa versa: 
- Building is in terms of one complete repository.
- Pull requests and reviews are in terms of one complete repository.
- Dependencies are expressed in terms of one complete repository.

alfbr:
> Should we be strict or relaxed/pragmatic when reading input?

I don't think that discussion belongs here, unless it is somehow related.

> Should we split the code in many modules/repositories[1] - or should we coalesce to few/one(??) large?

Generally speaking, I really hope that this discussion never goes away. As the code evolves there will probably always be a need to revisit how we organize it in repositories. There will also probably always be differing views on how large repositories should be.

In this case I suggest being specific. We are talking about unifying saturation functions. opm-core and opm-material had two different implementations. Where should the unified one reside. Both repositories will exist after this PR regardless. It is not at all clear to me that either choice would make the code more or less modular. Is it clear to you or Rob?

There is a third recurring issue that is actually more related to my concern here, and that is bringing in new non-optional dependencies. For the single developer it is as easy as making a change to the cmake file and having the dependency installed. For others it can quickly turn into a very expensive endeavour. On general note I firmly believe that all developers should think twice before pulling in new dependencies in production code, and discuss it before hand to avoid unnecessary tension. I am happy to discuss why at length.

> Well from a everyday-developer point of view I would say the situation in Opm is that one module feels like one repository and visa versa: 

The autodiff based polymer code showed the opposite. The code is now more modular and less bloated by refactoring and joining. For repositories like opm-parser and resinsight you are of course correct, but even for opm-parser it is tricky to find the right split with opm-core. For some of the other repositories I am afraid it is much more complicated.

andlaus:
> Why did you make opm-material a dependency instead of including the work directly in opm-core?

@dr-robertk summarized it quite nicely: I think it is a mistake to continue feeding the monster instead of setting it on diet. (case in point: `opm-{material,parser,cornerpoint}` are quite self contained and there's no principal problem using them for external projects like DuMuX, dune-fem or dune-pdelab. this cannot really be said about `opm-core`.)

> I only asked for an improved relperm implementation in opm-core and got a new low level dependency on production code which includes the kitchen sink.

you should see it this way: you get a modular and flexible rewrite of the buggy old satfuncs which would have been necessary for reasons like API proliferation anyway. Since `opm-material` had quite a bit of infrastructure for this already, adding even more copy-and-pasted code into `opm-core` just to avoid the dependency would be pretty bad style. (On the point if re-implementing this was a good idea: I would never have found the bugs above without it...)

For the usage-in-production argument: I have *very* limited pitty for people who complain that they're sometimes getting a bloody nose because they're using the bleeding-edge development version instead of the latest release ;): doing changes without occasionally breaking stuff is simply *not possible*. releases are done  *precisely* because they represent a well tested state of the code. (and that's also the reason why they should be done fairly frequently IMO: that people do not feel compelled to use some random master version in production.)

> The performance hit is unfortunate, so I would really like to see that investigated further. Can you provide a more conclusive analysis on that?

I will spend some time to analyze it, but you'll probably be interested in the fact that it does not converge at all with the old saturation functions if the "hysteresis shifts" are set to zero. The only reason it converges is the scaling invertion bug. (Thus this a kind of "bug cancellation".)

@joakim-hove
I fully agree with you, and thanks for the compliment :)
 
@alfbr
> There is a third recurring issue that is actually more related to my concern here, and that is bringing in new non-optional dependencies. For the single developer it is as easy as making a change to the cmake file and having the dependency installed.

I understand this concern, but what's the problem with using the packages of the release version or -- if you like to live on the bleeding edge but don't want to be concerned with nitty-gritty things like this -- using a meta-build (and download) tool similar to `dunecontrol`? I thought that we agreed on this on the Trondheim meeting. @akva2: did you do something in this direction? if no, I am sure that I can whip up a simple bash script within a few days that encapsulates most of this stuff...

> The autodiff based polymer code showed the opposite. The code is now more modular and less bloated by refactoring and joining. For repositories like opm-parser and resinsight you are of course correct, but even for opm-parser it is tricky to find the right split with opm-core. For some of the other repositories I am afraid it is much more complicated.

If you're really serious about reducing the number of modules: How about merging the simulator parts of `core` as well as `opm-polymer` into `opm-autodiff`, merging the grid parts from `core` (i.e., the remainder, more or less) into `dune-cornerpoint` and then merging `opm-upscaling` into `opm-porsol`? _that_ would get rid of quite a bit of confusion!

dr-robertk:
I think @joakim-hove and @andlaus gave a nice and comprehensive summary on how we should proceed.  

> It is not at all clear to me that either choice would make the code more or less modular. Is it clear to > you or Rob?
@alfbr: With the arguments presented above I would say it is clear to me. I would put this stuff in opm-material instead of opm-core. Of course both repos will exist afterwards. I think the suggested cleanup strategy from @andlaus sounds very good. We would do ourselfs a big favor if this happens sooner than later. 
And I also agree to the argument that "production code" by definition cannot be the bleeding edge version from the last checkout of the master branch of some repository. If it does, then that code is dead because changes are only allowed under severe restrictions.

joakim-hove:
> I think the suggested cleanup strategy from @andlaus sounds very good.

I also support the strategy outlined by @andlaus. 

>And I also agree to the argument that "production code" by definition cannot be the bleeding edge version from the last checkout of the master branch of some repository. If it does, then that code is dead because changes are only allowed under severe restrictions.

I fully agree with this - and in particular I think the last sentence is very important. If we for various reasons start getting very scared to change things the project will inevitably stagnate. Instead of resisting change we should make certain the safety gear (i.e. automatic tests) is in mint condition and then embrace change. A piece of software which is constantly bent and twisted will stay healthy and relevant, whereas software which is allowed to sit still - maybe grow but not really change, will become less flexible and difficult to work with - just like the human body!

So - not focusing on the particular issue at hand, I think it is important that the Opm development community as a whole embrace change - I feel that @alfbr's arguments have some resemblance to 'fear of change' - and for a development project I really think that is unfortunate. Remember - it is not such a big deal - if the organisation proposed by @andlaus turns out to be suboptimal we can change it again. 

You might disagree and say that changing module/repository organisation *is* a big deal - but in that case I would reply that that is the real problem. If module/repository organisation really is a big deal we have to get better at it - and the only way to get better is by practicing!

alfbr:
Thanks for all your input! Unfortunately most people this change touches are on vacation, so we will need to wait until next week to move further. Just another couple of comments to the discussion from me.

> @andlaus :For the usage-in-production argument: I have very limited pitty for people who complain that they're sometimes getting a bloody nose because they're using the bleeding-edge development version instead of the latest release

I believe there was some miscommunication from my end. The problem is that codes used in production *needs* additional resources to ensure that they work correctly. Somebody will have to pick up the bill for those costs. While we are doing our best, there are clear limits as to how much code we can properly maintain at the moment. We have been struggling with lack of review resources for some time, leading to some developers merging their own changes in production code repositories. Hence for the robustness of OPM we should not throw everything and the kitchen sink into the codes that put bread on the table. We already have multiple repositories where everybody is free to experiment to their hearts content, and the threshold for creating more of them has actually always been very low (still is). Pitty or not, our track record for frequent releases, bugs and testing does not inspire the confidence needed.

>  @joakim-hove :If we for various reasons start getting very scared to change things the project will inevitably stagnate.

I am not afraid of change, if any I should probably be more afraid of change (so I believe you are pushing me in the wrong direction here). I prefer to look at the data, opm-core has many active developers contributing, and I have seen no signs of stagnation. opm-material on the other hand has only one contributor and no review process.

andlaus:
> While we are doing our best, there are clear limits as to how much code we can properly maintain at the moment. [...] Hence for the robustness of OPM we should not throw everything and the kitchen sink into the codes that put bread on the table. 

I agree and I guess  @dr-robertk and @joakim-hove do as well. as I see this, our point rather is that it is infeasible  for developers to make sure that _every_ revision in master is free of regressions. on the other hand, we try very hard to avoid regressions in releases. (this requires a few weeks of testing, though. I guess you see the point.)

> Pitty or not, our track record for frequent releases, bugs and testing does not inspire the confidence needed.

actually, I'm not aware of any regressions in released code so far. which ones did I miss?

> opm-material on the other hand has only one contributor and no review process.

this is more or less correct. I'd like to add that your statement misses a "*so far*", though. If somebody wants the be co-maintainer for opm-material and/or do reviews, please raise your hand!

alfbr:
> actually, I'm not aware of any regressions in released code so far. which ones did I miss?

With the low release frequency we have never been able to stay with a release, so I haven't kept track of regressions. Until we improve testing we will not be able to do more frequent releases. It is a bit of a catch22. Do you believe we are where we should be with respect to frequency of releases, bugs and testing?

andlaus:
> Do you believe we are where we should be with respect to frequency of releases, bugs and testing?

IMO the current state of the art is pretty decent in that respect, but of course things can always be improved -- foremost because some modules desperately need more unit tests. for the 2015.04 release though, I guess that everyone involved agrees that it was far too long in the making. As I see it, the reason for this was the feature-based approach of that release ("We won't release until Norne runs!"), and in retrospect it would have been better to do two intermediate releases for which Norne would only have run partially.

For the future, I suppose that what we are trying to adhere to now (one release every six months) is a good compromise between having time to add new features and testing them. (Also, I suppose the guys who deploy that stuff on clusters -- and the people who employ them -- don't want to do this more frequently.) Personally, I'm open for a quicker release cycle, though.

totto82:
I have tested this PR on Norne and can comfirm that it is 15% slower, even if it uses almost 10% less linear iterations than the current master. Also, for some wells the results deviaties more from Eclipse than before.

andlaus:
> I have tested this PR on Norne and can comfirm that it is 15% slower, even if it uses almost 10% less linear iterations than the current master. Also, for some wells the results deviaties more from Eclipse than before. 

I'm on this, but it may take a few days until I get my final verdict. Currently I suspect that some of these performance issues are caused by associating the derivatives of some of the kinks (which are in the relperms and capillary pressures curves because Eclipse uses piecewise linear interpolators) with the "wrong" side of the curve.

atgeirr:
Note: I am making this comment after reading the above discussion, but I have not yet read the code (review will follow eventually).

1. Module/repository structure. I agree that the current structure is imperfect, there are some good comments in the above discussion about this. I *do* think that modules and repositories should be more or less synonymous, for the reasons outlined by @joakim-hove . A module should be coherent, which means that opm-core as it is now must die. However it is not obvious how to make a replacement module structure, and therefore I raise the question: do we want to make the effort now (meaning before the next release)?

2. Bugs and features in the saturation function codes. I hope @osae can take a very careful look at this. The questions I would like to get answered are: a) How can we currently solve Norne in a satisfactory manner if there are such severe bugs? b) Are they really bugs, or just a different interpretation of how things should be? Does Eclipse have the same bugs? (That sounds extremely strange to me, but we do know that there were bugs with hysteresis and 3-point scaling so it is not completely unthinkable).

3. Testing and results. When running with the new code in "bug compatibility mode" you get the same results as before for Norne. Do you then also get the same performance?

4. Maintenance. If this is merged as is, opm-material enters the "high-value chain" of codes that go into Flow. That means we should add at least one more maintainer to opm-material, as @andlaus suggested above. (I can volunteer for this if necessary.) However it also means that opm-material will be developed using the slower, more bureaucratic approach of pull requests and reviews, compared to its current, single-owner approach. Are you ready for and ok with that @andlaus? Also, do we want to make the rest of the opm-core saturation functions use opm-material? Is that overkill, or will it cause trouble and duplication if we don't do it?

5. Reviewing and merging. I will try to review this and test this a bit before I go ahead and merge. It will probably take some time. Initially I think I will focus on the "bug-compatible" branch, until it is clear what is correct and why.

andlaus:
> 2) Bugs and features in the saturation function codes. I hope @osae can take a very careful look at this. The questions I would like to get answered are: a) How can we currently solve Norne in a satisfactory manner if there are such severe bugs? b) Are they really bugs, or just a different interpretation of how things should be? Does Eclipse have the same bugs? (That sounds extremely strange to me, but we do know that there were bugs with hysteresis and 3-point scaling so it is not completely unthinkable).

@osae and me had some private discussions about it: we now agree that the inversion bug is non-existing (i.e., I misread the code), and that the saturation shifts should also be correct for Norne (but only as long as you don't try to set IMBNUM != SATNUM). Also we agree that wetting phase hysteresis and EPS and hysteresis for any threephase models other than "GwSeg" are not implemented in the current opm-core framework (they are in the opm-material stuff.), and that the saturations used for the hysteresis updates are inconsistent (though it's unclear what Eclipse does as of yet).

> 4) Maintenance. If this is merged as is, opm-material enters the "high-value chain" of codes that go into Flow. That means we should add at least one more maintainer to opm-material, as @andlaus suggested above. (I can volunteer for this if necessary.) However it also means that opm-material will be developed using the slower, more bureaucratic approach of pull requests and reviews, compared to its current, single-owner approach. Are you ready for and ok with that @andlaus?

good, I'm okay with this: If somebody actually is interested in these patches, I'll start to open PRs for opm-material from now on. So far there simply was no point in doing it this way :). also, if you're happy with any PR for opm-material (i.e., the github discussion has settled down and/or you're willing to take the fire), feel free to push Green for opm-material PRs even if you're not an official maintainer!

> Also, do we want to make the rest of the opm-core saturation functions use opm-material? Is that overkill, or will it cause trouble and duplication if we don't do it?

as far as I can see, this is already done by this PR as there won't be any `SatFunc*` classes left in opm-core afterwards. what we could think of in the medium term is also getting rid of the `SaturationPropsFromDeck` class and unifying the PVT relations, though. (But that's not a high priority, IMO.)

> 5) Reviewing and merging. I will try to review this and test this a bit before I go ahead and merge. It will probably take some time. Initially I think I will focus on the "bug-compatible" branch, until it is clear what is correct and why.

ok. I'll also want fix (or at least to find out the reason of) the performance regression and the deviation of the results for Norne, so please hold off with merging a bit.

atgeirr:
> ok. I'll also want fix (or at least to find out the reason of) the performance regression and the deviation of the results for Norne, so please hold off with merging a bit.

Fine, although I assume that these fixes go in opm-material and not opm-core?

alfbr:
For the record, I believe this pull request needs extensive testing, beyond what one maintainer can do. Moreover it touches a part of the code where improved testing framework is needed. If the approach here is lauded during revision, we will probably need to coordinate testing/bugfixing efforts.

andlaus:
> Fine, although I assume that these fixes go in opm-material and not opm-core?

probably. in the process I might find some issues in opm-core, though. (e.g. #838)

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
```diff
@@ -0,0 +1,132 @@
...
+ * uses opm-material's local automatic differentiation framework to allow implicit
+ * treatment of saturation derivatives.
+ */
+class ExplicitArraysSatDerivativesFluidState
+{
+public:
+    enum { numPhases = 3 };
+    enum { numComponents = 3 };
+
+    typedef Opm::LocalAd::Evaluation<double, SaturationDerivativesTag, numPhases> Evaluation;
```
atgeirr: I think the name Evaluation is non-obvious, what about calling it AutoDiffScalar? We could remove the equivalent things from opm-autodiff. I also do not see much benefit from putting it in a nested namespace.

(I know this is really an opm-material issue.)

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <opm/material/localad/Evaluation.hpp>
+#include <opm/material/localad/Math.hpp>
+
+namespace Opm
+{
+
+// this class does not need to be implemented. its only purpose is to make compiler
+// messages for local-AD framework more explicit (because the class name of the
+// Evaluation will include a tag name so that it is clear which derivatives are handled).
+class SaturationDerivativesTag;
```
atgeirr: Evidently it works to only forward-declare the tag, but wouldn't it be better to explicitly declare it as an empty class? If nothing else, a reader would not go looking for the implementation of the class.

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
```diff
@@ -0,0 +1,132 @@
...
+     * index.
+     */
+    void setIndex(unsigned arrayIdx)
+    {
+        arrayIdx_ = arrayIdx;
+
+        // copy the saturations values from the global value. the derivatives do not need
+        // to be modified for these...
+        for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
+            saturation_[phaseIdx].value = globalSaturationArray_[arrayIdx*numPhases + phaseIdx];
```
atgeirr: It looks like this keeps the derivatives unchanged, so I think it is correct (for what it tries to do), but it is perhaps a little unsafe to give such direct access to the value member?

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        // copy the temperature value
+        if (globalTemperatureArray_)
```
atgeirr: The usual nitpick...

(add explicit braces)

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
```diff
@@ -0,0 +1,132 @@
...
+     * This array is supposed to be of size 'size' and is not allowed to be
+     * deleted while the ExplicitArraysFluidState object is alive.
+     */
+    void setTemperatureArray(const double* temperature)
+    { globalTemperatureArray_ = temperature; }
+
+    /*!
+     * \brief Returns the saturation of a phase for the current cell index.
+     */
+    const Evaluation& saturation(int phaseIdx) const
```
atgeirr: I assume that the general idea is that the low-level functions will use the Scalar type of this class, thereby automatically generating the correct derivatives (with respect to saturations)?

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -67,195 +76,16 @@ namespace Opm
...
-                    || deck->hasKeyword("KRORW")
-                    || deck->hasKeyword("KRORG")
-                    || deck->hasKeyword("ENKRVD")
-                    || deck->hasKeyword("IKRG")
-                    || deck->hasKeyword("IKRO")
-                    || deck->hasKeyword("IKRWR")
-                    || deck->hasKeyword("IKRGR")
-                    || deck->hasKeyword("IKRORW")
-                    || deck->hasKeyword("IKRORG") ) {
-                    OPM_THROW(std::runtime_error,"Currently hysteresis and relperm value scaling cannot be combined.");
```
atgeirr: I have not looked at the implementation in opm-material yet, but does it do all the same checks that have been removed from here (not just this line)?

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -67,195 +76,16 @@ namespace Opm
...
 
-
-
-
     /// \return   P, the number of phases.
     inline
     int SaturationPropsFromDeck::numPhases() const
     {
-        return phase_usage_.num_phases;
+        return BlackoilPhases::MaxNumPhases;
```
atgeirr: This means that this class now *only* can work with 3 phases. I think this is moving in the wrong direction, we should rather work to reinstate the (once upon a time working) support for 2-phase runs.

I suspect that the reason for this is the need for Evaluation to know how many derivatives to support, and I see that it should be known at compile time. However, then we should use a template argument or something equivalent to enable us to instantiate either 2 or 3 phases (which could be decided at runtime, the virtual interface is all that is needed downstream).

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -280,31 +110,44 @@ namespace Opm
...
             for (int i = 0; i < n; ++i) {
                 fluidState.setIndex(i);
-                if (do_hyst_) {
-                   satfunc_[cell_to_func_[cells[i]]].evalKrDeriv(fluidState, kr + np*i, dkrds + np*np*i, &(eps_transf_[cells[i]]), &(eps_transf_hyst_[cells[i]]), &(sat_hyst_[cells[i]]));
-                } else if (do_eps_) {
-                   satfunc_[cell_to_func_[cells[i]]].evalKrDeriv(fluidState, kr + np*i, dkrds + np*np*i, &(eps_transf_[cells[i]]));
-                } else {
-                   satfunc_[cell_to_func_[cells[i]]].evalKrDeriv(fluidState, kr + np*i, dkrds + np*np*i);
+                int cartesianCellIdx = cells[i];
+                if (globalCells_)
```
atgeirr: Braces.

atgeirr: On second thought, I think this may be clearer with an arithmetic if:
'''
const int cartesianCellIdx = globalCells_ ? globalCells_[cells[i]] : cells[i];
'''

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -280,31 +110,44 @@ namespace Opm
...
+                    for (int satPhaseIdx = 0; satPhaseIdx < np; ++satPhaseIdx)
+                        dkrds[np*np*i + satPhaseIdx*np + krPhaseIdx] = relativePerms[krPhaseIdx].derivatives[satPhaseIdx];
                 }
             }
         } else {
-// #pragma omp parallel for
+            ExplicitArraysFluidState fluidState;
+            fluidState.setSaturationArray(s);
+
+            double relativePerms[3];
```
atgeirr: You are using enums rather than explicit constants in other places, I think that should be carried through everywhere.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -280,31 +110,44 @@ namespace Opm
...
-                if (do_hyst_) {
-                   satfunc_[cell_to_func_[cells[i]]].evalKrDeriv(fluidState, kr + np*i, dkrds + np*np*i, &(eps_transf_[cells[i]]), &(eps_transf_hyst_[cells[i]]), &(sat_hyst_[cells[i]]));
-                } else if (do_eps_) {
-                   satfunc_[cell_to_func_[cells[i]]].evalKrDeriv(fluidState, kr + np*i, dkrds + np*np*i, &(eps_transf_[cells[i]]));
-                } else {
-                   satfunc_[cell_to_func_[cells[i]]].evalKrDeriv(fluidState, kr + np*i, dkrds + np*np*i);
+                int cartesianCellIdx = cells[i];
+                if (globalCells_)
+                    cartesianCellIdx = globalCells_[cells[i]];
+                const auto& params = materialLawManager_->materialLawParams(cartesianCellIdx);
```
atgeirr: I was a little surprised by this: why do you not compress the arrays and only keep the active data? That would save a little space and a little time, and make the current code a little simpler.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -280,31 +110,44 @@ namespace Opm
...
+            double relativePerms[3];
             for (int i = 0; i < n; ++i) {
-                if (do_hyst_) {
-                   satfunc_[cell_to_func_[cells[i]]].evalKr(fluidState, kr + np*i, &(eps_transf_[cells[i]]), &(eps_transf_hyst_[cells[i]]), &(sat_hyst_[cells[i]]));
-                } else if (do_eps_) {
-                   satfunc_[cell_to_func_[cells[i]]].evalKr(fluidState, kr + np*i, &(eps_transf_[cells[i]]));
-                } else {
-                   satfunc_[cell_to_func_[cells[i]]].evalKr(fluidState, kr + np*i);
+                fluidState.setIndex(i);
+                int cartesianCellIdx = globalCells_?globalCells_[cells[i]]:cells[i];
```
atgeirr: I see here you did exactly what I proposed above (apart from the const and readable [to me] spacing).

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -280,31 +110,44 @@ namespace Opm
...
+
+                const auto& params = materialLawManager_->materialLawParams(cartesianCellIdx);
+                MaterialLaw::relativePermeabilities(relativePerms, params, fluidState);
```
atgeirr: I guess this is directly from Dumux and therefore not good to change, but I do prefer to put output arguments last (the example of memcpy is unfortunate and out of touch with the rest of the world...).

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -332,28 +175,45 @@ namespace Opm
...
-                   satfunc_[cell_to_func_[cells[i]]].evalPcDeriv(fluidState, pc + np*i, dpcds + np*np*i, &(eps_transf_[cells[i]]));
-                } else {
-                   satfunc_[cell_to_func_[cells[i]]].evalPcDeriv(fluidState, pc + np*i, dpcds + np*np*i);
+                int cartesianCellIdx = globalCells_?globalCells_[cells[i]]:cells[i];
+                const auto& params = materialLawManager_->materialLawParams(cartesianCellIdx);
+                MaterialLaw::capillaryPressures(capillaryPressures, params, fluidState);
+
+                // copy the values calculated using opm-material to the target arrays
+                for (int pcPhaseIdx = 0; pcPhaseIdx < np; ++pcPhaseIdx) {
+                    double sign = (pcPhaseIdx == BlackoilPhases::Aqua)?-1.0:1.0;
```
atgeirr: Is this to handle the unfortunate convention that `pcow = po - pw` while `pcog = pg - po`?

atgeirr: Also: spacing please.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -280,31 +110,44 @@ namespace Opm
...
+                    cartesianCellIdx = globalCells_[cells[i]];
+                const auto& params = materialLawManager_->materialLawParams(cartesianCellIdx);
+                MaterialLaw::relativePermeabilities(relativePerms, params, fluidState);
+
+                // copy the values calculated using opm-material to the target arrays
+                for (int krPhaseIdx = 0; krPhaseIdx < np; ++krPhaseIdx) {
+                    kr[np*i + krPhaseIdx] = relativePerms[krPhaseIdx].value;
+
+                    for (int satPhaseIdx = 0; satPhaseIdx < np; ++satPhaseIdx)
+                        dkrds[np*np*i + satPhaseIdx*np + krPhaseIdx] = relativePerms[krPhaseIdx].derivatives[satPhaseIdx];
```
atgeirr: Braces please.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -332,28 +175,45 @@ namespace Opm
...
+                for (int pcPhaseIdx = 0; pcPhaseIdx < np; ++pcPhaseIdx) {
+                    double sign = (pcPhaseIdx == BlackoilPhases::Aqua)?-1.0:1.0;
+                    pc[np*i + pcPhaseIdx] = sign*capillaryPressures[pcPhaseIdx].value;
+
+                    for (int satPhaseIdx = 0; satPhaseIdx < np; ++satPhaseIdx)
+                        dpcds[np*np*i + satPhaseIdx*np + pcPhaseIdx] = sign*capillaryPressures[pcPhaseIdx].derivatives[satPhaseIdx];
                 }
             }
         } else {
-// #pragma omp parallel for
```
atgeirr: Some of my comments to the relperm method apply equally here.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -371,67 +231,27 @@ namespace Opm
...
-                smin[np*cellIdx + p] = satFunc.smin_[p];
-                smax[np*cellIdx + p] = satFunc.smax_[p];
-            }
+        const int np = BlackoilPhases::MaxNumPhases;
+        for (int i = 0; i < n; ++i) {
+            int cartesianCellIdx = (globalCells_)?globalCells_[cells[i]]:cells[i];
+            const auto& scaledDrainageInfo =
+                materialLawManager_->oilWaterScaledEpsInfoDrainage(cartesianCellIdx);
+
+            smin[np*i + BlackoilPhases::Aqua] = scaledDrainageInfo.Swl;
```
atgeirr: I assume that the Swl and Swu are correct to start from here, is there a scaledImbibitionInfo?

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -442,13 +262,16 @@ namespace Opm
...
-        if (do_hyst_) {
-// #pragma omp parallel for
+        if (materialLawManager_->enableHysteresis()) {
+            ExplicitArraysFluidState fluidState;
+            fluidState.setSaturationArray(s);
+
             for (int i = 0; i < n; ++i) {
-                satfunc_[cell_to_func_[cells[i]]].getSatFuncBase().updateSatHyst(s + np*i, &(eps_transf_[cells[i]]), &(eps_transf_hyst_[cells[i]]), &(sat_hyst_[cells[i]]));
+                fluidState.setIndex(i);
+                int cartesianCellIdx = globalCells_?globalCells_[cells[i]]:cells[i];
```
atgeirr: Same issue here as above: write all in the same way, or (better) use active cells.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
```diff
@@ -461,402 +284,9 @@ namespace Opm
...
-            }
-        }
-
-        if (std::fabs(pcmax_tab) < 1.0e-8 || pcmax.empty() || pcmax_tab*pcmax[cell] < 0.0) {
-            data.pcFactor = 1.0;
-        } else {
-            data.pcFactor = pcmax[cell]/pcmax_tab;
-        }
-
+        int cartCellIdx = globalCells_[cell];
```
atgeirr: This is probably a bug on a cartesian grid (you can just write it like the others).

atgeirr:
I have now reviewed one part of this PR: the code in opm-core. Before we merge this I want to
 - Go through the code in opm-material.
 - Get confirmation from @totto82 and @osae that it is correct and performant.
 - Be very sure that I understand the situation with respect to bugs (or not) in the current code, and/or eclipse (if the behaviour we see now is correct, then the new, different, behaviour must be bad, right?).
 - Have consensus about the module issue. No-one has yet replied to my question about doing a reorganisation now or not, please state your opinions!

About the code here, then. With the exception of the use of cartesian indices instead of active indices this looks quite good! I am very happy that you managed to keep using the same (medium-level, SaturationPropsFromDeck) interfaces, I am not saying we should keep them forever, but it makes this PR so much easier to evaluate.

When the minor issues here have been fixed (and the other points above addressed) I'll be happy to merge this.

joakim-hove:
>Have consensus about the module issue. No-one has yet replied to my question about doing reorganisation now or not, please state your opinions!

My general attitude to this is that yes; there is no better time than the current. But more people should voice their opinion - maybe we should have a dedicated hangout? If we intend to have 2015.10 release this should be completed well before that.

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
andlaus: IMO, `FunctionEvaluation` would be a better name, but that's quite a bit to write...

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
andlaus: yep. `saturation_[phaseIdx]' represents an evaluation of the function `S_phaseIdx`, i.e., regardless of the value, its derivative is 1 if it is derived w.r.t. to `S_phaseIdx` and zero else.

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
andlaus: if I understood you correctly, the idea is slightly different: if the partial derivatives are required, then this fluid state class is used if not, then the variant which works on plain scalars (i.e., `ExplicitArraysFluidState`) are passed to the material laws of opm-material.

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
andlaus: fixed.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: modulo bugs, it should do the same for norne.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: > This means that this class now only can work with 3 phases. I think this is moving in the wrong direction, we should rather work to reinstate the (once upon a time working) support for 2-phase runs.

I thought about this, and it can be relatively easily done by adding the material laws to the `EclMultilexerMaterial` class. in the interest of getting this PR done within this decade, I decided to first concentrate on the threephase case, though...

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: changed

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: > Is this to handle the unfortunate convention that pcow = po - pw while pcog = pg - po?

yes. opm-material is consistent here, so pc_alpha,beta = p_alpha - p_beta

> Also: spacing please.

changed something, not sure if you're confident with this.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: all this has been changed.

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
andlaus: yes, it would be possible, but the comment above this declaration is pretty explicit, isn't it?

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: I have done so already, but I need to clean that patch up a bit before I push it.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: I prefer explicit `if`s: if nothing else, people who are not so confident with c++ have less trouble with it. 

Anyway, the if has been made unnecessary because of the move to compressed indices.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: I have problems to get what you mean: the output argument is `relativePerms`.

andlaus: ah, I read you inverted: I suppose that there are different opinions on this: most other standard C functions (e.g. `memset`) also seem to put the output argument in front. personally I don't care.

opm/core/props/satfunc/SaturationPropsFromDeck_impl.hpp
andlaus: that's not needed because SWATINIT seems to only affect the drainage parameters. (at least it does so in the current opm-core implementation.)

Maybe @osae can comment on this?

andlaus:
> My general attitude to this is that yes; there is no better time than the current. But more people should voice their opinion - maybe we should have a dedicated hangout? If we intend to have 2015.10 release this should be completed well before that.

For the record: I think so too. (doing this at least 2 months before a release is a good idea IMO.)

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
atgeirr: Actually, I only understood after a while what you meant by the "need not be implemented", so I think replacing that byt the actual implementation (`{}`) is better. The rest of the comment explains usage nicely and should stay, though there is another effect of the tag if I am not mistaken: it prevents operations between objects with different tags.

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
atgeirr: Maybe I was unclear: are the material laws templates, that can be called with either Evaluation or double arguments?

andlaus: yep. actually they can be also called with any other scalar like `float` or `__float128`. Note that deciding which code-paths are used is done at compile time (mainly by applying the `MathToolbox` concept), and thus these templates are as efficient as implementing all possibilities manually.

opm/core/simulator/ExplicitArraysSatDerivativesFluidState.hpp
andlaus: ok, braces added.

andlaus:
ok, I did some small changes to this last week and now the difference in results seems to be gone. for the full Norne deck, it somehow still exhibits an performance regression of around 6.5% while for the first three time steps it is about 0.5% faster. I strongly suspect that the remaining difference is caused by the fact that the opm-material relperm hysteresis code also does wetting-phase hysteresis because the number of non-linear iterations is reduced from 2453 to 2433, the number of linear iterations goes from 7727 to 8474. I'll disable wetting-phase hysteresis and report back as soon as the run is finished...

totto82:
I have also tested Norne I can confirm that the results are much better now. There are still small deviations in some of the wells. Maybe disabling the wetting-phase hysteresis will fix these?

andlaus:
> Maybe disabling the wetting-phase hysteresis will fix these?

hopefully; I'm running the simulation to check this right now. if it does fix it and the performance issue, I'm wondering if it  is better to bite the bullet and take a small performance hit for the sake of correctness or not. I think it would probably be better to be more correct, because decks other than Norne could be more sensitive...

andlaus:
ok, results are in: restricting hysteresis to the wetting phase, made things worse (12% instead of 7% performance regression compared to master). But I analyzed the timing of the substeps: most of them are actually *faster* with the opm-material satfuncs, but some of them are much worse for some reason. new guess: using the "wrong" derivatives at kinks. I'll investigate it...

atgeirr:
> I'll investigate it...

Good, I have hope that this should result in a net speedup once merged. The substeps that are worse: do they have the same time step size as the base run?

andlaus:
> The substeps that are worse: do they have the same time step size as the base run?

I'm having trouble interpreting the results: for some report steps, one of the versions requires significantly more substeps than the other, for other substeps it is the other way around. overall it looks like the opm-material version draws a blank more often, but I have no idea why this is. On the positive side, if the number of newton iterations are the same for a report step, the opm-material version seems to be about 1% faster...

I'll send you an email with the log files, so you can have a look yourself.

andlaus:
Next results: adding `1e-13` to the saturations before deciding whether to use imbibition or drainage made the performance identical (3662 seconds for the new and 3670 seconds for the old saturation functions). hm, what now?

atgeirr:
> hm, what now?

Are the (displayed) residuals identical, too, then?

Just a question that could (or could not) be relevant: is the extension of tables implemented in the same way as before (even by copying the relevant code)? I am thinking of the SPE9 case for example, which has undersaturated PVTO data only for the last data point.

andlaus:
ok, I finally got the same performance as the master version, but unfortunately the discrepancy to Eclipse is larger than in the master version:

https://poware.org/jaGae8oh/

(note the 'comparison' directory there)

the results are pretty spot-on with the branch of #846 though, so I suspect that either the connate water must be treated differently or that this is caused by the kr regularization. the master version is quite fishy in both aspects, so I'd appreciate some advice by people wiser than me...

andlaus:
ok. I've moved the implementation of `SaturationPropsFromDeck` to a cpp file so that `Evaluation.hpp` does not need the be included before all the other stuff anymore if only `SaturationPropsFromDeck.hpp` is included. I also updated the PRs for the other modules (OPM/opm-autodiff#422 and OPM/opm-polymer#139) to reflect this. Note that the descriptions of these PRs has become obsolete with this.

opm/core/simulator/ExplicitArraysFluidState.hpp
```diff
@@ -47,7 +50,17 @@ class ExplicitArraysFluidState
      * index.
      */
     void setIndex(unsigned arrayIdx)
-    { arrayIdx_ = arrayIdx; }
+    {
+        int np = phaseUsage_.num_phases;
+        for (int phaseIdx = 0; phaseIdx < BlackoilPhases::MaxNumPhases; ++phaseIdx) {
+            if (phaseUsage_.phase_used[phaseIdx] < 0) {
```
atgeirr: This looks like a bug, it's supposed to be used as a boolean flag.

opm/core/simulator/ExplicitArraysFluidState.hpp
```diff
@@ -47,7 +50,17 @@ class ExplicitArraysFluidState
...
+                sats_[phaseIdx] = 0.0;
+            }
+            else {
+                sats_[phaseIdx] = saturations_[np*arrayIdx + phaseUsage_.phase_pos[phaseIdx]];
```
atgeirr: So the sats_ object will always have 3 elements, even for a two-phase simulation? I assume that is what the opm-material code expects, since that is what the saturation(int phaseIndex) function below gives? I think we need to separate the concepts of "canonical/possible phases" (in other parts of the code: Water, Oil, Gas) from "active phases in the current simulations" here (which can be dynamic). If the code is correct, and the opm-material functions indeed always expect oil saturation in the second slot (even if there is no water phase in the simulation), I think at least a comment should be made for the saturation() method. This also applies to any other FluidState objects.

opm/core/simulator/ExplicitArraysFluidState.hpp
andlaus: yeah, correct. fixed & thanks.

lacking decks to test the twophase stuff, there are probably more gotchas like this in the code...

atgeirr:
It looks like one of my questions went unanswered: do the opm-material functions always expect "canonical" positions for phase saturations (water, oil, gas) even if only two phases are used?

Second, test_satfunc now fails.

Third, I cannot run a two-phase case anymore (I get "Program threw an exception: No valid saturation keyword family specified"), but this may not be the fault of this PR.

Fourth, I have run Norne to completion, and while I have not checked the output it seemed to run as well as before.

atgeirr:
> Third, I cannot run a two-phase case anymore (I get "Program threw an exception: No valid saturation keyword family specified"), but this may not be the fault of this PR.

Actually, it is... The checks being done in EclEpsScalingPoints.hpp lines 193-194 do not account for a two-phase situation. This must be fixed before I can merge this PR.

andlaus:
> It looks like one of my questions went unanswered: do the opm-material functions always expect "canonical" positions for phase saturations (water, oil, gas) even if only two phases are used?

yes, the phase indices must be determined at compile time. but there may be "gaps" in the phases which are used, so there is a patch in OPM/opm-material#44 which implements twophase material laws that way.

> Second, test_satfunc now fails.

confirmed. if you look at the source of that test though, it the failiures are exclusively caused by the relperm derivatives, and it also looks like the reference data which the test compares the results with has been calculated using the old satfuncs, i.e., the unit test checks if the values _change_, but it does not check if they are _correct_? anyway I've added a patch to 'fix' this.

andlaus:
> Actually, it is... The checks being done in EclEpsScalingPoints.hpp lines 193-194 do not account for a two-phase situation. This must be fixed before I can merge this PR.

please send me the deck. I only have SPE10 from opm-data and this cannot be loaded for different reasons...

atgeirr:
> the unit test checks if the values change, but it does not check if they are correct?

We should figure that out before we let them be changed... This test served its purpose, then, to alert us to changes. I hope you can verify that the differences are due to different roundoff etc., but to me it looked more serious (at least to a quick look), with some differences that were quite large.

andlaus:
the values were always the same, and since it is difficult to get the derivatives wrong in the new implementation, I guess that they were wrong in the old one. I suspect that the major culprit for this is that the old  implementation does not consider `min()` and `max()` of saturations in the derivatives.

atgeirr:
> since it is difficult to get the derivatives wrong in the new implementation, I guess that they were wrong in the old one

If you could verify this for at least one single data-point I would rest much easier...

andlaus:
afaics, there were some derivatives which were the same. if that's what you mean, I can verify this...

andlaus:
ok, consider for example the derivative of the oil relperm w.r.t. the water saturation: the old implementation says that this -2 for the first two saturations (which are probably below the connate water saturation):

https://github.com/andlaus/opm-core/commit/2d5852dd460269c5b51bdad26b773f4b2aba8b62#diff-baa1f63f41bb5622afbbf223d2a472e9L189

this is a very strong hint that a `max()` is not considered when calculating the derivatives. the new implementation says that the derivatives are 0:

https://github.com/andlaus/opm-core/commit/2d5852dd460269c5b51bdad26b773f4b2aba8b62#diff-baa1f63f41bb5622afbbf223d2a472e9R189

a way to decide what's correct is to use finite differences and see which derivative is closer, but given that the first three values are identical, I'd bet one crate of beer that the outcome of the FDs is zero.

andlaus:
ok, I rebased this and OPM/opm-autodiff#422 on top of the current master to solve the conflicts with the recent TableManager change.

atgeirr:
So I'll merge this, except: are you working on OPM/opm-autodiff#422 and OPM/opm-polymer#139 to eliminate the "globalcell" code duplication issue reported there?

I am willing to merge all of it now, and get that fix later, but if you have this just around the corner, tell me so and I'll hold a little.

andlaus:
> So I'll merge this, except: are you working on OPM/opm-autodiff#422 and OPM/opm-polymer#139 to eliminate the "globalcell" code duplication issue reported there?

sorry, I forgot about this yesterday. will do it in the afternoon.

andlaus:
the opm-autodiff and opm-polymer PRs now feature this function.

atgeirr:
Hold your horses, here we go!

PR #836, opened 2015-08-03
OPM-218: Write restart files at same interval as eclipse

chflo:
Added substep as a parameter to writeTimeStep to avoid writing adaptive timesteps for restart files.
Write of restart files does not fully support the writing of substeps, and for now leave out the writing of these to write correct restart files with the same dates as eclipse. 
When this PR is merged, PR for opm-218 for autodiff must also be merged


joakim-hove:
Minor update here: https://github.com/OPM/opm-core/pull/837

PR #837, opened 2015-08-04
Opm 218

joakim-hove:
See: https://github.com/OPM/opm-core/pull/836

Added one commit with test data update.


joakim-hove:
There are currently three PR's which must be merged syncronously:

1. https://github.com/OPM/opm-parser/pull/533
2. https://github.com/OPM/opm-core/pull/837
3. https://github.com/OPM/opm-autodiff/pull/424

I have tested this - and I am happy to merge the trio; unless someone screams up I will merge them tomorrow afternoon.

opm/core/io/OutputWriter.hpp
```diff
@@ -88,7 +88,8 @@ class OutputWriter {
      */
     virtual void writeTimeStep(const SimulatorTimerInterface& timer,
                                const SimulatorState& reservoirState,
-                               const WellState& wellState) = 0;
+                               const WellState& wellState,
+                               bool  substep = false) = 0;
```
atgeirr: This (default arguments in virtual functions) should in general be avoided if possible. One reason is that the default value actually used will be the one of the static type of the argument (not the dynamic type), which can be surprising, although it probably would not matter for this particular case.

I am usually ok with default values when adding arguments to existing functions, to avoid large rewrites, but would it take much effort to remove it here?

joakim-hove: I agree with you - frankly I would take it a bit further: "*lazy defaults to avoid rewrites quite often come back to bite*". Will update.

joakim-hove: OK - default argument to `writeTimeStep()` has been removed - is now a regular argument.

PR #838, opened 2015-08-05
properly initialize the attributes of the SatHyst class

andlaus:
this caused the hysteresis saturation shifts to be initially wrong and
was probably the source of a lot of my confusion w.r.t. the saturation
scaling and hysteresis shift code. I guess this also fixes a valgrind
complaint, but I haven't checked...

I propose this PR as a precursor to #834 because it makes the comparison stuff easier because it fixes a bug in the current code.

### DIFF
--- a/opm/core/props/satfunc/SatFuncBase.hpp
+++ b/opm/core/props/satfunc/SatFuncBase.hpp
@@ -65,6 +65,14 @@ namespace Opm
     
     // Hysteresis
     struct SatHyst {
+        SatHyst() {
+            sg_hyst = -1.0;
+            sow_hyst = -1.0;
+
+            sg_shift = 0.0;
+            sow_shift = 0.0;
+        }
+
         double sg_hyst;
         double sg_shift;
         double sow_hyst;


atgeirr:
This looks innocent enough. I assume this changes the results on Norne, that we get slightly better results now? Or at least better reproducability?

opm/core/props/satfunc/SatFuncBase.hpp
```diff
@@ -65,6 +65,14 @@ namespace Opm
     
     // Hysteresis
     struct SatHyst {
+        SatHyst() {
+            sg_hyst = - 1.0;
```
atgeirr: I actually find
```
sg_hyst = -1.0
```
easier to read (it is a literal so it should be kept together).

andlaus:
I'd be very surprised if this would change the results in a significant way: because newly allocated memory is usually initialized to 0, the SatHyst objects become initialized as soon as the non-wetting saturation changes to the s*_hyst value (which is 0 because of this fact). i.e., after a few time steps the saturation shifts should be correct for most cells. if the uninitialized memory holds some random numbers, it would be a different story. (but in this case we probably would not have gotten a match of the results.)

atgeirr:
Kai and I have now tried this with Norne, and we cannot observe any differences.

opm/core/props/satfunc/SatFuncBase.hpp
andlaus: changed.

atgeirr:
I had to read up on initialisation a bit (http://en.cppreference.com/w/cpp/language/value_initialization), but judging from that, I believe that in all versions of C++ (03, 11, 14) the statement
```
sat_hyst_.resize(n);
```
(where `sat_hyst` is a vector of SatHyst) will have all zero-initialised members, guaranteed (not just by accident since newly allocated memory often is zero). So the change made by this PR would be to make two of the members -1 instead of 0, which should not matter since it is compared to actual saturation values, presumably in [0, 1]. Do you agree?

I still think the change is worthwhile, since it clearly marks the values as unused.

andlaus:
> (where sat_hyst is a vector of SatHyst) will have all zero-initialised members, guaranteed (not just by accident since newly allocated memory often is zero). 

I'd be surprised if it was that way because I suppose that `std::vector` just calls the default constructor which did not do anything so far.

> So the change made by this PR would be to make two of the members -1 instead of 0, which should not matter since it is compared to actual saturation values, presumably in [0, 1]. Do you agree?

That's actually the point of this PR: setting 'sg_hyst' and 'sow_hyst' to zero is wrong because the imbibition shifts will initially not be updated if the initial solutions set the gas or the oil saturation to zero. (i.e., these shifts can initially be zero even if this is wrong).

atgeirr:
> I'd be surprised if it was that way because I suppose that std::vector just calls the default constructor which did not do anything so far.

Actually, it does not quite do that. Read the link I pasted above, to see what it does.

> That's actually the point of this PR: setting 'sg_hyst' and 'sow_hyst' to zero is wrong because the imbibition shifts will initially not be updated if the initial solutions set the gas or the oil saturation to zero. (i.e., these shifts can initially be zero even if this is wrong).

But if this is the case, why do we not observe a difference on Norne?

andlaus:
> Actually, it does not quite do that. Read the link I pasted above, to see what it does.

if this is true, my c++-rather-lets-you-shoot-yourself-into-the-foot-than-loose-a-smige-of-performance believe just got a (slight) dent ;)

> But if this is the case, why do we not observe a difference on Norne?

as I wrote above, that's because as soon as the relevant saturations go to non-zero values, the `SatHyst` objects are updated and thus correct thereafter.

andlaus:
>>    But if this is the case, why do we not observe a difference on Norne?
>
> as I wrote above, that's because as soon as the relevant saturations go to non-zero values, the SatHyst objects are updated and thus correct thereafter.

also, thinking about it: for a non-wetting saturation the drainage curve is always chosen, so that the imbibition shifts are not used. that said, I'm pretty sure that it makes valgrind complain and it also makes my job of comparing the saturation functions harder than necessary, and if you get the idea that you want to store SOG instead of SG or SW instead of SOW you'll have a hard encounter with the pavement if you do not initialize the attributes explicitly...

andlaus:
can this be merged, please? it is not performance critical because the time spend in this part of the initialization is negligible.

PR #839, opened 2015-08-06
Define the UF_long type if necessary.

atgeirr:
The UF_long type is deprecated in UMFPack, and actually seems to be removed from the latest versions.

In the long term, we should switch to using SuiteSparse_long instead of UF_long, but that would increase the minimum version required to 4.0.0 (I think). The proposed solution defines a macro only in the .c file where it is used.

### DIFF
--- a/opm/core/linalg/call_umfpack.c
+++ b/opm/core/linalg/call_umfpack.c
@@ -42,6 +42,11 @@
 #include <opm/core/linalg/sparse_sys.h>
 #include <opm/core/linalg/call_umfpack.h>
 
+/* To handle new versions of SuiteSparse. */
+#ifndef UF_long
+#define UF_long SuiteSparse_long
+#endif
+
 struct CSCMatrix {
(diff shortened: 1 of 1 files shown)


atgeirr:
(Note: comments should be made in the diff, not the commit to be better readable from the PR page, they will disappear upon rebasing for example.)

I did it in that way since that is the way SuiteSparse already does things. I would never do it like this in new C++ code, but this is an older C code, and I duplicated the approach used to avoid errors. If replacing it with a typedef is just as good here that is OK with me, should I change it? I guess I still need the #ifndef, will that also work for things that are not macros?

atgeirr:
I looked into this a little, and `#ifndef UF_long` works as intended because `UF_long` is defined by a macro, not a typedef, in older versions of SuiteSparse. I think therefore that the proposed patch (using a macro instead of a typedef) is better since it is consistent with older versions.

PR #844, opened 2015-08-10
Vertical flow performance tables

babrodtk:
This sets up for using vertical flow performace tables for production and injection wells (i.e., tubing head pressure control). 

Most of the actual VFP code resides in opm-autodiff, but this should compile and run without merging that first.

### DIFF
--- a/opm/core/wells.h
+++ b/opm/core/wells.h
@@ -212,6 +212,8 @@ add_well(enum WellType  type     ,
  *
  * \param[in] type       Control type.
  * \param[in] target     Target value for the control.
+ * \param[in] alq        Artificial lift quantity for control (for THP type only)
+ * \param[in] vfp        VFP table number for control (for THP type only)
  * \param[in] distr      Array of size W->number_of_phases or NULL.
  * \param[in] well_index Index of well to receive additional control.
  * \param[in,out] W  Existing set of well controls.
@@ -222,6 +224,8 @@ add_well(enum WellType  type     ,
 int
 append_well_controls(enum WellControlType type  ,
                      double               target,
+                     double               alq,
+                     int                  vfp,
                      const double        *distr,
                      int                  well_index,
                      struct Wells        *W);
--- a/opm/core/wells/WellsGroup.cpp
+++ b/opm/core/wells/WellsGroup.cpp
@@ -26,6 +26,12 @@
 #include <memory>
 #include <iostream>
 
+namespace
+{
+    static double invalid_alq = -1e100;
+    static double invalid_vfp = -2147483647;
+} //Namespace
+
 namespace Opm
 {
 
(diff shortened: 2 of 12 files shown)


opm/core/wells.h
```diff
@@ -222,6 +224,8 @@ add_well(enum WellType  type     ,
 int
 append_well_controls(enum WellControlType type  ,
```
atgeirr: I know we cannot overload in C, but would it not be good to create a new method, append_well_controls_thp() that takes the extra arguments?

opm/core/wells/WellsGroup.cpp
```diff
@@ -921,13 +931,14 @@ namespace Opm
 
         if (group_control_index_ < 0) {
             // The well only had its own controls, no group controls.
-            append_well_controls(wct, ntarget, distr, self_index_, wells_);
+            append_well_controls(wct, ntarget, -1e100, -1e100, distr, self_index_, wells_);
```
atgeirr: Here you use two invalid doubles, but one argument is an int.

opm/core/wells/WellsManager.cpp
```diff
@@ -452,6 +452,8 @@ namespace Opm
 
                     ok = append_well_controls(SURFACE_RATE,
                                               injectionProperties.surfaceInjectionRate, 
+                                              -std::numeric_limits<int>::max(),
+                                              -std::numeric_limits<int>::max(),
```
atgeirr: Here you use two int arguments, but one should be a double. Consider using named invalid constants.

atgeirr: Note that this kind of uglyness could be eliminated by adding a separate function as suggested above (at the cost of checking specifically for that case in some places).

opm/core/wells.h
babrodtk: We discussed this back and forth a bit, and the reason for keeping it this way was that other parameters already sent to this function were "special". With that in mind, this extension of the existing method follows the existing pattern. 

If you think it's better to have it as a separate function, that will surely be possible

atgeirr: You are right, we can keep it as it is proposed here. The ideal would be perhaps be to have a set of functions such as append_bhp_control(), append_thp_control() etc., but we are not going to do a major refactoring of the controls code at this point since it may be replaced with use of opm-parser:Well directly in the future.

I would ask for readable constants for invalid things then.

babrodtk: Sounds like a good idea. Perhaps these constants could belong together with the logger and error macros under opm-common?

atgeirr: Perhaps we could do that, but for now I would just like something like invalid_alq or invalid_vfp_table defined locally.

atgeirr:
This cannot be merged without also changing opm-autodiff, since the well_controls_add_new() function is called from SimulatorBase_impl.hpp in that module. If those changes can be put in its own PR we may merge them first, otherwise (and probably the best) we have to merge this together with OPM/opm-autodiff#430.

babrodtk:
I think this (and OPM/opm-autodiff#430) can be merged now, can you take a look at it @atgeirr ?

atgeirr:
Looks good to a cursory overview, I'll merge this and the opm-autodiff PR now. Thanks for your effort!

PR #845, opened 2015-08-10
Fix abs bug and warnings

atgeirr:
This fixes a bug in an internal utility that affected the grid_equal() function (it would give false positives).

Also, many warnings have been silenced. (The bug was discovered during the warning-squashing process as a compiler warning was given for the buggy behaviour.)

### DIFF
--- a/opm/core/grid/grid.c
+++ b/opm/core/grid/grid.c
@@ -25,6 +25,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <math.h>
 
 
 void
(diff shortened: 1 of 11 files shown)


opm/core/grid/grid.c
```diff
@@ -550,14 +551,15 @@ static int memcmp_double(const double * p1 , const double *p2 , size_t num_eleme
...
     else {
-        const double epsilon = 1e-5;
+        const double abs_epsilon = 1e-8;
+        const double rel_epsilon = 1e-5;
 
         for (size_t index = 0; index < num_elements; index++) {
-            double diff = abs(p1[index] - p2[index]);
-            if (diff != 0) {
-                double sum = abs(p1[index]) + abs(p2[index]);
+            double diff = fabs(p1[index] - p2[index]);
```
bska: This is wrong.  I know the change (`abs()` to `fabs()`) looks appropriate, but there are some subtleties that you're missing.  You should read the link referenced in the comment.

atgeirr: Why is this wrong? We are not doing the ULPs here, just a straightforward comparison. The original code did not use the ULPs either (no casting or union as far as I can see).

bska: I tried to do something like this before and it broke `make_test` on
````C
grid_equal(G, G)
```

atgeirr: The proposed patch does not break test_ug.cpp, which does contain
```
    BOOST_CHECK( grid_equal( cgrid1 , cgrid1 ));
    BOOST_CHECK( grid_equal( cgrid2 , cgrid2 ));
    BOOST_CHECK( !grid_equal( cgrid1 , cgrid2 ));
```

I am curious what you meant by the abs() comment though, did you refer to the ULP-counting technique?

dr-robertk:
I agree with @atgeirr that fabs should be used when dealing with double, since this is a C function, and not a C++ function (where abs would be the right choice). Other than that, comparing doubles like in line 557 is wrong to my opinion. I don't see a good reason not to merge this PR.

@bska:  Do you have an easy example where you can show that the suggestions from @atgeirr fail?

joakim-hove:
Now - I am not suggesting this: https://github.com/joakim-hove/opm-core/tree/fix-abs-bug-and-warnings be merged - but as far as I can see it confirms that the approach suggested by @atgeirr is fine - and should be merged.

atgeirr:
Ok, will merge.

PR #846, opened 2015-08-11
Small gwseg fixes

andlaus:
these two changes should not change the results obtained by 'SatFuncGwseg' significantly, but they make it much easier to match its behaviour for saturations where the relperms get regularized.

with this, the full Norne deck runs through even if the results of the saturation functions are compared with the opm-material version (#834)


atgeirr:
I do not understand this sentence:

> with this, the full Norne deck runs through even if the results of the saturation functions are compared with the opm-material version

What do you mean here?

Also, I would appreciate if @osae or @totto82 could review and check that this indeed does not change Norne results. (Unless I was mistaken and this actually changes them?)

andlaus:
> What do you mean here?

this means that the Norne deck with the full schedule runs through if you use the satfunc_instrumentation branches:

https://github.com/andlaus/opm-material/tree/satfunc_instrumentation
https://github.com/andlaus/opm-core/tree/satfunc_instrumentation
https://github.com/andlaus/opm-autodiff/tree/satfunc_instrumentation

for autodiff, this branch is identical to the generic "satfunc_refactoring" of the PR, for opm-material the difference to the master version is tiny (it adds some hack to detect if a curve is evaluated close to a kink and does the hysteresis update the same way as opm-core does) and the opm-core branch is the union of the current master version plus the "satfunc_refactoring" branch from the PR with making sure that the first ten digits of the values and the derivatives are identical (but if the curve is evaluated close to a kink, it skips the comparison of the derivatives). in these branches, the simulation `abort()`s if any discrepancy is detected.

atgeirr:
> making sure that the first ten digits of the values and the derivatives are identical

Are you saying that in these branches you evaluate all properties through both the old opm-core path and the new opm-material path, and verify that they are essentially identical?

andlaus:
> Are you saying that in these branches you evaluate all properties through both the old opm-core path and the new opm-material path, and verify that they are essentially identical?

yes. But I slowly get the feeling that I've got bugs there...

totto82:
I have tried this PR on norne and the results are worse than without it compared to Eclipse. See also comments from @andlaus in OPM/opm-core#834

andlaus:
ok, I digged into this a bit: the difference is caused by the second patch (i.e., the connate water treatment). the first one does not seem to make any difference.

opm/core/props/satfunc/SatFuncGwseg.hpp
```diff
@@ -438,18 +444,15 @@ namespace Opm
...
             const double sw = std::max(fluidState.saturation(BlackoilPhases::Aqua), swco);
             const double sg = fluidState.saturation(BlackoilPhases::Vapour);
-            const double eps = 1e-6;
-            swco = std::min(swco,sw-eps);
             const double ssw = sg + sw;
-            const double ssg = std::max(sg + sw - swco, eps);
-            const double d = ssg; // = sw - swco + sg (using 'd' for consistency with mrst docs).
 
             double ssow = 1.0-ssw;
-            double ssog = 1.0-ssg-swco;
```
totto82: It should be ssog = 1.0-sg-sw
The same must be changed in opm-material to get match with master.

andlaus: thanks, with this it looks like all results match. interestingly, the changes of the scaling points on line 501 and 501 seem like they do not influence the results... (and the code there looked *very* fishy before IMO.)

andlaus: I'll report back on the performance as soon as it is finished.

andlaus: ok, new results: the summary files for Norne are now not discernable anymore, but performace of the opm-material version is worse because the number of linear and non-linear iterations is larger. Now I "only" have to figure out why.

the results can be found here: https://poware.org/jaGae8oh/

andlaus:
ok, the connate water change was incorrect for decks which had a connate water saturation > 0 in unscaled space. I've thus removed the patch.

I've also added a patch which subtracts 1e-13 before deciding if imbibition or drainage is used. (this gets deals with numerical noise when comparing.)

andlaus:
ok. this PR does not change much, and it is not much benefit of merging it before the ones using opm-material for the satfuncs. closing.

PR #847, opened 2015-08-12
Fix abs bug and warnings

joakim-hove:
This PR should **not** be merged - it is only a convenient demonstration.


atgeirr:
Is it because it makes the memcmp function public you don't want it merged? Because it is better to have tests for it than not.

joakim-hove:
> Is it because it makes the memcmp function public you don't want it merged?

Well yes; so before someone hit me in the head with a clue-stick I was very upfront about not merging. But yes - I agree that having a test is always better than not having one; I could move it to e.g `opm/core/utility/memcmp_double.h`?

atgeirr:
Sounds good to me!

joakim-hove:
OK - micro PR here: https://github.com/OPM/opm-core/pull/849

PR #848, opened 2015-08-12
Fix bugs related to ExplicitArraysFluidState

atgeirr:
This makes the two-phase simulators work again. It could have a negative effect on #834, if that requires the number of phases to be known at compile time. If so, we need to discuss how to work that out.


atgeirr:
Merging this as it fixes a critical error for the two-phase sims.

PR #849, opened 2015-08-12
Added small utility memcmp_double()

joakim-hove:
Very small utility for comparing two C double * vectors.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -130,6 +130,7 @@ list (APPEND MAIN_SOURCE_FILES
 	opm/core/utility/VelocityInterpolation.cpp
 	opm/core/utility/WachspressCoord.cpp
 	opm/core/utility/miscUtilities.cpp
+       opm/core/utility/memcmp_double.c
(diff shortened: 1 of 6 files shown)


atgeirr:
Looks fine, merging.

bska:
According to the C standard we (users) are not allowed to introduce any symbols (i.e., functions) with external linkage whose names begin with `mem*`--at least if we're using a hosted implementation (which we are).  Those symbols are in the implementations's namespace to allow for future expansion of the standard.

joakim-hove:
> According to the C standard we (users) are not allowed to introduce any symbols (i.e., functions) with external linkage whose names begin with mem*

Alriiight - that was certainly new to me; will come with an update tomorrow.

atgeirr:
OK, let's rename it. I suggest `opm_memcmp_double`.

bska:
>> According to the C standard we (users) are not allowed to introduce any symbols (i.e., functions) with external linkage whose names begin with `mem*`
>
> Alriiight - that was certainly new to me;

It's one of those details that escape notice for years.  The same holds for functions named `str*` and a few other names that I have to look up.

PR #851, opened 2015-08-17
Catch more istl exceptions

atgeirr:
This can be thrown by the ILU0 preconditioner when using the interleaved solver approach. Catching it allows us to shorten timesteps etc. as we do with other kinds of linear solver trouble.

This makes it possible to run through Norne with the interleaved solver.


GitPaean:
Having been using the implementation suggested in the PR for about two weeks. It has been working pretty well.

atgeirr:
Since other maintainers are more or less unavailable, and this is a very minor change I'll self-merge.

PR #852, opened 2015-08-17
adding command line option full_timestep_initially_

GitPaean:
To set its value to be `true`, the grammar is `full_timestep_initially=true`. 

When the option is true, for each report step, the size of the adaptive time step always beginning with the size of the report step.

The implementation of the current master branch is to use the average time step from the running of the previous report step. Under some conditions, it can result in relative small time step for the subsequent report steps. 

A preliminary test with all the other options defaulted, `full_timestep_initially_=true` reduce the number of Newton iterations from 2585 (current master branch with dune2.3) to 2143. And total running time from 5504 seconds to 4867 seconds. 

Before we decide to make this option to be `true` by default, it is `false` by default at the moment.


atgeirr:
Thanks, will merge.

PR #854, opened 2015-08-20
Opm 215 EclipseReader restart file read of solution data

chflo:
Extended EclipseReader with read of temperature, pressure, sgas and swat from restart file. 
Added unit test.
This code is not in use yet, it will be when Flow supports restarts. 

This PR is not dependent upon other PRs.


joakim-hove:
This is a required component to support restarting based Eclipse formatted restart files, but currently there are no consumers of this code. Merging.

PR #857, opened 2015-08-24
Threshold pressure

iLoop2:
Dependent on : https://github.com/OPM/opm-parser/pull/562.

### DIFF
--- a/opm/core/utility/thresholdPressures.hpp
+++ b/opm/core/utility/thresholdPressures.hpp
@@ -49,14 +49,7 @@ namespace Opm
     std::vector<double> thresholdPressures(const ParseMode& parseMode ,EclipseStateConstPtr eclipseState, const Grid& grid)
     {
         SimulationConfigConstPtr simulationConfig = eclipseState->getSimulationConfig();
-        const std::vector<double>& thresholdPressureTable;
-        if(simulationConfig->hasThresholdPressure()){
-           thresholdPressureTable = simulationConfig->getThresholdPressureTable();
-        }else{
-            std::string msg = "Inferring threshold pressure from the initial state is not supported.";
-            parseMode.handleError( ParseMode::UNSUPPORTED_INITIAL_THPRES , msg );
-        }
-
+        const std::vector<std::pair<bool, double>>& thresholdPressureTable = simulationConfig->getThresholdPressureTable();
         std::vector<double> thpres_vals;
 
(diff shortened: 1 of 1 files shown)


bska:
> merged commit 9ad6e59 into OPM:master 3 minutes ago

That was (a bit too) quick.  I'd really preferred if the branch were rebased on top of current master to remove the needless merge in the middle of the commit series.  On a technical note, some of the explicit types should probably be
```C++
const auto& Object = expression;
```
in light of our decision to use full C++11.  This is in the interest of resilience with respect to future API changes.  I've seen that objects of type `std::vector<std::pair<T,U>>` tend to be subject to change more frequently than those that are based on `std::vector<T>` for some `T` (built-in or user defined).

joakim-hove:
>That was (a bit too) quick.

I agree - sorry; to the extent it is possible I will push updates.

>I'd really preferred if the branch were rebased on top of current master to remove the needless merge in the middle of the commit series.

Is this seriously to late now? I have never really understood the "Revert" functionality of git/github.

> On a technical note, some of the explicit types should probably be [...]

Now I am all for the use of `auto` - but as long as the internals of the method use `item.first` and `item.second` - i.e. explicitly accessing the properties of a `std::pair<>` I fail to see that this will give any resilience towards API changes? But I'd be happy to push an update using `auto`.

bska:
>> I'd really preferred if the branch were rebased on top of current master to remove the needless merge in the middle of the commit series.
>
> Is this seriously too late now? 

Yes.  We never rebase once something is published in master (i.e., in official, public project history).

>  I have never really understood the "Revert" functionality of git/github.

In this context, "revert", means "add commit (or commits)" to reverse/undo **effect of** previous commits.  Revert does not change or remove any existing commits so those remain part of the history indefinitely.

> as long as the internals of the method use `item.first` and `item.second` - i.e. explicitly accessing the properties of a `std::pair<>` I fail to see that this will give any resilience towards API changes

My argument in this case is, admittedly, weak.  I suppose one could say that using `auto` means there are fewer locations that require explicit intervention in the case of an API change, but `auto` does not remove the maintenance cost here.  That said, I am slowly inching towards the point of view that `std::pair<>` is almost never the correct choice for a vocabulary type (i.e., something used or needed in public interface of class).  One is usually better off defining a custom class (or `struct`) and giving the fields meaningful names.  In this case I wouldn't know the semantic relation between `i.first` and `i.second` without the RM readily available.

PR #858, opened 2015-08-24
Modify region mapping to keep empty regions.

osae:
This seems to be the most convenient way to maintain consistency between the EQUIL records and the region mapping give by EQLNUM. (In "Model 2" EQLNUM refers to several regions that are empty when
ACTNUM is accounted for.)

### DIFF
--- a/opm/core/utility/RegionMapping.hpp
+++ b/opm/core/utility/RegionMapping.hpp
@@ -171,7 +171,7 @@ namespace Opm
                 const std::pair<CI,CI>
                     m = std::minmax_element(reg.begin(), reg.end());
 
-                low  = *m.first;
+                low  = 0;
 
                 const typename Region::size_type
                     n = *m.second - low + 1;
--- a/opm/core/simulator/initStateEquil.hpp
+++ b/opm/core/simulator/initStateEquil.hpp
@@ -404,6 +404,8 @@ namespace Opm
                          r < nr; ++r)
                     {
                         const typename RMap::CellRange cells = reg.cells(r);
+                        if (cells.size() == 0)
+                            continue;
 
                         const int repcell = *cells.begin();
                         const RhoCalc calc(props, repcell);
--- a/tests/test_regionmapping.cpp
+++ b/tests/test_regionmapping.cpp
@@ -49,8 +49,8 @@ BOOST_AUTO_TEST_CASE (RegionMapping)
     for (size_t i = 0; i < regions.size(); ++i) {
         BOOST_CHECK_EQUAL(rm.region(i), regions[i]);
     }
-    std::vector<int> region_ids = { 2, 3, 4, 5, 6, 7 };
-    std::vector< std::vector<int> > region_cells = {  { 0, 2, 4 },  { 7 },  { 3 },  { 1 },  { 6, 8 },  { 5 }  };
+    std::vector<int> region_ids = { 0, 1, 2, 3, 4, 5, 6, 7 };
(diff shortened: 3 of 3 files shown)


opm/core/utility/RegionMapping.hpp
```diff
@@ -171,7 +171,7 @@ namespace Opm
                 const std::pair<CI,CI>
                     m = std::minmax_element(reg.begin(), reg.end());
 
-                low  = *m.first;
+                low  = 0;
```
bska: Don't do that.  If this is really needed, then larger rewrites are needed too.  Supporting an arbitrary lower bound was one of the design restrictions for class `RegionMapping`.

osae:
OK, admit that this PR borders to vandalism...
Will it be acceptable to include this as an optional behavior?  Something like:
    RegionMapping(const Region& reg, bool keep_leading_empty_regions=false)
    
Note that the current implementation and use of RegionMapping seems inconsistent for non-zero "low":
    // In autodiff/RateConverter.hpp and core/simulator/initStateEguil.hpp:
   for (typename RMap::RegionId r = 0, nr = m.numRegions(); r < nr; ++r)
   { ... m.cells(r) ... }
This seems to cause a negative index for the p-vector: rev_.p[r-rev_.low]

atgeirr:
> Will it be acceptable to include this as an optional behavior? Something like:
> RegionMapping(const Region& reg, bool keep_leading_empty_regions=false)

I think this looks like a pragmatic solution. The proposal would not change any existing behaviour. Longer term, we might have to redo the RegionMapping to work in such a way that empty regions are not considered (currently if the regions 2, 3, 6 and 7 have active cells we would get numRegions() == 6, not 4).

joakim-hove:
I am not really into this - but my 0.02$ is:

1. Optional behavior usually comes back to bite you.
2. In the parser region values are first class values - whereas they are (ab)used as indices in the `RegionMapping` implementation.

So - I would strongly recommend 'doing it properly' - rather than hacking around a problem with optional behavior. It will come back.

bska:
> would strongly recommend 'doing it properly'

Agreed.  I still don't understand what problem the proposed change is supposed to fix.

atgeirr:
> This seems to cause a negative index for the p-vector: rev_.p[r-rev_.low]

It is clear that there is confusion between this class and its uses: I agree that it looks like it cannot really be working for non-zero low. What is it we want/need from this class, anyway? Do we want to refer to regions by original number? Or do we want to refer to them from 0..(number of regions - 1)? I prefer the second, but that is not what is done by this class (but it is expected in the places it is used). We may additionally need to have the original numbers available for each region. I'll make a proposal.

joakim-hove:
>Do we want to refer to regions by original number? Or do we want to refer to them from 0..(number of regions - 1)?

There is no hard and fast guarantee that the region numbers start at 1 - I have encountered both 0 (for  inactive cells 'which are ignored anyway') and minimum values greater than 1.

bska:
The intention of class `RegionMapping` is (was) to provide a reverse lookup from (active) regions to active cells of that region.  In particular, the input argument to `RegionMapping<>::cells()` is supposed to be a region identifier such as EQUILNUM, SATNUM, &c.  At the moment we're missing a level of indirection that generally identifies the non-empty regions.  We could add a method (and backing data) to support a method
```C++
RegionRange activeRegions() const;
```
(with `RegionRange` **intentionally** not specified, except that it would have to support `std::begin()` and `std::end()`).  Then we could do something like
```C++
for (const auto& reg : rmap.activeRegions()) {
    const auto& c = rmap.cells(reg);
    use(c);
}
```
if we needed to loop the active regions.

PR #859, opened 2015-08-25
Use estimated time-step instead of average in next report step

totto82:
The last estimated time-step is used for the initial sub-step in the report step instead of the average of the time-steps used in the previous report step.
 . 
Testes: 
Norne: 3.7% speedup
SPE1: (run using: flow SPE1DECK.DATA): 37% speedup
SPE9: (run using: flow SPE9.DATA) 15% speedup.


atgeirr:
Good!

I think also the name `last_timestep` is misleading. Would `recommended_timestep` or something like that be better?

totto82:
Renaming done.

atgeirr:
Thanks, merging!

PR #860, opened 2015-08-25
Using map base ThresholdPressure implementation.

joakim-hove:
Followup to: https://github.com/OPM/opm-parser/pull/565

### DIFF
--- a/opm/core/utility/thresholdPressures.hpp
+++ b/opm/core/utility/thresholdPressures.hpp
@@ -17,12 +17,14 @@
   along with OPM.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#include <vector>
-#include <opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.hpp>
-#include <opm/parser/eclipse/Parser/ParseMode.hpp>
-
 #ifndef OPM_THRESHOLDPRESSURES_HEADER_INCLUDED
 #define OPM_THRESHOLDPRESSURES_HEADER_INCLUDED
+
+#include <vector>
+#include <opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.hpp>
+#include <opm/parser/eclipse/EclipseState/SimulationConfig/ThresholdPressure.hpp>
+#include <opm/parser/eclipse/Parser/ParseMode.hpp>
+
 
 namespace Opm
 {
@@ -49,14 +51,11 @@ namespace Opm
     std::vector<double> thresholdPressures(const ParseMode& parseMode ,EclipseStateConstPtr eclipseState, const Grid& grid)
     {
         SimulationConfigConstPtr simulationConfig = eclipseState->getSimulationConfig();
-        const std::vector<std::pair<bool, double>>& thresholdPressureTable = simulationConfig->getThresholdPressureTable();
         std::vector<double> thpres_vals;
-
-        if (thresholdPressureTable.size() > 0) {
-
+        if (simulationConfig->hasThresholdPressure()) {
+            std::shared_ptr<const ThresholdPressure> thresholdPressure = simulationConfig->getThresholdPressure();
             std::shared_ptr<GridProperty<int>> eqlnum = eclipseState->getIntGridProperty("EQLNUM");
             auto eqlnumData = eqlnum->getData();
-            int maxEqlnum = *std::max_element(eqlnumData.begin(), eqlnumData.end());
 
             // Set values for each face.
             const int num_faces = UgGridHelpers::numFaces(grid);
@@ -70,14 +69,18 @@ namespace Opm
                     // Boundary face, skip this.
                     continue;
                 }
-                const int eq1 = eqlnumData[gc[c1]] - 1;
-                const int eq2 = eqlnumData[gc[c2]] - 1;
-                std::pair<bool, double> value = thresholdPressureTable[eq1 + maxEqlnum*eq2];
-                if(value.first){
-                    thpres_vals[face] = value.second;
-                }else{
-                    std::string msg = "Inferring threshold pressure from the initial state is not supported.";
-                    parseMode.handleError( ParseMode::UNSUPPORTED_INITIAL_THPRES , msg );
+                const int gc1 = (gc == 0) ? c1 : gc[c1];
+                const int gc2 = (gc == 0) ? c2 : gc[c2];
+                const int eq1 = eqlnumData[gc1];
+                const int eq2 = eqlnumData[gc2];
+
+                if (thresholdPressure->hasRegionBarrier(eq1,eq2)) {
+                    if (thresholdPressure->hasThresholdPressure(eq1,eq2)) {
+                        thpres_vals[face] = thresholdPressure->getThresholdPressure(eq1,eq2);
+                    } else {
+                        std::string msg = "Initializing the THPRES pressure values from the initial state is not supported - using 0.0";
+                        parseMode.handleError( ParseMode::UNSUPPORTED_INITIAL_THPRES , msg );
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


andlaus:
what's the advantage of this? better performance?

opm/core/utility/thresholdPressures.hpp
```diff
@@ -70,14 +70,16 @@ namespace Opm
...
-                const int eq1 = eqlnumData[gc[c1]] - 1;
-                const int eq2 = eqlnumData[gc[c2]] - 1;
-                std::pair<bool, double> value = thresholdPressureTable[eq1 + maxEqlnum*eq2];
-                if(value.first){
-                    thpres_vals[face] = value.second;
-                }else{
-                    std::string msg = "Inferring threshold pressure from the initial state is not supported.";
-                    parseMode.handleError( ParseMode::UNSUPPORTED_INITIAL_THPRES , msg );
+                const int eq1 = eqlnumData[gc[c1]];
+                const int eq2 = eqlnumData[gc[c2]];
```
bska: Since you're already changing these lines, can you add some special casing to the `gc` expressions too?  We support `gc == nullptr` to mean the identity mapping so we really should say something along the lines of
```C++
const int gc1 = (gc == 0) ? c1 : gc[c1];
const int gc2 = (gc == 0) ? c2 : gc[c2];

const int eq1 = eqlnumData[gc1];
const int eq2 = eqlnumData[gc2];
```

opm/core/utility/thresholdPressures.hpp
```diff
@@ -49,10 +49,10 @@ namespace Opm
     std::vector<double> thresholdPressures(const ParseMode& parseMode ,EclipseStateConstPtr eclipseState, const Grid& grid)
     {
         SimulationConfigConstPtr simulationConfig = eclipseState->getSimulationConfig();
-        const std::vector<std::pair<bool, double>>& thresholdPressureTable = simulationConfig->getThresholdPressureTable();
+        const std::shared_ptr<ThresholdPressure> thresholdPressure = simulationConfig->getThresholdPressure();
         std::vector<double> thpres_vals;
 
-        if (thresholdPressureTable.size() > 0) {
+        if (thresholdPressure->size() > 0) {
```
atgeirr: I find this a bit strange to ask: would it be better to ask `hasThresholdPressure` or `isActive`?

opm/core/utility/thresholdPressures.hpp
atgeirr: I assume this means that the opm-parser methods deals with region numbers from 1..n? Is that universal across APIs?

opm/core/utility/thresholdPressures.hpp
```diff
@@ -49,14 +51,13 @@ namespace Opm
     std::vector<double> thresholdPressures(const ParseMode& parseMode ,EclipseStateConstPtr eclipseState, const Grid& grid)
     {
         SimulationConfigConstPtr simulationConfig = eclipseState->getSimulationConfig();
-        const std::vector<std::pair<bool, double>>& thresholdPressureTable = simulationConfig->getThresholdPressureTable();
+        std::shared_ptr<const ThresholdPressure> thresholdPressure = simulationConfig->getThresholdPressure();
         std::vector<double> thpres_vals;
 
-        if (thresholdPressureTable.size() > 0) {
+        if (thresholdPressure->size() > 0) {
```
atgeirr: I don't get why my comment became outdated, so again: would not `isActive` be a better query than the size? We do not use the size since we don't access the table explicitly.

atgeirr:
I realized one thing while looking at this: are we handling this correctly if there are nncs? Or: should nncs be subject to THPRES?

joakim-hove:
Thank you for the review comments - I will push updates.

joakim-hove:
> what's the advantage of this? better performance?

Simpler and clearer implementation - IMO.

joakim-hove:
>would not isActive be a better query than the size? We do not use the size since we don't access the table explicitly.

Absolutely - stay tuned.

joakim-hove:
> I assume this means that the opm-parser methods deals with region numbers from 1..n? 

Ahhh tricky; my take on this is that region numbers are *values* just like the string 'YES' and the floating point number 2.71 - so we should not upfront force the values into 0...n-1 range; it *is* not an index per se.

> Is that universal across APIs?

The other examples I am aware of are the `xxxxREG` keywords which apply a modifier to all cells in a region, and the `MULTREGT` keyword. For those keywords the implementation is as here - i.e. the region values are used as is. But I guess the `ThresholdPressure` implementation is the first point where that decision has leaked out of opm-parser.

I guess my argument could be extended to apply to i,j,k values as well - but there dual use of i,j,k as values and as indices into some linear storage is firmly entrenched that I would *never* suggest something like it for those values.

joakim-hove:
> I realized one thing while looking at this: are we handling this correctly if there are nncs? Or: should nncs be subject to THPRES?

The current implementation goes through all faces - so I guess that should be correct even in the case of NNCs?

joakim-hove:
OK; this must be merged synchronously with https://github.com/OPM/opm-parser/pull/565. Please indicate when/if you are ready to merge this and I will squash the last two commits.

opm/core/utility/thresholdPressures.hpp
```diff
@@ -49,14 +51,10 @@ namespace Opm
     std::vector<double> thresholdPressures(const ParseMode& parseMode ,EclipseStateConstPtr eclipseState, const Grid& grid)
     {
         SimulationConfigConstPtr simulationConfig = eclipseState->getSimulationConfig();
-        const std::vector<std::pair<bool, double>>& thresholdPressureTable = simulationConfig->getThresholdPressureTable();
-        std::vector<double> thpres_vals;
-
-        if (thresholdPressureTable.size() > 0) {
+        if (simultaionConfig.hasThresholdPressure()) {
```
bska: This looks like a misprint (simulTAion vs. simulATion).

bska: Also, we should probably use `->` rather than `.` on a `SimulationConfigConstPtr`.

joakim-hove: > This looks like a misprint (simulTAion vs. simulATion).

Holy diver.

joakim-hove: Hold on.

atgeirr:
> The current implementation goes through all faces - so I guess that should be correct even in the case of NNCs?

I don't think so, since NNCs are separate from grid faces (a decision I resisted a long time, but that was impossible to avoid). For example when building the discrete div and grad operators, we use both grid topology and nncs to build the correct operators. I guess that transitioning to using a more abstract graph (rather than a grid) after both grid and nncs have been processed would make this a little simpler, @bska has suggested this before.

joakim-hove:
When making changes to `thresholdPressure.hpp` you must recompile `flow` in opm-autodiff to actually get your changes compiled - that is just broken. For the future I insist that header only code should at least have a test which ensures that it is compiled.

I actually tried to put my money where my mouth is by adding such a test - but that test can not even compile because the `thresholdPressure.hpp` implementation requires the GridHelpers from opm-autodiff - come on guys!

bska:
> That test can not even compile because the `thresholdPressure.hpp` implementation requires the `GridHelpers` from opm-autodiff - come on guys!

I suppose the easiest, short-term way around that is to move the implementation of threshold pressures to opm-autodiff.  Long term this is (yet) another reason to restructure the prerequisite graph among the modules.

joakim-hove:
No further comments - on this; I interpret that as "Go ahead and merge it?"

bska:
> I interpret that as "Go ahead and merge it?"

Fine by me.  You **may** want to hold off until the currently running Jenkins builds have completed, though.

bska:
I said
> You **may** want to hold off until the currently running Jenkins builds have completed, though.

The build bots are quiet.  Go ahead on the merge.

PR #861, opened 2015-08-27
Zero-porosity cells and MINPV.

osae:
Model 2 partly uses PORO=0.0 to signal inactive cells. This PR
prevents such cells from causing unwanted zcorn modifications in
the current min-pv algoritm. (The problem is however more general as
zero PORO or NTG should kill a cell whether MINPV is active or not,
refer kw ACTNUM.)


atgeirr:
I think this looks reasonable, it will probably become obsolete when the new MINPV-processing is done, but until then this is an acceptable stopgap I think, so I'll merge.

The more general problem should also be addressed. Can you make an issue of it?

PR #862, opened 2015-08-27
Cell center depth also from zcorn average.

osae:
There is a sligth difference between the cell depth taken from the
centroid (opm) and a straightforward average of corner coordinates
(ecl). For the initialisation of Norne this causes some differences
when deciding cell center positions relative contact depths.

### DIFF
--- a/opm/core/grid/GridHelpers.cpp
+++ b/opm/core/grid/GridHelpers.cpp
@@ -70,6 +70,7 @@ double cellCenterDepth(const UnstructuredGrid& grid, int cell_index)
     // Traverse the bottom and top cell-face
     for (int i=grid.cell_facepos[cell_index+1]-2; i<grid.cell_facepos[cell_index+1]; ++i) {
         // Traverse the vertices associated with each face
+        assert(grid.face_nodepos[grid.cell_faces[i]+1] - grid.face_nodepos[grid.cell_faces[i]] == nv/2);
         for (int j=grid.face_nodepos[grid.cell_faces[i]]; j<grid.face_nodepos[grid.cell_faces[i]+1]; ++j) {
             zz += (grid.node_coordinates+nd*(grid.face_nodes[j]))[nd-1];
         }
--- a/opm/core/grid/GridHelpers.cpp
+++ b/opm/core/grid/GridHelpers.cpp
@@ -57,6 +57,17 @@ const double* beginCellCentroids(const UnstructuredGrid& grid)
     return grid.cell_centroids;
 }
 
+double cellCenterDepth(const UnstructuredGrid& grid, int cell_index) 
+{
+    double zz = 0.0;
+    for (int i=grid.cell_facepos[cell_index+1]-2; i<grid.cell_facepos[cell_index+1]; ++i) {
+        for (int j=grid.face_nodepos[grid.cell_faces[i]]; j<grid.face_nodepos[grid.cell_faces[i]+1]; ++j) {
+            zz += (grid.node_coordinates+dimensions(grid)*(grid.face_nodes[j]))[2];
+        }
+    }
+    return zz/8.0;
+}
+
 double cellCentroidCoordinate(const UnstructuredGrid& grid, int cell_index,
                                  int coordinate)
 {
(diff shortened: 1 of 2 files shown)


opm/core/grid/GridHelpers.cpp
```diff
@@ -57,6 +57,17 @@ const double* beginCellCentroids(const UnstructuredGrid& grid)
     return grid.cell_centroids;
 }
 
+double cellCenterDepth(const UnstructuredGrid& grid, int cell_index) 
+{
+    double zz = 0.0;
+    for (int i=grid.cell_facepos[cell_index+1]-2; i<grid.cell_facepos[cell_index+1]; ++i) {
```
atgeirr: This uses the convention that the last two faces for a cell are the bottom and top faces, and should say so in a comment (otherwise it is hard to understand).

opm/core/grid/GridHelpers.cpp
```diff
@@ -57,6 +57,17 @@ const double* beginCellCentroids(const UnstructuredGrid& grid)
     return grid.cell_centroids;
 }
 
+double cellCenterDepth(const UnstructuredGrid& grid, int cell_index)
```
atgeirr: This method only works for UnstructuredGrid, to be applicable to the parallel version of flow it must also work with CpGrid, and should use the "GridHelpers" functional grid API.

opm/core/grid/GridHelpers.cpp
```diff
@@ -57,6 +57,17 @@ const double* beginCellCentroids(const UnstructuredGrid& grid)
     return grid.cell_centroids;
 }
 
+double cellCenterDepth(const UnstructuredGrid& grid, int cell_index) 
+{
+    double zz = 0.0;
+    for (int i=grid.cell_facepos[cell_index+1]-2; i<grid.cell_facepos[cell_index+1]; ++i) {
+        for (int j=grid.face_nodepos[grid.cell_faces[i]]; j<grid.face_nodepos[grid.cell_faces[i]+1]; ++j) {
+            zz += (grid.node_coordinates+dimensions(grid)*(grid.face_nodes[j]))[2];
```
atgeirr: This of course is very specific to three-dimensional grids, so I think it should assert() that is indeed the case.

opm/core/grid/GridHelpers.cpp
atgeirr: I noticed now that the PR indeed extends the functional API... Sorry for misunderstanding. However, a cellCenterDepth() must also be provided for CpGrid, and it could be easier to make a (templated) single function rather than two specific overloads. Up to you.

osae:
This PR is associated with PR dune-cornerpoint #168.

opm/core/grid/GridHelpers.cpp
```diff
@@ -57,6 +57,26 @@ const double* beginCellCentroids(const UnstructuredGrid& grid)
...
+    // For cornerpoint grids, this is likely to give slightly different depths that seem
+    // to agree with eclipse.
+    assert(grid.dimensions == 3);
+    const int nd = 3; // Assuming 3-dimensional grid ...
+    const int nv = 8; // Assuming 2*4 vertices ...
+    double zz = 0.0;
+    // Traverse the bottom and top cell-face
+    for (int i=grid.cell_facepos[cell_index+1]-2; i<grid.cell_facepos[cell_index+1]; ++i) {
+        // Traverse the vertices associated with each face
+        for (int j=grid.face_nodepos[grid.cell_faces[i]]; j<grid.face_nodepos[grid.cell_faces[i]+1]; ++j) {
```
atgeirr: I think you should assert() that there are four corners here (or nv/2), since the code depends on it (it should be true, but to guard against future changes breaking things).

atgeirr:
Looks good, apart from the assert() I want (above) I'll merge this when the build test succeeds.

osae:
OK, I'll put in the assert.

atgeirr:
The multi-pr jenkins test seem to have hit a snag not related to this, so I'll merge.

PR #863, opened 2015-08-28
Makes setupCompressedToCartesian a free function.

blattms:
This lets us reuse this code at various places. E.g.
I need this functionality in dune-cornerpoint to create
a graph representing the well completions connect the cells.

LWe should probably move this function out of the
WellsManager as it has nothing to do with it. Feel free to propse a header and source file.

### DIFF
--- a/opm/core/wells/WellsManager.cpp
+++ b/opm/core/wells/WellsManager.cpp
@@ -391,8 +391,8 @@ namespace Opm
         well_collection_.applyExplicitReinjectionControls(well_reservoirrates_phase, well_surfacerates_phase);
     }
 
-    void WellsManager::setupCompressedToCartesian(const int* global_cell, int number_of_cells,
-                                                  std::map<int,int>& cartesian_to_compressed ) {
+    void setupCompressedToCartesian(const int* global_cell, int number_of_cells,
+                                    std::map<int,int>& cartesian_to_compressed ) {
         // global_cell is a map from compressed cells to Cartesian grid cells.
         // We must make the inverse lookup.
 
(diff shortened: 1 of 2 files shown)


atgeirr:
Good to make it a free function, I suggest moving it to setupCompressedToCartesian.hpp/cpp, in the opm/core/grid/ directory.

bska:
> Good to make it a free function,

I don't agree.  This function is extremely poorly designed and should really be moved to `/dev/null`.  We certainly should not expose it to further use outside the (moribund) `WellsManager`.

atgeirr:
> This function is extremely poorly designed

Do we have better implementations of the same that can be used easily?

bska:
>> This function is extremely poorly designed
>
> Do we have better implementations of the same that can be used easily?

Not at the moment, but an immediate improvement would be to replace the `std::map<>` with a `std::vector<int>` of size "number of global cells".

blattms:
@bska I don't agree. In a sequential run that might make sense with few deactivated cells. In a parallel run it wil definitely waste a lot of memory.

bska:
> In a sequential run that might make sense with few deactivated cells. In a parallel run it will definitely waste a lot of memory.

It still does not make sense to chase `O(log ngc)` pointers only to look up a single integer.  That's a cache miss nightmare.  I suppose we **could** replace it by `std::unordered_map<>`, but I'd also consider a sorted `std::vector<std::pair<int,int>>` of size "number of active cells on processor" and bisect (`std::lower_bound()`) my way to associativity on top of it.

blattms:
I am closing this for now until we come up with a better solution.

PR #864, opened 2015-09-01
Improve RegionMapping and fix related bugs.

atgeirr:
After this, the RegionMapping can handle arbitrary empty regions, and the correct EQUIL records are used by the initEquil system.

This is proposed to fix the issues described in the discussion of #858.


bska:
I have given this some thought and I'm inclined to reject it.

This provides a strict subset of the features I suggested in my earlier [comment](https://github.com/OPM/opm-core/pull/858#issuecomment-136387762) and is furthermore no easier to use in client code.  Add to that that the [`compress()`](https://github.com/atgeirr/opm-core/blob/fix-regionmapping/opm/core/utility/RegionMapping.hpp#L222) method jars with the overall terse style of the class and I'd consider this to be a breaking change for little benefit.

In short, I oppose this proposal and won't merge it as is but if someone else wants to merge it and take on the maintenance then feel free.

atgeirr:
> This provides a strict subset of the features I suggested in my earlier comment and is furthermore no easier to use in client code. Add to that that the compress() method jars with the overall terse style of the class and I'd consider this to be a breaking change for little benefit.

The current proposal can be rewritten to support an iterator interface like you suggested, but that would take a little more effort. I think a traditional "provide a size and access by index" interface is ok for this, but I acknowledge there might be a risk of confusion between these indices in [0..(numRegions()-1)] and the region ids (although the risk is less than before, with the explicit regionId() method and the documentation). I'd appreciate further opinions on this to settle it.

About the compress() method: I cannot see that terseness in itself is a virtue, the main intention was to avoid rewriting the whole init() method, which indeed is very terse. Statements like
```
c[ p[ reg[i] - low + 1 ] ++ ] = i;
```
have a certain beauty, but unless well-explained the terseness works against the understanding of the code.

bska:
> Statements like
```C++
c[ p[ reg[i] - low + 1 ] ++ ] = i;
```

are the shortest possible way of achieving the desired effect (insert `i`--a cell index--into the bin identified by `reg[i]` and prepare for next insertion by incrementing the associate end pointer).  Splitting it up just for the sake of splitting it up or introducing a work array to play the role of `p` is wasteful and useless.  I expect readers to know the language and context.  The only wart is the `+ 1`, but that's required by construction (using end pointers rather than start pointers).

babrodtk:
I truly believe this deserves two lines of code, as its starting go look like a good ol' Perl one-liner...
Would not an optimizing compiler provide just as efficient code for the (in my opinion) much more readable
```
int& k = p[ reg[i] - low + 1 ];
c[ k++ ] = i;
```
?

bska:
@babrodtk
> Would not an optimizing compiler provide just as efficient code for the (in my opinion) much more readable
```C++
int& k = p[ reg[i] - low + 1 ];
c[ k++ ] = i;
```

Yeah, I suppose.  I adopted that convention when I implemented (well, wrote the code, didn't compile it or  update the tests or client code) my [`active-regions`](https://github.com/OPM/opm-core/compare/master...bska:active-regions?expand=1) branch.  The diff-stat is
```
git diff -w --stat master..active-regions 
opm/core/utility/RegionMapping.hpp | 123 ++++++++++++++++---------------------
1 file changed, 52 insertions(+), 71 deletions(-)
```
with the savings being from replacing the `CellRange` class with a `boost::iterator_range<>`.

atgeirr:
While I love a good style discussion (and acknowledge that I started it), the most important point to discuss is the interface of RegionMapping. I took a look at your active-regions branch @bska and I think the proposed interface is ok, although it requires follow-up changes in opm-autodiff (unlike this PR).

atgeirr:
Closing as we'll be merging #872.

PR #867, opened 2015-09-02
Fixed several warnings

babrodtk:
This fixes several warnings in core about shadowed variables / functions. Should have no other side effects


babrodtk:
Rebased changes so that it's possible to merge this PR.

atgeirr:
Thanks, merging.

(I do think it is a little annoying that -Wshadow triggers on these silly things, because we do not want to disable it for fear of missing important cases. We could reduce it by adopting the underscore_style for variables and camelCase style for functions though, which happens to be my favourite...)

PR #869, opened 2015-09-03
Fix complaint about initial declaration in foor loops.

blattms:
gcc-4.7.2 (Debian 4.7.2-5) complained about:
```
'for’ loop initial declarations are only allowed in C99 mode
note: use option -std=c99 or -std=gnu99 to compile your code"
```
when seeing somethink like this:
```c
for(int i=0; i<end; i++)
```
This is fixed by moving the declaration before the for loop with
this commit. Altenatively, we could use the above option.

### DIFF
--- a/opm/core/wells/wells.c
+++ b/opm/core/wells/wells.c
@@ -544,7 +544,8 @@ wells_equal(const struct Wells *W1, const struct Wells *W2 , bool verbose)
         return are_equal;
     }
 
-    for (int i=0; i<W1->number_of_wells; i++) {
+    int i;
+    for (i=0; i<W1->number_of_wells; i++) {
         if (are_equal) {
             /*
               The name attribute can be NULL. The comparison is as
--- a/opm/core/utility/memcmp_double.c
+++ b/opm/core/utility/memcmp_double.c
@@ -14,8 +14,8 @@ int memcmp_double(const double * p1 , const double *p2 , size_t num_elements) {
     else {
         const double abs_epsilon = 1e-8;
         const double rel_epsilon = 1e-5;
-
-        for (size_t index = 0; index < num_elements; index++) {
+        size_t index;
+        for (index = 0; index < num_elements; index++) {
             double diff = fabs(p1[index] - p2[index]);
             if (diff > abs_epsilon) {
                 double sum = fabs(p1[index]) + fabs(p2[index]);
--- a/opm/core/wells/well_controls.c
+++ b/opm/core/wells/well_controls.c
@@ -364,7 +364,8 @@ well_controls_get_current_distr(const struct WellControls * ctrl) {
 void 
 well_controls_iset_distr(const struct WellControls * ctrl, int control_index, const double * distr) {
     int offset = control_index * ctrl->number_of_phases;
-    for (int p=0; p < ctrl->number_of_phases; p++)
+    int p;
+    for (p=0; p < ctrl->number_of_phases; p++)
         ctrl->distr[offset + p] = distr[p];
 }
 


opm/core/wells/wells.c
```diff
@@ -544,7 +544,8 @@ wells_equal(const struct Wells *W1, const struct Wells *W2 , bool verbose)
         return are_equal;
     }
 
-    for (int i=0; i<W1->number_of_wells; i++) {
+    int i;
```
bska: If we want to have valid C89 then this variable should be moved to the start of the function and we should return `int` rather than `bool`.

bska:
I'm not at all opposed to using C89 in opm-core, in fact I'd prefer it if we were targeting that standard, but the ERT requires a C99 compiler so it's also not unreasonable to require that users switch to C99 mode with a compiler flag.

atgeirr:
Could we not put -std=c99 in our build system, similar to what we do with c++?

joakim-hove:
> Could we not put -std=c99 in our build system, similar to what we do with c++?

I think that is a good idea; however this: 

> ERT requires a C99 compiler 

Is only partly true:

1. Yes - the ERT build system specifies C99 and the complete ERT application requires a C99 compiler.
2. The Windows build system used by Resinsight only supports C89, so I try hard to guarantee that the part of ERT which is used by Resinsight/OPM compiles cleanly with C89. 

Personally I would like to get away from this unofficial C89 support - and would suggest using `-std=c99` also in OPM, but if the community makes a conscious decision to use C89 I guess we will do the same for ERT.

opm/core/wells/wells.c
blattms: I was kind of surprised that the compiler did not bark at this. Probably had his mouth full from the for loop :)

blattms:
I would assume that opm-core is not totally c89 compliant. There seem to be a lot of declaration in the middle of functions.
Mandating a c99 compiler and using the correct switches in cmake might save some work.

bska:
> There seem to be a lot of declaration in the middle of functions.

In a few files, certainly.  Others are strictly C89.

atgeirr:
> The Windows build system used by Resinsight only supports C89

I assume this means Visual Studio? In any case, that is unfortunate because it means we may not want to depend on C99 too much. But one can always make .cpp files of them and use the C++ compiler, I think that may be an ok workaround. We did this a while back, and had to insert some explicit pointer casts to make the C code valid C++, but it was not too much work.

alfbr:
I don't think moving to c99 is a problem. Let med check on the resinsight side.

joakim-hove:
> I assume this means Visual Studio? 

Yes - however the world is progressing and Visual Studio 2015 has C99 support; although it seems the jury is still out regarding the extent of the support.

> In any case, that is unfortunate because it means we may not want to depend on C99 too much.

Now don't get me wrong - I'm all for portability - but do we have immediate plans of supporting OPM on Windows?

>But one can always make .cpp files of them and use the C++ compiler, I think that may be an ok workaround. 

Yes - that is certainly an idea.

dr-robertk:
I will merge this if nobody objects, because it's really annoying to have a master branch that does not compile.

opm/core/wells/wells.c
bska: > I was kind of surprised that the compiler did not bark at this

Me too, although GNU C89 allows that as an extension.  Unless you specify `-ansi` (or `-std=...`) you get GNU C89 in GCC prior to version 5.0.  The default C dialect changed to GNU C11 in that major release.

I personally guard against that surprising behaviour by **always** using
```
-std=... -pedantic
```
when I compile C.

atgeirr:
I don't object, I'll even press the button :-)

PR #870, opened 2015-09-03
Limit the time step growth for all time-stepping algorithm

totto82:
The time step restriction is moved to AdaptiveTimeStepping_impl.hpp to
make it apply to all time-stepping algorithms.


dr-robertk:
Looks good. I'll merge.

PR #872, opened 2015-09-07
RegionMapping<>: Support arbitrary active regions

bska:
This change-set addresses the issue that was brought up in PR #858.  We introduce support for arbitrary region lookup provided the region ID supports `std::hash<>` and simplify the interface by replacing the custom `CellRange` class with a `boost::iterator_range<>`.

Method `cells()` returns an empty range if the region ID corresponds to an inactive region.

Note: The changes to `EquilibrationHelpers.hpp` and `initStateEquil_impl.hpp` are, strictly speaking, ancillary, but the issues were highlighted by this work and the changes included for convenience.

Note: This supersedes and closes #864.

### DIFF
--- a/opm/core/simulator/EquilibrationHelpers.hpp
+++ b/opm/core/simulator/EquilibrationHelpers.hpp
@@ -887,9 +887,9 @@ namespace Opm
         }
 
         /// Return true if capillary pressure function is constant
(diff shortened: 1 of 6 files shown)


atgeirr:
This has been tested and verified to work, merging.

PR #873, opened 2015-09-07
[bugfix] Account for shut completions when checking well in parallel

blattms:
Previously well with just some shut completions erroneously triggered an exception in parallel runs. This is fixed with this commit.

Due to the logic shut completions will always be marked as existing on a process. (Initially all  completions are marked as found. For each open completion we check whether the Cartesian index belongs to the local grid. If that is not the case we mark it as not found). Therefore we now check whether the found number of completions is either the number of shut completions or the number of all completions. In the former case the well is not stored on this process, and in the latter case it is. In other cases we throw an exception.

### DIFF
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -136,7 +136,9 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
 
         {   // COMPDAT handling
             CompletionSetConstPtr completionSet = well->getCompletions(timeStep);
+            // shut completions and open ones stored in this process will have 1 others 0.
             std::vector<std::size_t> completion_on_proc(completionSet->size(), 1);
+            std::size_t shut_completions_number = 0;
             for (size_t c=0; c<completionSet->size(); c++) {
                 CompletionConstPtr completion = completionSet->get(c);
                 if (completion->getState() == WellCompletion::OPEN) {
@@ -195,6 +197,7 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                         wellperf_data[well_index].push_back(pd);
                     }
                 } else {
+                    ++shut_completions_number;
                     if (completion->getState() != WellCompletion::SHUT) {
                         OPM_THROW(std::runtime_error, "Completion state: " << WellCompletion::StateEnum2String( completion->getState() ) << " not handled");
                     }
@@ -205,7 +208,7 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                 // Set wells that are on other processor to SHUT.
                 std::size_t sum_completions_on_proc = std::accumulate(completion_on_proc.begin(),
                                                                       completion_on_proc.end(),0);
-                if ( sum_completions_on_proc == 0 )
+                if ( sum_completions_on_proc == shut_completions_number )
                 {
                     // Mark well as not existent on this process
                     wells_on_proc[wellIter-wells.begin()] = 0;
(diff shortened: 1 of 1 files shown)


opm/core/wells/WellsManager_impl.hpp
```diff
@@ -214,7 +217,10 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                 // Check that the complete well is on this process
                 if( sum_completions_on_proc < completionSet->size() )
                 {
-                    OPM_THROW(std::runtime_error, "Wells must be completely on processor!");
+                    OPM_THROW(std::runtime_error, "Each well must be completely stored 
+                              <<"on processor! Not the case for "<< well->name()<<": "
```
atgeirr: This looks weird, I cannot understand how it can compile? To me it looks like the starting " is ended by the one after the << token. Also things are a bit mashed together making it harder to read.

atgeirr:
Apart from the OPM_THROW issue noted above this is good. Will merge after revision.

blattms:
You were absolutely right. It did not compile. I had fixed this
locally but never committed it.

Added it to this PR. I can smash the two commits into one if you
prefer.

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -214,7 +217,10 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                 // Check that the complete well is on this process
                 if( sum_completions_on_proc < completionSet->size() )
                 {
-                    OPM_THROW(std::runtime_error, "Wells must be completely on processor!");
+                    std::size_t missing = completionSet->size()-shut_completions_number-sum_completions_on_proc;
```
atgeirr: I now noticed another little thing: is it correct that while we test if `sum_completions_on_proc < completionSet()->size()`, we let `missing` be equal to their difference minus `shut_completions_number`. Should not these be the same (so that missing would be 0 if the test succeeded).

blattms: As I am the kind of mathematician that has troubles counting to three there migh be issue. Here is my reasoning:

If ```sum_completions_on_proc < completionSet->size()``` then either it is exactly the number of shut completions or bigger that that. The former can cannot occur as we test for it directly before this statement and continue.
Therefore we should have found ```shut_completions_number+sum_completions_on_proc``` completions (shut ones are always found). In this case missing should be correct.

BTW I agree the code is not very clear. Maybe rephasing is along
```c++
if ( sum_completions_on_proc == shut_completions_number )
{
    // Mark well as not existent on this process
    wells_on_proc[wellIter-wells.begin()] = 0;
    continue;
}
else
{
    // Check that the complete well is on this process
    if( sum_completions_on_proc < completionSet->size() )
    {
         std::size_t missing = completionSet->size()-shut_completions_number-sum_completions_on_proc;
         OPM_THROW(std::runtime_error, "Each well must be completely stored "
                                 <<"on one process! Not the case for "<< well->name()<<": "
                                 <<missing<<" completions missing.");
    }
}```
together with some comments would help?
Hopefully we will never run into this if branch...

atgeirr: I agree that rewriting it as you suggest is a little clearer. However, I still thing there is a bug. If a well contains 10 completions, three of which are shut, then completionSet->size() is 10. If then sum_completions_on_proc is 9, missing will become 10 - 9 - 3 = -2 instead of 1 as it should. Therefore I think the shut_completions_number should not be in the computation of `missing` (it is already included in sum_completions_on_proc).

(Also a nitpick: please use spaces around operators such as <<, and after `if`.)

blattms: Finally got it. You are right.

blattms: An convenient amount of spaces should now be included.

atgeirr:
Thanks, no need to smash, I'll merge.

PR #875, opened 2015-09-08
Change default values

totto82:
- use time stepping algorithm pid instead of pid + iter
Adjusting the time-steps on the number of linear iterations does
currently not give any improvents on the time-stepping.
- Change the pid tolerance. The time-stepper will take longer time-steps
and thus reduce the simulation time significantly. The Norne and the SPE
results does not degrade
- Less aggressive reduction of time-steps after convergence problems


totto82:
For norne this gives 20% reduction in run time compared to master.

alfbr:
Have you compared it to Kai's parameter file (i.e., should I rerun at my end with these settings)?

totto82:
These values gives slight improvement in simulation time over Kai's parameters. It uses more time-steps but reduces the number of convergence issues.

alfbr:
Thanks, great if this is merged today. Anybody up for the task?

PR #876, opened 2015-09-08
Allow a few well completions in the overlap.

blattms:
If on one process a well completion is next to border then
it might also be stored in the neighbor process. Still not
all the completions of the well are known to the neighbor.
This breaks the previous assumption that for each well all
completions must belong to the partition of the process.

Therefore with this commit we allow wells that only have a
part of their completions assigned to the partition of the process.
This wells are deactivated under the assumption that they must
exist completely on another process due to the partitioning.

### DIFF
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -152,10 +152,6 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                     if (cgit == cartesian_to_compressed.end()) {
                         if ( is_parallel_run_ )
                         {
-                            // \todo remove hack for parallel runs! HACK!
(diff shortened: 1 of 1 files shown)


atgeirr:
To see if I got this right: if a completion in located in the overlap set of a partition (but not the owned set) it would still be counted as a "completion_on_proc", triggering an exception (before this PR)?

Would it be better, when checking completion location, to ask the grid somehow if the completion is in the owned or overlap sets? Then the current check should still hold.

blattms:
@atgeirr Exactly. This occured e.g. for Norne and well E-1H. Looking at the disjoint partitioning of cells (without overlap) this all well completions are only on one process but some of them are directly at the border. Therefore with overlap these are seen by another process, but that one is missing the rest.

We could perform this check by passing more information to the WellManager. But this would need to precomputed to make it efficient. Currently we actually assure that our assumption holds with PR OPM/dune-cornerpoint#167

PR #879, opened 2015-09-14
Remove arguments from BlackoilPropertiesFromDeck constructors.

atgeirr:
Two arguments (cell centroids and dimension) are unused, and have been removed.

That makes all constructors (and the init() methods) regular functions (not templates), so they have been moved into the .cpp file and the _impl.hpp has been removed. There are followups in opm-autodiff and opm-polymer.


bska:
Since you're in this area doing cleanup anyway, can I persuade you to make the code block that initialises the `materialLawManager` with something that's more targeted and doesn't introduce a new variable in the block scope?  I.e., replace the block
```C++
std::vector<int> compressedToCartesianIdx(number_of_cells);
for (int cellIdx = 0; cellIdx < number_of_cells; ++cellIdx) {
    if (global_cell) {
        compressedToCartesianIdx[cellIdx] = global_cell[cellIdx];
    }
    else {
        compressedToCartesianIdx[cellIdx] = cellIdx;
    }
}
auto materialLawManager = std::make_shared<MaterialLawManager>();
materialLawManager->initFromDeck(deck, eclState, compressedToCartesianIdx);
```
in the constructors for `BlackoilPropertiesFromDeck` with something along the lines of
```C++
auto materialLawManager = std::make_shared<MaterialLawManager>();
{
    std::vector<int> gc;
    if (global_cell != nullptr) {
        gc.assign(global_cell, global_cell + number_of_cells);
    }
    else {
        gc.reserve(number_of_cells);
        for (int c = 0; c < number_of_cells; ++c) {
            gc.push_back(c);
        }
    }

    materialLawManager->initFromDeck(deck, eclState, gc);
}
```

atgeirr:
> Since you're in this area doing cleanup anyway

I'll make changes along the lines you suggest, and some more. Hold off merging.

atgeirr:
I have now made the change you requested (using a free function in an anonymous namespace). I have also made a change to decouple compilation of saturation function implementations a bit more from the rest, but no longer including the EclMaterialLawManager in header files (only in implementation files). This may cause a minor improvement in compilation times, but is also intended to make my task easier as I work on removing compile warnings in that opm-material subsystem.

bska:
> I have now made the change you requested (using a free function in an anonymous namespace). I have also made a change to decouple compilation of saturation function implementations a bit more from the rest

Much appreciated.  I'll give it a test drive and merge unless anything turns up.

bska:
> I'll give it a test drive and merge unless anything turns up.

No issues in simple testing.  I'll merge this and the companion PRs.

PR #881, opened 2015-09-15
Use cell thickness from EclipseGrid in the well transsmissibility

totto82:
The `dz` calculated in `WellDetails::getCubeDim()` is not correct in cases where the face centroid of the horizontal faces is located above or below the face centroid or the vertical faces. The cell thickness in EclipseGrid, calculated using the Z-coordinates, is therefore used instead.

Tested on SPE9 with corner-point geometry. 

Questions to the reviewers.

 1. Should the call to `WellDetails::getCubeDim()` be moved outside `createWellsFromSpecs` and instead the cell dimensions passed directly.

 2. Do we need to calculate `dx` and `dy` from the z-coordinates as well?

### DIFF
--- a/opm/core/wells/WellsManager.hpp
+++ b/opm/core/wells/WellsManager.hpp
@@ -164,6 +164,7 @@ namespace Opm
                                    const int* cart_dims,
                                    FC begin_face_centroids, 
                                    int dimensions,
+                                   std::vector<double>& dz,
                                    std::vector<std::string>& well_names,
                                    std::vector<WellData>& well_data,
                                    std::map<std::string, int> & well_names_to_index,
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -107,6 +107,7 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                                         const int* cart_dims,
                                         FC begin_face_centroids,
                                         int dimensions,
+                                        std::vector<double>& dz,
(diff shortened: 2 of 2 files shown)


joakim-hove:
I am (as usual ...) not really into this - but to me the `WellDetails::getCubeDim()` and `EclipseGrid::getCellThickness()` seems to be very different beasts; the former seems to calculate the dimensions of an *embedding cube* - whereas the latter calculates an average thickness. The might both be right in their respective context, but they are quite different and to me it seems a bit suspicious to just change one with the other?

bska:
>  `WellDetails::getCubeDim()` seems to calculate the dimensions of an embedding cube

Yes, at least mostly.  The function computes the geometric extents (i.e., (`dx`, `dy`, `dz`)) of a cube defined by the locations of the centroids of the cell's connections.

atgeirr:
This looks ok, it is a nice touch that the feature is optional (in the sense that one can pass an empty dz and have the old behaviour). I'll merge.

I wonder what we would observe on a more complicated cornerpoint grid though. On the Norne case we have given well connection factors in the deck, so this code path is not used there.

bska:
> This looks ok,

Only partly.  The block below will fail if `global_cell` is `nullptr`.
```C++
std::vector<double> dz(number_of_cells);
for (int cell = 0; cell < number_of_cells; ++cell) {
    dz[cell] = eclGrid->getCellThicknes(global_cell[cell]);
}
```

atgeirr:
> Only partly. The block below will fail if global_cell is nullptr.

I'll look at this, and try not to repeat myself all over the place...

PR #882, opened 2015-09-15
WellState: move mapping from well name to well index to here.

dr-robertk:
This PR moves the wellname -> wellIndex mapping from WellStateFullyImplicitBlackoil to the base class. 
This is needed for the ParallelDebugOutput where we only have WellSate objects and need this mapping (which is already available in the Blackoil stuff).

### DIFF
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -63,6 +63,7 @@ namespace Opm
 
                     // setup wellname -> well index mapping
                     {
+                        assert( wells->name[ w ] );
                         std::string name( wells->name[ w ] );
                         assert( name.size() > 0 );
                         mapentry_t& wellMapEntry = wellMap_[name];
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -23,6 +23,8 @@
 #include <opm/core/wells.h>
 #include <opm/core/well_controls.h>
 #include <vector>
+#include <array>
+#include <map>
 #include <cassert>
 #include <cstddef>
 
@@ -32,6 +34,10 @@ namespace Opm
     /// The state of a set of wells.
     class WellState
     {
+    protected:
+        typedef std::array< int, 3 >  mapentry_t;
+        typedef std::map< std::string, mapentry_t > WellMapType;
+
     public:
         /// Allocate and initialize if wells is non-null.
         /// Also tries to give useful initial values to the bhp() and
@@ -41,6 +47,9 @@ namespace Opm
         template <class State>
         void init(const Wells* wells, const State& state)
         {
+            // clear old name mapping
+            wellMap_.clear();
+
             if (wells) {
                 const int nw = wells->number_of_wells;
                 const int np = wells->number_of_phases;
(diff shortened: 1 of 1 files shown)


opm/core/simulator/WellState.hpp
```diff
@@ -51,6 +60,19 @@ namespace Opm
                 for (int w = 0; w < nw; ++w) {
                     assert((wells->type[w] == INJECTOR) || (wells->type[w] == PRODUCER));
                     const WellControls* ctrl = wells->ctrls[w];
+
+                    // setup wellname -> well index mapping
+                    {
+                        std::string name( wells->name[ w ] );
+                        assert( name.size() > 0 );
```
bska: That assert is not really valid.  It is possible, though unlikely, that `wells->name[w]` is `nullptr`.

dr-robertk: So it's correct then. In case the name empty we have a problem and let the user/developer know by throwing an assertion. If the name is empty the mapping would not work.

bska: > So it's correct then.

No it's not because you (or the original author) are checking the wrong condition.  If it's an error to call this function with a null name, then the check should be
```C++
assert (wells->name[w] != nullptr)
```
and callers of `init()` had better verify that the condition is satisfied.

You can't use `assert()` to check user input--it is only to check internal program logic and therefore present only if built without `NDEBUG`.  Besides, `->name[w]` may be `nullptr` as a result of an earlier allocation failure (see function [`add_well()`](https://github.com/OPM/opm-core/blob/master/opm/core/wells/wells.c#L405)), which is admittedly rather unlikely.

> If the name is empty the mapping would not work.

If the name is empty you can't ask for that particular well by name.  Other wells may have non-null names and you can ask for their well IDs by name.

opm/core/simulator/WellState.hpp
```diff
@@ -23,6 +23,8 @@
 #include <opm/core/wells.h>
 #include <opm/core/well_controls.h>
 #include <vector>
+#include <array>
+#include <map>
```
atgeirr: Must also include <string>.

opm/core/simulator/WellState.hpp
```diff
@@ -32,6 +34,10 @@ namespace Opm
     /// The state of a set of wells.
     class WellState
     {
+    protected:
```
atgeirr: Please place protected items after the public section.

atgeirr:
When the two minor issues indicated have been fixed this is ready to be merged in my opinion (but must of course wait until the opm-autodiff PR also is ready).

opm/core/simulator/WellState.hpp
dr-robertk: I has to be before wellMap(), otherwise the header does not compile. So there will be a protected section in the public section. I don't know how this should be resolved.

atgeirr: I did not notice the wellMap() method, that of course requires a public typedef. If that method was protected (so there is no instance of the type in the public interface) the type could have been protected, too. The implementation of a method, even inline, is always allowed to use types (also data members and methods) declared further down. The following class compiles:
```
class Foo
{
public:
    int bar() { Grouch x = 8; return x; }
private:
    typedef int Grouch;
};
```

atgeirr:
I cannot see that the missing `<string>` include has been fixed?

dr-robertk:
Fixed.

atgeirr:
Thanks, merging!

PR #883, opened 2015-09-15
Account for possible null global_cell

atgeirr:
The `compressedToCartesian()` function has been moved to a separate header and used to avoid error in the case a `nullptr` is passed for `global_cell` in `WellsManager`.

I expect that this very minor utility can be useful a few other places as well.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -415,6 +415,7 @@ list (APPEND PUBLIC_HEADER_FILES
 	opm/core/transport/reorder/tarjan.h
 	opm/core/utility/Average.hpp
 	opm/core/utility/CompressedPropertyAccess.hpp
+	opm/core/utility/compressedToCartesian.hpp
 	opm/core/utility/DataMap.hpp
 	opm/core/utility/ErrorMacros.hpp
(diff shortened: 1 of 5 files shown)


bska:
> The `compressedToCartesian()` function has been moved to a separate header

Thanks.  I don't immediately see the need for an inline function here.  Would it not be equally useful if implemented in a `.cpp` file?

atgeirr:
> Thanks. I don't immediately see the need for an inline function here

Just the shortest route. I'll move it to a .cpp file.

bska:
>> Thanks. I don't immediately see the need for an inline function here
>
> Just the shortest route. I'll move it to a `.cpp` file.

Much appreciated.  Then you can move the `<numeric>` header too as that's a detail of the implementation.

atgeirr:
Done. For a minor utility it sure got a lot of attention and care :-)

bska:
> For a minor utility it sure got a lot of attention and care :-)

Indeed.  We still need a little more TLC, though, because the header still includes `<numeric>`.

bska:
> Remove unnecessary include.

PR #886, opened 2015-09-21
Changed Table code to use TableContainer

joakim-hove:
Followup to: https://github.com/OPM/opm-parser/pull/585

Comments are welcome - but this should *not* be merged before the opm-parser PR is merged.

With this PR the `DeckWithLiveOil` test in `test_equil.cpp` fails. To me it seems that the test should have failed all along, but that out of bounds access to the previous `std::vector<RsvdTable>` implementation failed silently?

### DIFF
--- a/opm/core/props/pvt/PvtDead.cpp
+++ b/opm/core/props/pvt/PvtDead.cpp
@@ -34,7 +34,7 @@ namespace Opm
     // Member functions
     //-------------------------------------------------------------------------
     /// Constructor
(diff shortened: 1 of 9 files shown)


atgeirr:
The code looks ok to me. Have anyone investigated the test_equil.cpp failure?

joakim-hove:
> Have anyone investigated the test_equil.cpp failure?

Not as I am aware of.

joakim-hove:
> >Have anyone investigated the test_equil.cpp failure?

> Not as I am aware of.

That was a misunderstanding on my part - fixed now.

PR #887, opened 2015-09-21
Support for equil-initialisation for datum depth outside oil zone.

osae:
As needed for model 2 ...

### DIFF
--- a/opm/core/simulator/initStateEquil_impl.hpp
+++ b/opm/core/simulator/initStateEquil_impl.hpp
@@ -337,7 +337,7 @@ namespace Opm
                   const Region&               reg   ,
                   const std::array<double,2>& span  ,
                   const double                grav  ,
-                  const double                po_woc,
+                  double&                     po_woc,
                   const CellRange&            cells ,
                   std::vector<double>&        press )
             {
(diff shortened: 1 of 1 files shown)


atgeirr:
I have read this very superficially, and it looks ok so far. However I have not done a deep review. I assume this has been tested to give correct results on both Norne and the new model?

bska:
> and it looks ok so far

Possibly.  As far as I can tell this implements something almost but not quite unlike the process stipulated by ECL's `EQUIL` keyword.

osae:
Testing of EQUIL functionality ...

- Norne:  Same results as before.  Compared to eclipse you will find significant deviations in the usual 15 cells due to disagreement regarding cell centre depth versus contacts.  This issue can be completely eliminated by replacing use of cell-centroid by cell-centre-depth (zcoorn average) a few places in the code.  Question whether we should go for this change or simply ignore the difference; cells centred close to the contact will be 50% off anyway ...
- Model 2:  Reasonable results.  More difficult to match than Norne that uses SWATINIT.  Some problems regarding geometry.  At least five cells where flow and eclipse disagree on active/inactive status.  Also a lot of cells where the current MINPV treatment leads to disagreements regarding cell depths.

osae:
Initialisation - some further testing for model 2 ...

Reasonable initial saturations for model 2 can be obtained as follows:

- Ignore two cells that are active in flow but inactive in eclipse: [74,29,28] and [84,39,8]
- Ignore two cells that are inactive in flow but active in eclipse: [104,90,21] and [56,178,28]
- Suspend the corner-point adjustments for "cell below" in MinpvProcessor.hpp (line 112-119).
- Increase resolution of the hydrostatic pressures versus depth calculations in the equilibration algorithm in opm.  This compensates for a less sophisticated algorithm occasionally leading to much larger vertical spans then in eclipse. (E.g. a small goc or a large woc used to signal no gas/water phase is included in the span.)  Current choice in opm is 100 points, identical to the eclipse default. By instead using 2000 points, we see a maximum deviation of ca 0.005 for the initial saturations.

atgeirr:
Thanks for the results. Since this update does not adversely affect our ability to run Norne, I think we should merge this, unless you have more you want to add here?

> This issue can be completely eliminated by replacing use of cell-centroid by cell-centre-depth (zcoorn average) a few places in the code. Question whether we should go for this change or simply ignore the difference

We should make this available by default, I believe. That should be done as a separate PR.

osae:
Nothing more to add as this point.  
Will come back with a separate PR for the centroid vs centre-depth issue.

PR #888, opened 2015-09-22
WARN instead of THROW if completion is not found in the grid

totto82:
Completions that are not found in the grid are ignored. 

This is a necessary fix to make Model 2 run.

### DIFF
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -159,8 +159,9 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                         }
                         else
                         {
-                            OPM_THROW(std::runtime_error, "Cell with i,j,k indices " << i << ' ' << j << ' '
-                                      << k << " not found in grid (well = " << well->name() << ')');
+                            OPM_MESSAGE("****Warning: Cell with i,j,k indices " << i << ' ' << j << ' '
+                                      << k << " not found in grid. The completion will be igored (well = "
+                                      << well->name() << ')');
                         }
                     }
                     else


joakim-hove:
Please implement a error mode in the ParseMode: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Parser/ParseMode.hpp to handle this.

totto82:
@joakim-hove Do you suggest to add ParseMode to the COMPDAT handling in opm-parser? In other words make sure that completions that are not in the grid never turns up in the simulator. 
What about MINPV? Does the parser know about cells removed using MINPV?

joakim-hove:
> Do you suggest to add ParseMode to the COMPDAT handling in opm-parser? 

My main suggestion was that instead of just globally changing this from an Exception to a Warning we should create a error mode variable (like this this: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/Parser/ParseMode.hpp#L121) for this, so that the behavior can be controlled on a per case basis.

Adding the `ParseMode::GRID_COMPLETION_NOT_IN_GRID` is of course the simple part - the more interesting question is how to get the `ParseMode` instance available at the call site; I see  two possible solutions:

1. Just pass the `ParseMode`instance all the way down from `flow::main()`
2. Let the parser handle this - that will probably create create problems with `MINPV`.

As part of the current code-reorganisation process we should move the `ParseMode` class to `opm-common`, rename into something like `OpmError` and possibly make it a singleton?!

alfbr:
Very interesting suggestion. That would certainly open up for a much more powerful parsemode implementation, while keeping the parser mean and lean. I hope more developers  share opinions here to help us get a good basis for a decision.

totto82:
I agree that something like OpmError would be a powerful tool. For instance in situations when users combine exotic features like the solvent option with the segmented wells that is not tested. OpmError could work as a traffic light. First stop you, but then if you persist, give you a warning that from here on you are at your own risk. 

For me this sounds like a large job that needs to be discussed as a whole. Could this be merged in the meantime? It is needed in order to continue testing Model 2.

joakim-hove:
> For me this sounds like a large job

With the merging of: https://github.com/OPM/opm-parser/pull/589 you can get a `ParseMode`instance from `EclipseState::getParseMode()` - so now this is quite trivial:

1. Add the error mode, i.e. invent a suitable name, to the `ParseMode` implementation.
2. Handle the error with `ParseMode::handleError()

> that needs to be discussed as a whole

Well as I see it this has been discussed extensively - and the `ParseMode` implementation is (start at least) of a conclusion on that discussion.

> For me this sounds like a large job

Implementing the `ParseMode` as singleton, integrating it with `OPM_ERROR` and so on is of course a more involved task which should be preempted by a more extensive discussion.

bska:
@totto82
> [Error handling] needs to be discussed as a whole

@joakim-hove 
> Well as I see it this has been discussed extensively - and the `ParseMode` implementation is (start at least) of a conclusion on that discussion.

Agreed.  The `ParseMode`, possibly renamed, is good candidate for a centralised and unified error handling facility.

atgeirr:
While we discuss changes to error handling, I'll just merge this PR since it gives the behaviour expected by engineers: not stopping due to a well completion in a non-active cell (we assumed before that we should stop on this).

joakim-hove:
> I'll just merge this PR since

As I have already said we have discussed error handling repeatedly; the whole `ParseMode` machinery has been an attempt to address the topic. I have received limited feedback on that implementation - but my general impression has been that people have been quite happy with it. 

Vague statements like:

> While we discuss changes to error handling

is just procrastination - we have a system for error handling. It can certainly be improved, but when you as a maintainer choose to bypass it completely instead of mandating it's use we will never get beyond the "*We need to discuss error handling*" phase. For the current issue the changes required would be less than 30 minutes:

1. Add a named error, like e.g. "WELL_COMPLETED_OUTIDE_GRID" to the `ParseMode` in opm-parser.
2. In opm-core - handle the error with `ParseMode::handleError()`

bska:
> when [we] choose to bypass [the error handling mechanism] completely instead of mandating its use we will never get beyond the *"We need to discuss error handling"* phase

I tend to agree.  We should not allow code that implements ad-hoc error handling when better alternatives exist--especially  when they're as easily accessible as the recipe outlined in @joakim-hove's [comment](https://github.com/OPM/opm-core/pull/888#issuecomment-143686969)

atgeirr:
I disagree that we have agreed on using ParseMode to handle issues like the current one. We actually need to discuss this, it seems to be getting more urgent that we do so. I'll try to make this happen.

PR #889, opened 2015-09-22
Use correct units for the accumulated output

totto82:
Avoid division on time in the total output of rates.


joakim-hove:
This seems obviously right - merging.

bska:
> This seems obviously right - merging.

Yeah, apart from the confusing name `targetRate` which implies that the unit should be per time.  It would be good if we could rename that to `volumeUnit` or some such.

PR #890, opened 2015-09-23
Prevent WellsManager from subscripting empty container.

blattms:
While hopefully not a bug it raises an exception with gcc's
libc debugging mode. Therefore we resort to using C++11's
std::vector::data instead.

The exception was raised when running SPE9 in parallel.


bska:
Thanks!  I'll merge into master.

PR #892, opened 2015-09-30
PinchProcessor

qilicun:
This PR aims to provide a rough  PinchProcessor for PINCH and MINPV keywords. Build non-neighbor connections for the cells which's pore volume less than MINPV or the thickness less than the value of item 2 in PINCH keyword. This processor just support TOPBOT in item 4 and TOP in item 5, ALL in item 5 is also implemented but not verified, since ECLIPSE output for this part is a little strange.

The general ideas is:
-  Mark all the Minpved cells, including small thickness cells.
-  Indentify the Pinchous column, because NNC needs top and bottom cell of the column are active. So the pinchouts column can be lots situation, the kernel part of this processor is to take care of the following cases:
    - both sides are active --- best situation,
    - one side are active. Look up or down this side from the pillar if active cell can be found, rebuild this column, if not drop this one.
    - both side are inactive. Same stratedy as above.

Norne is tested with this processor. The output is too long I think it's better to paste it in the comments.
For most of the NNC in NORNE due to MINPV, OPM create very close results to ECLIPSE, but for some cells, the transmissbiliy is far from ECLIPSE, needs to find out what happens for these cells.
So I think this processor is not ready to merge, please fell free to make comments to help me to improve.


qilicun:
[nnc.txt](https://github.com/OPM/opm-core/files/4926/nnc.txt)

qilicun:
pinch

atgeirr:
Thanks for the contribution! Just a few initial comments:
 - You must add the new file to CMakeLists_files.cmake.
 - Is this supposed to be used from opm-autodiff? Asking since there are no such changes in this PR.
 - Would it be possible to add a unit test? Perhaps with just a few cells, verifying that the expected nncs get created?
 - I am not sure what the posted data are: are they transmissibilities at NNCs? One-sided transmissibilities? Also: I assume these data look better than without this PR?

qilicun:
@atgeirr 

>You must add the new file to CMakeLists_files.cmake.

Sorry, I just forgot, will do it later.
>Is this supposed to be used from opm-autodiff? Asking since there are no such changes in this PR.

Yes.

>Would it be possible to add a unit test? Perhaps with just a few cells, verifying that the expected nncs get created?

Unit test is possible, but I can't find a place in opm-core to add it. probably in opm-autodiff?

>I am not sure what the posted data are: are they transmissibilities at NNCs? One-sided transmissibilities? 

The posted data here is face transmibisslities, not one-sided. 

>Also: I assume these data look better than without this PR?

Yes, some of the data looks not good, but that's why I need the suggestion. Since it's really hard to guess what ECLIPSE really does  for this keyword. From the NNCs cell point of view, I think my guess is correct, but the trans is not that correct.

joakim-hove:
> Unit test is possible, but I can't find a place in opm-core to add it. probably in opm-autodiff?

I always bitch about tests - I would be happy to help you adding it. I have not looked at the code - but as long as the implementation is in `ppm-core` I would *really* say that it should be possible to have the test there as well.

qilicun:
@joakim-hove 

>I always bitch about tests - I would be happy to help you adding it.

That's right, tests is very important.

> I would really say that it should be possible to have the test there as well.

Sure, I will add some tests after national holiday!

atgeirr:
> Sure, I will add some tests after national holiday!

I'll come back to this next week, then. I think we should try to get this in before the upcoming release.

bska:
> I think we should try to get this in before the upcoming release.

No disagreement from me.  From a cursory reading I do spot one potential performance issue, though.  The function [`getActiveCellIdx_()`](https://github.com/qilicun/opm-core/blob/pinchprocessor/opm/core/grid/PinchProcessor.hpp#L277-L291) is called in a number of contexts, sometimes within loops.  As the implementation of that function is linear in the number of active cells, that could become a bottleneck on larger models.  As an immediate remedy we could cache the result as a constant-time lookup table.

That said, I'm not entirely sure that the function is needed at all.  This implementation appears to do everything in terms of global cell indices because that's what class [`Opm::NNC`](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/NNC.hpp) requires.  It might be possible, though probably quite a bit of work, to switch to doing the processing in terms of active cells and convert to global indices only when forming the result--i.e., when calling `NNC::addNNC()`.  I have not tried to do that work.

qilicun:
>to switch to doing the processing in terms of active cells and convert to global indices only when forming the result--i.e., when calling NNC::addNNC()

Good point, will modify later.

qilicun:
I have already made a test case for this processor and the result looks very good, after https://github.com/OPM/opm-core/issues/898 solved, the test will be updated.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef OPM_PINCHPROCESSOR_HEADER_INCLUDED
+#define OPM_PINCHPROCESSOR_HEADER_INCLUDED
+
+#include <opm/core/utility/ErrorMacros.hpp>
```
totto82: Needs to be changed to <opm/common/ErrorMacros.hpp>

totto82:
@qilicun I will try to implement this into the simulator to test it on norne.

qilicun:
@totto82  That could be great! If you find any interesting, please let me know.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+#include <array>
+#include <iostream>
+#include <algorithm>
+#include <unordered_map>
+#include <limits>
+
+namespace Opm
+{
+
+    using namespace Opm::UgGridHelpers;
```
atgeirr: This should not be done at namespace scope in a header. Put it inside the functions that need it instead.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        /// \param[in]    transMode    item 4 in PINCH keyword
+        /// \param[in]    multzMode    item 5 in PINCH keyword
+        PinchProcessor(const double minpvValue,
+                       const double thickness,
+                       const PinchMode::ModeEnum transMode,
+                       const PinchMode::ModeEnum multzMode);
+        /// Generate NNCs for cells which pv is less than MINPV.
+        /// \param[in]    Grid    cpgrid or unstructured grid
+        /// \param[in]    htrans  half cell transmissibility
+        /// \param[in]    multz   Z+ transmissibility multiplier
```
atgeirr: What are the expected sizes of htrans, multz, pv and dz? In particular: the doc should state if is it all logical cartesian cells, or active cells since that is easily confused.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+            }
+        }
+        return hf_ix;
+    }
+
+
+
+    template<class Grid>
+    inline int PinchProcessor<Grid>::getActiveCellIdx_(const Grid& grid,
+                                                       const int cellIdx)
```
atgeirr: This parameter should be called globalCellIdx to more clearly separate it from the active cell index.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+    {
+        const int nc = numCells(grid);
+        const int* global_cell = globalCell(grid);
+        int idx = -1;
+        for (int i = 0; i < nc; ++i) {
```
atgeirr: This linear search may give performance problems for large grids, since it is called many times. Either you could use a binary search (std::lower_bound for example), but even better (and more robust in case we drop the convention that globalCell(grid) is an increasing sequence) is to create a map or vector for the lookup.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        auto ijk1 = getCartIndex_(cellIdx1, cartDims(grid));
+        auto ijk2 = getCartIndex_(cellIdx2, cartDims(grid));
+
+        if (commonFace == -1) {
+            OPM_THROW(std::logic_error, "Couldn't find the common face for cell " << cellIdx1<< "("<<ijk1[0]<<","<<ijk1[1]<<","<<ijk1[2]<<")"<< " and " << cellIdx2<<"("<<ijk2[0]<<","<<ijk2[1]<<","<<ijk2[2]<<")");
+        }
+
+        return commonFace;
+    }
+
```
atgeirr: In general, I would appreciate more whitespace between the function implementations for easier readability. From 3-6 blank lines is good.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+            trans[f] = 1. / trans[f];
+        }
+
+        return trans;
+    }
+
+    template<class Grid>
+    inline std::vector<std::vector<int>> PinchProcessor<Grid>::getPinchoutsColumn_(const Grid& grid,
+                                                                           const std::vector<int>& actnum,
+                                                                           std::vector<int>& minpvCells)
```
atgeirr: While I think the code is correct, it is a little surprising that minpvCells is changed by this method. I would therefore prefer the tiny overhead of taking a copy of the vector inside this function, alternatively the arguments could be changed so that the call to getMinpvCells_() is here instead of in transTopBot_().

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+    }
+
+
+    template<class Grid>
+    inline std::vector<int> PinchProcessor<Grid>::getMinpvCells_(const Grid& grid,
+                                                                 const std::vector<int>& actnum,
+                                                                 const std::vector<double>& pv,
+                                                                 const std::vector<double>& dz)
+    {
+        const int nc = numCells(grid);
```
totto82: This code assumes that the minpv cells is in the grid. They are not if they are removed by the MinpvProcessor. I believe the correct thing here would be to run over all logical Cartesian cells. 
for (int Idx = 0; idx < pv.size(); ++ Idx)  {
...
}
Then you don't need the global_cell mapping.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+                        if (actnum[topCell]) {
+                            break;
+                        } else {
+                            auto it = seg.begin();
+                            seg.insert(it, topCell);
+                        }
+                    }
+                    pinFaces.push_back(interface_(grid, topCell, Opm::FaceDir::ZMinus));
+                } else {
+                    pinFaces.push_back(interface_(grid, topCell, seg.front()));
```
totto82: Again. If the MinpvProcessor is ran before the PinchProcessor the seg.front() cell does not exists. I believe using Opm::FaceDir::Zminus instead of seg.front() will work just fine.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+                        botCell = getGlobalIndex_(ijk2[0], ijk2[1], botk, dims);
+                        if (actnum[botCell]) {
+                            break;
+                        } else {
+                            seg.push_back(botCell);
+                        }
+                    }
+                    pinFaces.push_back(interface_(grid, botCell, Opm::FaceDir::ZPlus));
+                } else {
+                    pinFaces.push_back(interface_(grid, seg.back(), botCell));
```
totto82: Same here.

tests/test_pinchprocessor.cpp
```diff
@@ -0,0 +1,112 @@
...
+    PinchProcessor<Grid> pinch(minpv, thickness, transMode, multzMode);
+    std::vector<int> actnum;
+    eclgrid->exportACTNUM(actnum);
+
+    if (actnum.empty() && (nc == eclgrid->getCartesianSize())) {
+        actnum.assign(nc, 1);
+    }
+    std::vector<double> htrans(Opm::UgGridHelpers::numCellFaces(grid));
+    Grid* ug = const_cast<Grid*>(& grid);
+    tpfa_htrans_compute(ug, rock.permeability(), htrans.data());
```
totto82: Note, that there are two tpfa_htrans_compute methods. See GeoProps.hpp. The eclipse standard is to assume local coordinate system for the permeability tensor. In the method you have used here a global coordinate systems is assumed. Both methods should give the same results for this test case, but not for norne.

tests/test_pinchprocessor.cpp
```diff
@@ -0,0 +1,112 @@
...
+    const double thickness = eclgrid->getPinchThresholdThickness();
+    BOOST_CHECK_EQUAL(thickness, 0.001);
+
+    auto transMode = eclgrid->getPinchOption();
+    BOOST_CHECK_EQUAL(transMode, PinchMode::ModeEnum::TOPBOT);
+
+    auto multzMode = eclgrid->getMultzOption();
+    BOOST_CHECK_EQUAL(multzMode, PinchMode::ModeEnum::TOP);
+
+    PinchProcessor<Grid> pinch(minpv, thickness, transMode, multzMode);
```
totto82: I think it would have been beneficial to add MinpvProcessor to the test case in order to test the combination of the two.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        auto minpvSeg = getPinchoutsColumn_(grid, actnum, minpvCells);
+        for (auto& seg : minpvSeg) {
+            std::array<int, 3> ijk1 = getCartIndex_(seg.front(), dims);
+            std::array<int, 3> ijk2 = getCartIndex_(seg.back(), dims);
+            auto tmp = seg;
+            if ((ijk1[2]-1) >= 0 && (ijk2[2]+1) < dims[2]) {
+                int topCell = getGlobalIndex_(ijk1[0], ijk1[1], ijk1[2]-1, dims);
+                int botCell = getGlobalIndex_(ijk2[0], ijk2[1], ijk2[2]+1, dims);
+                if (!actnum[topCell]) {
+                    for (auto topk = ijk1[2]-2; topk > 0; --topk) {
```
atgeirr: I am not sure this loop does what it is intended to:
 - Is the intention to add all inactive cells at the top to `seg`? In that case I think the one with K-coordinate `ijk[2]-1` will be missed (since the loop starts at `ijk1[2]-2`).
 - Should the limit be `topk >= 0`? Otherwise a potential top cell at the very top of the grid could be missed I think.

I would also use 'int' instead of 'auto' here: shorter, and making it clear that it is a signed integer.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        std::vector<int> pinFaces;
+        std::vector<int> pinCells;
+        std::vector<std::vector<int> > newSeg;
+        std::vector<int> minpvCells = getMinpvCells_(grid, actnum, pv, dz);
+        auto minpvSeg = getPinchoutsColumn_(grid, actnum, minpvCells);
+        for (auto& seg : minpvSeg) {
+            std::array<int, 3> ijk1 = getCartIndex_(seg.front(), dims);
+            std::array<int, 3> ijk2 = getCartIndex_(seg.back(), dims);
+            auto tmp = seg;
+            if ((ijk1[2]-1) >= 0 && (ijk2[2]+1) < dims[2]) {
```
atgeirr: A comment about the logic of this if-statement would be good (we are only looking to add nncs, so if the removed segment extends to the top or bottom there is nothing to add).

tests/test_pinchprocessor.cpp
qilicun: That's true, for Norne, I use the other local version in GeoProps.hpp

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+                    }
+                    pinFaces.push_back(interface_(grid, topCell, Opm::FaceDir::ZMinus));
+                } else {
+                    pinFaces.push_back(interface_(grid, topCell, seg.front()));
+                }
+                tmp.insert(tmp.begin(), topCell);
+                newSeg.push_back(tmp);
+                pinCells.push_back(topCell);
+                if (!actnum[botCell]) {
+                    for (auto botk = ijk2[2]+2; botk < dims[2]; ++botk) {
```
atgeirr: I am not sure about this loop's start point either, however the limit `botk < dims[2]` seems correct.

totto82:
I have tested this together with a private branch that incorporates the NNCs from the pinchProcessor in the simulator. The good news it that I can confirm that it reproduce the eclipse NNCs approximately. The  bad news is that the simulator does not convergence with these NNCs. I still don't know if the convergence problems is due to the problematic NNCs or the NNC implementation in the simulator. .

opm/core/grid/PinchProcessor.hpp
qilicun: I think after MinpvProcessor, some cells are moved to its upper or lower cells, and therefor the grid processor will delete that zero volume cells. So run minpv processor before pinch processor will create some troubles..

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+
+    template<class Grid>
+    inline int PinchProcessor<Grid>::interface_(const Grid& grid,
+                                                    const int cellIdx1,
+                                                    const int cellIdx2)
+    {
+        auto cell_faces = cell2Faces(grid);
+        std::vector<int> cellFaces1;
+        std::vector<int> cellFaces2;
+        int commonFace = -1;
```
atgeirr: Move commonFace down to just before line 197 (since that is where it is used).

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        for (auto& f1 : cellFaces1) {
+            for (auto& f2 : cellFaces2) {
+                if (f1 == f2) {
+                    commonFace = f1;
+                    break;
+                }
+            }
+        }
+
+        auto ijk1 = getCartIndex_(cellIdx1, cartDims(grid));
```
atgeirr: Move these inside the if-statement below.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        auto ijk2 = getCartIndex_(cellIdx2, cartDims(grid));
+
+        if (commonFace == -1) {
+            OPM_THROW(std::logic_error, "Couldn't find the common face for cell " << cellIdx1<< "("<<ijk1[0]<<","<<ijk1[1]<<","<<ijk1[2]<<")"<< " and " << cellIdx2<<"("<<ijk2[0]<<","<<ijk2[1]<<","<<ijk2[2]<<")");
```
atgeirr: This line is far too long, and must be broken up.

qilicun:
@totto82 
> I have tested this together with a private branch that incorporates the NNCs from the pinchProcessor in the simulator. 

Thanks!

>I still don't know if the convergence problems is due to the problematic NNCs or the NNC implementation in the simulator. .

Did you test with eclipse NNCs?

opm/core/grid/PinchProcessor.hpp
totto82: I agree. The MinpvProcessor needs to be changed. #112-119 needs to be optional. 
If MinpvMode::ModeEnum::OpmFIL then do #112-119

totto82:
> Did you test with eclipse NNCs? 
In progress....

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        for (auto cellFaceIter1 = cellFacesRange1.begin(); cellFaceIter1 != cellFacesRange1.end(); ++cellFaceIter1) {
+            cellFaces1.push_back(*cellFaceIter1);
+        }
+
+        auto cellFacesRange2 = cell_faces[actCellIdx2];
+        for (auto cellFaceIter2 = cellFacesRange2.begin(); cellFaceIter2 != cellFacesRange2.end(); ++cellFaceIter2) {
+            cellFaces2.push_back(*cellFaceIter2);
+        }
+
+        for (auto& f1 : cellFaces1) {
```
atgeirr: It is unnecessary to build `cellFaces1` and `cellFaces2`, you could write this loop nest just using the ranges, with the same notation (range-for-loop). Also, the type of the loop variables should not be references (you are not changing them), they should be `const auto&` or just `int` (best here I think).

opm/core/grid/PinchProcessor.hpp
qilicun: The only difference between OpmFil  and EclStd is that OpmFil use the original grid volume to compute porve volume, the other parts are just the same.

opm/core/grid/PinchProcessor.hpp
qilicun: That's true, this is a debug remaining issue...will fix it together with others.

opm/core/grid/PinchProcessor.hpp
qilicun: >he arguments could be changed so that the call to getMinpvCells_() is here instead of in transTopBot_().

That's  correct, since the MinpvCells is just the intermediate variable.

opm/core/grid/PinchProcessor.hpp
qilicun: This loop is aimed to find the *active* top and bottom cell in this column, if the original top (bottom)cell is inactive due to actnum, this cell should go into seg until we find the *active* cell, need to go back and check `ijk1[2]-1`

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        std::vector<double> trans(nf, 0);
+        int cellFaceIdx = 0;
+        auto cell2Faces = Opm::UgGridHelpers::cell2Faces(grid);
+        const auto& hfmap = getHfIdxMap_(grid); 
+        const auto& f2c = faceCells(grid);
+        for (int cellIdx = 0; cellIdx < nc; ++cellIdx) {
+            auto cellFacesRange = cell2Faces[cellIdx];
+            for (auto cellFaceIter = cellFacesRange.begin(); cellFaceIter != cellFacesRange.end(); ++cellFaceIter, ++cellFaceIdx) {
+                const int faceIdx = *cellFaceIter;                
+                if (std::find(pinFaces.begin(), pinFaces.end(), faceIdx) == pinFaces.end()) {
```
atgeirr: The std::find is a linear search, so it could be expensive. We may have to change this for acceptable performance on huge grids, but I think it is ok for now. A minor change that would improve performance (although not asymptotic complexity) would be to store the result of the find instead of redoing it below.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+        std::unordered_multimap<int, double> multzmap;
+        if (multzMode_ == PinchMode::ModeEnum::TOP) {
+            for (int i = 0; i < pinFaces.size()/2; ++i) {
+                multzmap.insert(std::make_pair(pinFaces[2*i], multz[getActiveCellIdx_(grid, pinCells[2*i])]));
+                multzmap.insert(std::make_pair(pinFaces[2*i+1],multz[getActiveCellIdx_(grid, pinCells[2*i])]));
+            }
+        } else if (multzMode_ == PinchMode::ModeEnum::ALL) {
+            for (auto& seg : segs) {
+                //find the right face.
+                auto index = std::distance(std::begin(pinCells), std::find(pinCells.begin(), pinCells.end(), seg.front()));
```
atgeirr: The definition of index should be moved below the next for loop since it is not used until the insert calls.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -0,0 +1,497 @@
...
+                    for (auto topk = ijk1[2]-2; topk > 0; --topk) {
+                        topCell = getGlobalIndex_(ijk1[0], ijk1[1], topk, dims);
+                        if (actnum[topCell]) {
+                            break;
+                        } else {
+                            auto it = seg.begin();
+                            seg.insert(it, topCell);
+                        }
+                    }
+                    pinFaces.push_back(interface_(grid, topCell, Opm::FaceDir::ZMinus));
```
atgeirr: It's possible I am confused, but should not the ZMinus and ZPlus be switched?

atgeirr:
This feature should definitely be merged, but rom the comments by @totto82 it looks like it needs a bit more testing and fixing before it is ready, especially to make it work with MINPV, too. I have also made some comments in the diff that require attention.

I general I will say that it looks very good, I appreciate your easy-to-read code both in terms of formatting and its use of C++ features and the standard library. I would in general appreciate a few more comments explaining the intents of code blocks and methods.

totto82:
I have successfully applied NNC generated by the pinchprocessor on norne using FLOW. The results agrees with the results I get if I run OPM with NNC generated using Eclipse.  

I will provide a patch with the necessary changes in opm-autodiff as soon as possible.

opm/core/grid/PinchProcessor.hpp
qilicun: For the topCell, we need the lower face, for botcell upper face is used, if I understood correctly.

atgeirr: If the top cell is the one with smallest K (or z) coordinate, the cells below it are in the increasing z (depth) direction. So the lower face (towards the deeper parts) would be ZPlus I think.

Do you have in your test something that check that this behaves correctly?

qilicun: I didn't test this behaves independently, just check the implementation and comments in opm-parser, I agree some tests need here to verify the code.

opm/core/grid/PinchProcessor.hpp
totto82: For now that's true. But OpmFil should be used in the MinpvProcessor now that the PinchProcessor is available. I have a patch that does the necessary changes in opm-core. I can either make a PR to your branch with the necessary changes, or I can make a separate PR after this is merged that makes it work with MINPV.

qilicun: Both is ok for me, please go ahead at your convenience

qilicun:
@atgeirr 
>So the lower face (towards the deeper parts) would be ZPlus I think.

Just test this part, you're right! The deeper face is ZPlus, other side is ZMinus.

atgeirr:
> Just test this part, you're right! The deeper face is ZPlus, other side is ZMinus

Good to have it tested. It looks to me like you have finished work on this PR, correct? Did you check the logic of the transTopbot_() methods' finding of the bottom and top connections? I no longer see my comments there in the diff so I know you did something, but I could not find what changes you did, which commit would that be in?

qilicun:
>Did you check the logic of the transTopbot_() methods' finding of the bottom and top connections? 

I think the commit is 1c2b04f1501719deeae0d54385c03c6beba752a0, this is fix last week, I think I forgot to make c clear message on this commit.

atgeirr:
Thank you, I'll merge this, then.

PR #896, opened 2015-10-06
Relocated warning headers.

joakim-hove:


bska:
I have performed a simple `make && make test` without ill effects.  I'll merge into master.  Thanks a lot!

PR #897, opened 2015-10-06
Add boolean flag to determine whether crossflow is allowed in a well

totto82:
This PR adds allow_cf to the wells structure that determine whether
crossflow is allowed or not. An extra argument is added to addWell(..)
to specify the allow_cf flag.

Should be considered together with OPM/opm-parser#600

@bska I would appreciate if you could take a look at this. I still feel shaky whenever I touching the wells in opm-core.

### DIFF
--- a/opm/core/wells.h
+++ b/opm/core/wells.h
@@ -106,6 +106,11 @@ struct Wells
      * Internal management structure.
      */
     void *data;
+
+    /**
+     * Array of booleans, if true crossflow is allowed in the well.
+     */
+    bool *allow_cf;
 };
 
 
@@ -185,6 +190,7 @@ create_wells(int nphases, int nwells, int nperf);
  *                       ideally be track ordered.
  * \param[in] WI         Well production index per perforation, or NULL.
  * \param[in] name       Name of new well. NULL if no name.
+ * \param[in] allow_cf   Boolean flag to determine whether crossflow is allowed or not.
  * \param[in,out] W      Existing set of wells to which new well will
  *                       be added.
  *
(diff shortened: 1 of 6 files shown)


opm/core/wells.h
```diff
@@ -106,6 +106,11 @@ struct Wells
      * Internal management structure.
      */
     void *data;
+
+    /**
+     * Array of booleans, if true crossflow is allowed in the well.
+     */
+    bool *allow_cf;
```
bska: Make this `int*` and change the calling interfaces accordingly.  Furthermore, move the declaration ahead of `data`.

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -247,6 +247,12 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                     wd.type = INJECTOR;
                 else
                     wd.type = PRODUCER;
+
+                if (well->getAllowCrossFlow()) {
+                    wd.allowCrossFlow = true;
+                } else {
+                    wd.allowCrossFlow = false;
+                }
```
bska: This is just
```C++
wd.allowCrossFlow = well->getAllowCrossFlow();
```
and should be written as such.

opm/core/wells/wells.c
```diff
@@ -85,6 +85,7 @@ wells_allocate(int nwells, struct Wells *W)
     void *type, *depth_ref, *comp_frac;
     void *well_connpos;
     void *ctrls, *name;
+    bool *allow_cf;
```
bska: Make this `void*`.

atgeirr:
Looks fine, will merge.

atgeirr:
I should of course have realized that a follow-up PR changing the add_well() calls in other modules was missing... Fixing now.

Issue #898, opened 2015-10-08
ACTNUM with cartesian grid 


qilicun:
Hi,

I'm making the tests for pinch processor, I suddenly found that iopm will not creat ACTNUM for cartesian grid, that means MinvpProcessor will not work also. Since this array will be empty, the following error will happen when we specify MINPV keyword together cartesian grid.
```
project/eor/user/liuming/opm/opm-core/opm/core/grid/MinpvProcessor.hpp:96] Wrong size of ACTNUM input, must have one element per logical cartesian cel
```

bska:
We used to treat an empty `ACTNUM` array as "all cells active".  The constructors of Cartesian grids assume that convention.  To be honest, I don't really know what the correct solution is here.  Would you mind testing the change below?

```diff
diff --git a/opm/core/grid/MinpvProcessor.hpp b/opm/core/grid/MinpvProcessor.hpp
index 072a29b..9ba8379 100644
--- a/opm/core/grid/MinpvProcessor.hpp
+++ b/opm/core/grid/MinpvProcessor.hpp
@@ -92,7 +92,7 @@ namespace Opm
         if (pv.size() != log_size) {
             OPM_THROW(std::runtime_error, "Wrong size of PORV input, must have one element per logical cartesian cell.");
         }
-        if (actnum.size() != log_size) {
+        if (!actnum.empty() && (actnum.size() != log_size)) {
             OPM_THROW(std::runtime_error, "Wrong size of ACTNUM input, must have one element per logical cartesian cell.");
         }
 
@@ -101,7 +101,7 @@ namespace Opm
             for (int jj = 0; jj < dims_[1]; ++jj) {
                 for (int ii = 0; ii < dims_[0]; ++ii) {
                     const int c = ii + dims_[0] * (jj + dims_[1] * kk);
-                    if (pv[c] < minpv && actnum[c]) {
+                    if (pv[c] < minpv && (actnum.empty() || actnum[c])) {
                         // Move deeper (higher k) coordinates to lower k coordinates.
                         std::array<double, 8> cz = getCellZcorn(ii, jj, kk, zcorn);
                         for (int count = 0; count < 4; ++count) {
```

qilicun:
@bska 
I think your modification will work, will test later. I'm thinking the same modification needs to be made for PinchProcessor.

>We used to treat an empty ACTNUM array as "all cells active".

[...]

PR #900, opened 2015-10-09
fix actnum with cartesian grid issue.

qilicun:
This fix relates to https://github.com/OPM/opm-core/issues/898


bska:
> This fix relates to #898

Thanks a lot.  I'll merge into master.

PR #902, opened 2015-10-12
Renamed memcmp_double -> opm_memcmp_double

joakim-hove:
The prefix `mem` is supposedly reserved for standard functions.


bska:
> The prefix `mem` is supposedly reserved for standard functions.

Not supposedly.  It **is** reserved for standard functions.

PR #903, opened 2015-10-13
fix build on GCC 4.4

andlaus:
again, the changes required for the actual library are pretty trivial,
but some tests need to be disabled. (do they really need lambdas?)


atgeirr:
I will not merge this. Of the the changes proposed, there are four kinds (one of them not in this particular PR, but I'll count it here):
 - Remove surplus `typename`.
 - Add missing standard includes (for features used in the file).
 - Turn range-for into traditional for loops.
 - Use the preprocessor to disable tests.

I am fine with the first two, will not merge the change of for-loops, and I am somewhat reluctant to accept the last one (you'd need a very good argument...). Feel free to resubmit the first two types of changes and I'll merge them (and possible others).

Why do you require the old gcc btw?

PR #904, opened 2015-10-20
update debian packaging

akva2:


debian/control
```diff
@@ -7,7 +7,8 @@ Build-Depends: build-essential, debhelper (>= 9), libboost-filesystem-dev,
                libdune-common-dev, libdune-istl-dev, cmake, libtinyxml-dev, bc,
                libert.ecl-dev, git, zlib1g-dev, libtool, doxygen, libopm-parser-dev,
                libopm-material-dev, texlive-latex-extra, texlive-latex-recommended,
-               ghostscript, libboost-iostreams-dev
+               ghostscript, libboost-iostreams-dev, libopm-common-dev,
+               libopm-material-dev
```
atgeirr: This should depend on opm-parser should it not? Perhaps I do not understand the purpose of this line...

(Otherwise: same question for other modules.)

akva2: yes it does, and it already did in 2015.04, hence it's already in the 
list (3 lines up).

atgeirr: Sorry, did not notice. Assumed "our" stuff came last for some reason.

PR #905, opened 2015-10-20
Fix timestep bug

atgeirr:
The currentStepNum() method was used (wrongly) instead of reportStepNum() in the RFT-related parts of EclipseWriter and in EclipseWriteRFTHandler.

Also improved formatting in that part of the EclipseWriter (separate commit).

@joakim-hove I have assigned you to review/merge this.

### DIFF
--- a/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
+++ b/opm/core/io/eclipse/EclipseWriteRFTHandler.cpp
@@ -60,7 +60,7 @@ namespace EclipseWriterDetails {
         std::vector<ecl_rft_node_type *> rft_nodes;
         for (std::vector<WellConstPtr>::const_iterator ci = wells.begin(); ci != wells.end(); ++ci) {
             WellConstPtr well = *ci;
-            if ((well->getRFTActive(simulatorTimer.currentStepNum())) || (well->getPLTActive(simulatorTimer.currentStepNum()))) {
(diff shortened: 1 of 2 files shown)


joakim-hove:
Well - I agree with this PR; however as I see it the underlying implementation of `SimtTimer` interface means that nothing is solved:

https://github.com/OPM/opm-core/blob/fbd6fce24f77dc7b999128409f51f6bda27f1926/opm/core/simulator/SimulatorTimerInterface.hpp#L50

atgeirr:
> Well - I agree with this PR; however as I see it the underlying implementation of SimtTimer interface means that nothing is solved:

This is not a big problem for the current code, it is just a default implementation that will be overridden in subclasses if appropriate (and is overridden in the adaptive timer). However I agree that in general one should not provide default implementations in interface classes (and I'd be happy to see it gone here as well).

joakim-hove:
> [ ... ]it is just a default implementation that will be overridden in subclasses if appropriate (and is overridden in the adaptive timer)

Well - to me it does not seem to be overridden in `flow`: https://github.com/OPM/opm-autodiff/blob/master/examples/flow.cpp#L405?

Feel free to ignore if I am misunderstanding here.

atgeirr:
> Well - to me it does not seem to be overridden in flow: https://github.com/OPM/opm-autodiff/blob/master/examples/flow.cpp#L405?

It is true that the main() function deals with the regular timer (and for this timer step == report step). The adaptive timer is created and used inside the run() function of the Simulator class, and there one needs to separate the two concepts. However that is an implementation detail from the point of view of main().

Issue #901, opened 2015-10-09
Missing extrapolation of unsaturated data for PVTG (live gas)


atgeirr:
Apparently we do not currently extrapolate to data points without explicitly given unsaturated data. This is done for PVTO, but not PVTG.

bska:
> Apparently we do not currently extrapolate to data points without explicitly given unsaturated data.

That is correct.  We do not do this.

> This is done for PVTO, but not PVTG.

Also correct.  The trouble, however, is that there is no easy algorithm for expanding the PVTG data according to the requirement that expansion satisfy piecewise constant compressibility.  The main reason is that undersaturated PVTG data is tabulated in terms of decreasing Rv at constant pressure whereas undersaturated PVTO is tabulated in terms of increasing Po at constant Rs.  This means that pressure derivatives are trivially accessible in PVTO but not in PVTG.

atgeirr:
Seems to be resolved, closing.

PR #906, opened 2015-10-20
Complete undersaturated PVTG tables by extrapolating from existing data

totto82:
This PR copies equivalent code from PvtLiveOil

### DIFF
--- a/opm/core/props/pvt/PvtLiveGas.cpp
+++ b/opm/core/props/pvt/PvtLiveGas.cpp
@@ -11,6 +11,7 @@
 //===========================================================================
 /*
   Copyright 2010 SINTEF ICT, Applied Mathematics.
+  Copyright 2015 IRIS AS
 
   This file is part of the Open Porous Media project (OPM).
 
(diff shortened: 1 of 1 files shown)


totto82:
With this PR SPE3CASE1 and SPE3CASE2 matches eclipse.

opm/core/props/pvt/PvtLiveGas.cpp
```diff
@@ -92,6 +93,44 @@ namespace Opm
                     undersat_gas_tables_[pvtTableIdx][i][1][j] = 1.0 / undersat_gas_tables_[pvtTableIdx][i][1][j];
                 }
             }
+
+             // Complete undersaturated tables by extrapolating from existing data
+            int iNext = -1;
+            for (int i=0; i<sz; ++i) {
```
atgeirr: Spaces around `=` and `<` please.

opm/core/props/pvt/PvtLiveGas.cpp
```diff
@@ -92,6 +93,44 @@ namespace Opm
...
+                if (undersat_gas_tables_[pvtTableIdx][i][0].size() > 1) {
+                    continue;
+                }
+                // Look ahead for next record containing undersaturated data
+                if (iNext < i) {
+                    iNext = i+1;
+                    while (iNext < sz && undersat_gas_tables_[pvtTableIdx][iNext][0].size() < 2) {
+                        ++iNext;
+                    }
+                    if (iNext == sz) OPM_THROW(std::runtime_error,"Unable to complete undersaturated table.");
```
atgeirr: Use braces always for if() clauses.

opm/core/props/pvt/PvtLiveGas.cpp
```diff
@@ -92,6 +93,44 @@ namespace Opm
...
+                }
+                // Add undersaturated data to current record while maintaining compressibility and viscosibility
+                // TODO: How to add 1/(B*mu) in this way?
+                typedef std::vector<std::vector<std::vector<double> > >::size_type sz_t;
```
atgeirr: A typedef to remember :-)

(No change required, just found it funny!)

opm/core/props/pvt/PvtLiveGas.cpp
```diff
@@ -92,6 +93,44 @@ namespace Opm
...
+                }
+                // Look ahead for next record containing undersaturated data
+                if (iNext < i) {
+                    iNext = i+1;
+                    while (iNext < sz && undersat_gas_tables_[pvtTableIdx][iNext][0].size() < 2) {
+                        ++iNext;
+                    }
+                    if (iNext == sz) OPM_THROW(std::runtime_error,"Unable to complete undersaturated table.");
+                }
+                // Add undersaturated data to current record while maintaining compressibility and viscosibility
```
atgeirr: Is this comment correct? It is not obvious to me that this would be conserved.

opm/core/props/pvt/PvtLiveGas.cpp
```diff
@@ -92,6 +93,44 @@ namespace Opm
...
+                // TODO: How to add 1/(B*mu) in this way?
+                typedef std::vector<std::vector<std::vector<double> > >::size_type sz_t;
+                for (sz_t j=1; j<undersat_gas_tables_[pvtTableIdx][iNext][0].size(); ++j) {
+                    double diffSolubility = undersat_gas_tables_[pvtTableIdx][iNext][0][j] - undersat_gas_tables_[pvtTableIdx][iNext][0][j-1];
```
atgeirr: To make the code more readable, I suggest the following modifications:
```
auto& from_table = undersat_gas_tables_[pvtTableIdx][iNext];
auto& to_table = undersat_gas_tables_[pvtTableIdx][i];
```
Also the use of { 0, 1, 2, 3 } as indices directly is not easy to understand unless you already know the code, I suggest adding a little enum giving these indices names (like Rs, B, Mu, InvBMu).

atgeirr:
Thanks for the contribution! Judging from the test results it looks like it is working properly, and I will merge this when the minor issues are handled.

opm/core/props/pvt/PvtLiveGas.cpp
totto82: Not for me as well. jf.  #901. I will add a warning in the comments.

totto82:
First,  I totally agree with all your suggestions. But as I said this PR copies the code in PvtLiveOil. Most of your comments thus also applies for the PvtLiveOil implementation. I will do the fixes you suggest for PvtLiveGas, but on the long term I think we need to restructure and clean the whole pvt implementation.

atgeirr:
> Most of your comments thus also applies for the PvtLiveOil implementation

True, just taking the opportunity to improve since this is a new use. Then this can be used as a pattern to improve the oil version later.

andlaus:
> on the long term I think we need to restructure and clean the whole pvt implementation. 

+1. I also volunteer to do the bulk of this. (it can be done almost in the same way as the satfunc refactoring, but it will be simpler because opm-core already depends on opm-material.)

atgeirr:
Thanks for the update, merging!

PR #907, opened 2015-10-20
implement determining the threshold pressure from the initial condition

andlaus:
This needs to be done if a equilibration region transition is
mentioned by the THPRES keyword, but no value is given for this record
in the third item. (it seems that this is used quite frequently.)

Also, the approach taken by this patch also does not collide with the
restart machinery as far as I can see. This is because the initial
condition is applied by the simulator before the state at the restart
time is loaded. (I interpreted the code that way, but I could be
wrong, could anyone verify this?)

since it is pretty elaborate to calculate initial condition, this
patch is pretty messy. I also do not know if Eclipse does include
capillary pressure in this calculation or not (this patch does) and
there may be other bugs lurking. happy testing!

### DIFF
--- a/opm/core/utility/thresholdPressures.hpp
+++ b/opm/core/utility/thresholdPressures.hpp
@@ -66,11 +66,11 @@ namespace Opm
         if (simulationConfig->hasThresholdPressure()) {
             std::shared_ptr<const ThresholdPressure> thresholdPressure = simulationConfig->getThresholdPressure();
             std::shared_ptr<GridProperty<int>> eqlnum = eclipseState->getIntGridProperty("EQLNUM");
-            auto eqlnumData = eqlnum->getData();
+            const auto& eqlnumData = eqlnum->getData();
 
-            int numPhases = initialState.numPhases();
-            int numCells = UgGridHelpers::numCells(grid);
-            int numPvtRegions = deck->getKeyword("TABDIMS")->getRecord(0)->getItem("NTPVT")->getInt(0);
+            const int numPhases = initialState.numPhases();
+            const int numCells = UgGridHelpers::numCells(grid);
+            const int numPvtRegions = deck->getKeyword("TABDIMS")->getRecord(0)->getItem("NTPVT")->getInt(0);
 
             // retrieve the minimum (residual!?) and the maximum saturations for all cells
             std::vector<double> minSat(numPhases*numCells);
@@ -88,19 +88,19 @@ namespace Opm
                 surfaceDensity[regionIdx].resize(numPhases);
 
                 if (pu.phase_used[BlackoilPhases::Aqua]) {
-                    int wpos = pu.phase_pos[BlackoilPhases::Aqua];
+                    const int wpos = pu.phase_pos[BlackoilPhases::Aqua];
                     surfaceDensity[regionIdx][wpos] =
                         densityKw->getRecord(regionIdx)->getItem("WATER")->getSIDouble(0);
                 }
 
                 if (pu.phase_used[BlackoilPhases::Liquid]) {
-                    int opos = pu.phase_pos[BlackoilPhases::Liquid];
+                    const int opos = pu.phase_pos[BlackoilPhases::Liquid];
                     surfaceDensity[regionIdx][opos] =
                         densityKw->getRecord(regionIdx)->getItem("OIL")->getSIDouble(0);
                 }
 
                 if (pu.phase_used[BlackoilPhases::Vapour]) {
-                    int gpos = pu.phase_pos[BlackoilPhases::Vapour];
+                    const int gpos = pu.phase_pos[BlackoilPhases::Vapour];
                     surfaceDensity[regionIdx][gpos] =
                         densityKw->getRecord(regionIdx)->getItem("GAS")->getSIDouble(0);
                 }
@@ -112,7 +112,7 @@ namespace Opm
             std::vector<int> pvtRegion(numCells);
             const auto& cartPvtRegion = eclipseState->getIntGridProperty("PVTNUM")->getData();
             for (int cellIdx = 0; cellIdx < numCells; ++cellIdx) {
-                int cartCellIdx = gc?gc[cellIdx]:cellIdx;
+                const int cartCellIdx = gc ? gc[cellIdx] : cellIdx;
                 pvtRegion[cellIdx] = std::max(0, cartPvtRegion[cartCellIdx] - 1);
             }
 
@@ -162,16 +162,16 @@ namespace Opm
 
             for (int cellIdx = 0; cellIdx < numCells; ++ cellIdx) {
                 assert(pu.phase_used[BlackoilPhases::Liquid]); // we currently hard-code the oil phase as the reference phase!
-                int opos = pu.phase_pos[BlackoilPhases::Liquid];
+                const int opos = pu.phase_pos[BlackoilPhases::Liquid];
                 phasePressure[opos][cellIdx] = initialState.pressure()[cellIdx];
 
                 if (pu.phase_used[BlackoilPhases::Aqua]) {
-                    int wpos = pu.phase_pos[BlackoilPhases::Aqua];
+                    const int wpos = pu.phase_pos[BlackoilPhases::Aqua];
                     phasePressure[wpos][cellIdx] = initialState.pressure()[cellIdx] + (capPress[cellIdx*numPhases + opos] - capPress[cellIdx*numPhases + wpos]);
                 }
 
                 if (pu.phase_used[BlackoilPhases::Vapour]) {
-                    int gpos = pu.phase_pos[BlackoilPhases::Vapour];
+                    const int gpos = pu.phase_pos[BlackoilPhases::Vapour];
                     phasePressure[gpos][cellIdx] = initialState.pressure()[cellIdx] + (capPress[cellIdx*numPhases + gpos] - capPress[cellIdx*numPhases + opos]);
                 }
             }
@@ -179,7 +179,7 @@ namespace Opm
             // calculate the inverse formation volume factors for the active phases and each cell
             if (pu.phase_used[BlackoilPhases::Aqua]) {
                 std::vector<double> dummy(numCells*BlackoilPhases::MaxNumPhases);
-                int wpos = pu.phase_pos[BlackoilPhases::Aqua];
+                const int wpos = pu.phase_pos[BlackoilPhases::Aqua];
                 const PvtInterface& pvtw = props.pvt(wpos);
                 pvtw.b(numCells,
                        pvtRegion.data(),
@@ -197,7 +197,7 @@ namespace Opm
 
             if (pu.phase_used[BlackoilPhases::Liquid]) {
                 std::vector<double> dummy(numCells*BlackoilPhases::MaxNumPhases);
-                int opos = pu.phase_pos[BlackoilPhases::Liquid];
+                const int opos = pu.phase_pos[BlackoilPhases::Liquid];
                 const PvtInterface& pvto = props.pvt(opos);
                 pvto.b(numCells,
                        pvtRegion.data(),
@@ -221,7 +221,7 @@ namespace Opm
 
             if (pu.phase_used[BlackoilPhases::Vapour]) {
                 std::vector<double> dummy(numCells*BlackoilPhases::MaxNumPhases);
-                int gpos = pu.phase_pos[BlackoilPhases::Vapour];
+                const int gpos = pu.phase_pos[BlackoilPhases::Vapour];
                 const PvtInterface& pvtg = props.pvt(gpos);
                 pvtg.b(numCells,
                        pvtRegion.data(),
@@ -236,7 +236,7 @@ namespace Opm
                     rho[gpos][cellIdx] = surfaceDensity[pvtRegion[cellIdx]][gpos]*b[cellIdx];
 
                     if (pu.phase_used[BlackoilPhases::Liquid]) {
-                        int opos = pu.phase_pos[BlackoilPhases::Liquid];
+                        const int opos = pu.phase_pos[BlackoilPhases::Liquid];
                         rho[gpos][cellIdx] +=
                             surfaceDensity[pvtRegion[cellIdx]][opos]*initialState.rv()[cellIdx]*b[cellIdx];
                     }
@@ -248,7 +248,7 @@ namespace Opm
             std::map<std::pair<int, int>, double> maxDp;
             const int num_faces = UgGridHelpers::numFaces(grid);
             thpres_vals.resize(num_faces, 0.0);
-            auto fc = UgGridHelpers::faceCells(grid);
+            const auto& fc = UgGridHelpers::faceCells(grid);
             for (int face = 0; face < num_faces; ++face) {
                 const int c1 = fc(face, 0);
                 const int c2 = fc(face, 1);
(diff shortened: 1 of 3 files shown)


atgeirr:
Thank you for the patch(es)! I intend to review and merge them eventually, but that will wait until after the release, unless there are significant reasons to make an exception for this (new variant) feature.

andlaus:
yeah, I agree. I forgot to mention that I consider this to be post-release stuff.

andlaus:
I suppose the same argument as for OPM/opm-material#81 applies here: the release is not affected anymore, so the PR can savely be merged (after review of course).

atgeirr:
> the release is not affected anymore, so the PR can savely be merged (after review of course).

Intend to review after release.

totto82:
I have compared the results with Eclipse. The results does not match. I will send you an email with the comparison.

opm/core/utility/thresholdPressures.hpp
```diff
@@ -57,7 +68,154 @@ namespace Opm
...
+            props.capPress(numCells, initialState.saturation().data(), cellIdxArray.data(), capPress.data(), NULL);
+
+            // compute the absolute pressure of each active phase
+            std::vector<std::vector<double> > phasePressure(numPhases);
+            assert(pu.phase_used[BlackoilPhases::Liquid]); // we currently hard-code the oil phase as the reference phase!
+            int refPhaseIdx = pu.phase_pos[BlackoilPhases::Liquid];
+            for (int phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+                phasePressure[phaseIdx].resize(numCells);
+                for (int cellIdx = 0; cellIdx < numCells; ++cellIdx) {
+                    phasePressure[phaseIdx][cellIdx] = initialState.pressure()[cellIdx] + (capPress[cellIdx*numPhases + phaseIdx] - capPress[cellIdx*numPhases + refPhaseIdx]);
```
totto82: Is the sign correct for the water phase?

opm/core/utility/thresholdPressures.hpp
```diff
@@ -57,7 +68,154 @@ namespace Opm
...
+                       initialState.gasoilratio().data(),
+                       cond.data(),
+                       b.data(),
+                       dummy.data(),
+                       dummy.data());
+                for (int cellIdx = 0; cellIdx < numCells; ++ cellIdx) {
+                    rho[opos][cellIdx] = surfaceDensity[pvtRegion[cellIdx]][opos]*b[cellIdx];
+
+                    if (pu.phase_used[BlackoilPhases::Vapour]) {
+                        rho[opos][cellIdx] +=
```
totto82: Shouldn't it be: 
```rho[opos][cellIdx] += surfaceDensity[pvtRegion[cellIdx]][gpos] *
initialState.gasoilratio()[cellIdx]*b[cellIdx];```

File: opm/core/utility/thresholdPressures.hpp [hunk 3]
```diff
@@ -57,7 +68,154 @@ namespace Opm
...
+                       cond.data(),
+                       b.data(),
+                       dummy.data(),
+                       dummy.data());
+                for (int cellIdx = 0; cellIdx < numCells; ++ cellIdx) {
+                    rho[gpos][cellIdx] = surfaceDensity[pvtRegion[cellIdx]][gpos]*b[cellIdx];
+
+                    if (pu.phase_used[BlackoilPhases::Vapour]) {
+                        rho[gpos][cellIdx] +=
+                            initialState.rv()[cellIdx]*b[cellIdx];
```
totto82: surfaceDensity[pvtRegion[cellIdx]][opos] is missing ?

opm/core/utility/thresholdPressures.hpp
```diff
@@ -74,12 +232,57 @@ namespace Opm
...
+                const int c2 = fc(face, 1);
+                if (c1 < 0 || c2 < 0) {
+                    // Boundary face, skip it.
+                    continue;
+                }
+                const int gc1 = (gc == 0) ? c1 : gc[c1];
+                const int gc2 = (gc == 0) ? c2 : gc[c2];
+                const int eq1 = eqlnumData[gc1];
+                const int eq2 = eqlnumData[gc2];
+
```
totto82: I think only phases present in the upstream cell should be considered?

opm/core/utility/thresholdPressures.hpp
andlaus: I'll have to check. (will do this tomorrow.) for opm-material it would be correct. opm-core is a different beast, though...

opm/core/utility/thresholdPressures.hpp
andlaus: true.

opm/core/utility/thresholdPressures.hpp
andlaus: for gravity correction you also need to calculate the density of the downstream cell, right?

andlaus:
ok, I added a fixup for the missing surface densities and the capillary sign of the capillary pressure.

opm/core/utility/thresholdPressures.hpp
```diff
@@ -74,12 +245,57 @@ namespace Opm
...
+                for (int phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+                    double z1 = UgGridHelpers::cellCenterDepth(grid, c1);
+                    double z2 = UgGridHelpers::cellCenterDepth(grid, c2);
+                    double zAvg = (z1 + z2)/2; // average depth
+
+                    // compute gravity corrected pressure potentials at the average depth
+                    double p1 = phasePressure[phaseIdx][c1];
+                    double p2 = phasePressure[phaseIdx][c2];
+                    p1 += rho[phaseIdx][c1]*gravity*(z1 - zAvg);
+                    p2 += rho[phaseIdx][c2]*gravity*(z2 - zAvg);
```
totto82: The average density should be used for both sides (as in the simulator).
Is the sign wrong?

opm/core/utility/thresholdPressures.hpp
totto82: What I mean is something like: 
  ```
double s1 = initialState.saturation()[numPhases*c1 + phaseIdx];
double s2 = initialState.saturation()[numPhases*c2 + phaseIdx];
if ( (p1 > p2 && s1 > 0) || (p2 > p1 && s2 > 0) ) {
    maxDp[barrierId] = std::max(maxDp[barrierId], std::abs(p1 - p2));
 }
```

Comment (andlaus) [inline opm/core/utility/thresholdPressures.hpp]: Ok, I see what you mean. if the density contains a meaningful value it should work as is I think.

Also, only doing this if there is saturation in the upstream cell means that the phase is discarded when it comes to calculating threshold pressure. is that desireable? (To me be both ways seem to be meaningful physically, but what does E100?)

File: opm/core/utility/thresholdPressures.hpp (hunk 5 above)
Comment (andlaus) [inline opm/core/utility/thresholdPressures.hpp]: hm, I'm not sure if that's a good idea: all values (this includes density) are constant within FV cells, right? is there an effect if this is changed for the conservation equations?

Comment (andlaus) [inline opm/core/utility/thresholdPressures.hpp]: also, the sign should be correct: if the depth of the cell is larger than the average, the pressure is increased, if it is smaller, pressure gets decreased. (for positive gravity constants, which seems to be the case in flow.)

File: opm/core/utility/thresholdPressures.hpp (hunk 4 above)
Comment (totto82) [inline opm/core/utility/thresholdPressures.hpp]: Some more reverse engineering point in the direction of using smin instead of 0

File: opm/core/utility/thresholdPressures.hpp [hunk 6]
```diff
@@ -57,7 +68,167 @@ namespace Opm
...
+                       b.data(),
+                       dummy.data(),
+                       dummy.data());
+                for (int cellIdx = 0; cellIdx < numCells; ++ cellIdx) {
+                    rho[opos][cellIdx] = surfaceDensity[pvtRegion[cellIdx]][opos]*b[cellIdx];
+
+                    if (pu.phase_used[BlackoilPhases::Vapour]) {
+                        int gpos = pu.phase_pos[BlackoilPhases::Vapour];
+                        rho[opos][cellIdx] +=
+                            surfaceDensity[pvtRegion[cellIdx]][gpos]*initialState.surfacevol()[cellIdx]*b[cellIdx];
```
totto82: This is still wrong.

opm/core/utility/thresholdPressures.hpp
totto82: The purpose of the thpress is to avoid flow between regions if not an external force is present. Therefore I will argue that we should use the same formulation as computeMassFlux in BlackoilModeBase_impl.hpp.

opm/core/utility/thresholdPressures.hpp
totto82: According to the manual differences in the oil phase should not effect the thpress in the water zone and so on. I guess the only way to figure this out is to test it.

andlaus: ok, I've updated the PR to consider the saturation of the upstream cell and use the average phase density.

opm/core/utility/thresholdPressures.hpp
andlaus: ok, that's an argument in favour of the average density. (agreed.) But the sign should be correct?

opm/core/utility/thresholdPressures.hpp
andlaus: *arg*: the names `surfacevolume()` and `gasoilratio()` are  actually worse than `foo()` and `bar()` would be: for the latter you'd actually be forced to look up what they mean instead of being mislead by the method name.

anyway, fixed. thanks.

andlaus:
ok, I've squashed the fixups and rebased it on top of the latest master version. If we're lucky, it can be merged now, but @totto82 will do some more tests.

opm/core/utility/thresholdPressures.hpp
totto82: Based on testing, I think this is more correct. 
```
 double smin1 = smin[numPhases*c1 + phaseIdx];
 double smin2 = smin[numPhases*c2 + phaseIdx];

 if ( (p1 > p2 && s1 > smin1) || (p2 > p1 && s2 > smin2) ) { ... } 
```

totto82:
I am able to reproduce Eclipse with this code, but then I have to use       
UgGridHelpers::cellCenterDepth(G, cell) instead of            
UgGridHelpers::cellCentroidCoordinate(G, cell, 2) 
in the initialization as well. I think cellCenterDepth is more similar to Eclipse and should be used in the initialization. I will test it and make a separate PR for this.

opm/core/utility/thresholdPressures.hpp
andlaus: ok, changed.

atgeirr:
Since the release is done, I have now looked at this.

I cannot argue with the fact that this code is tested to work correctly, if I understand @totto82. It does look like it contains a lot of pasted code from various places, though. I initially believed that the output from the initialization process would be enough (that the  fluid object for example would not be needed), I assume you have verified that such an approach would not be sufficient.

Given that it *is* needed, we must approach this as it is done in the model, otherwise we risk inconsistencies. But duplicating all that is needed for the calculations is a big and error-prone thing. Would it then not be far better to move this part to the model class itself, where the same function could be used for both purposes (determining the threshold pressures as well as the phase head differences and resulting fluxes)? This could also make the handling of NNCs easier.

andlaus:
> I initially believed that the output from the initialization process would be enough (that the fluid object for example would not be needed), I assume you have verified that such an approach would not be sufficient.

I have indeed verified that it is insufficient. (because I was quite annoyed by the fact that the phase densities need to be re-calculated from scratch at any place where they are required ;).) Long-term, I think it would be beneficial to split the "reservoir state" into a "primal state" which features only the primary variables on which the non-linear and linear solvers operate on and a "secondary state" which provides all quantities required for calculating the residual that are derived from the "primal" quantities. The "secondary state" would indeed be sufficient for this functionality.
 
> Given that it is needed, we must approach this as it is done in the model, otherwise we risk inconsistencies. Would it then not be far better to move this part to the model class itself, where the same function could be used for both purposes (determining the threshold pressures as well as the phase head differences and resulting fluxes)? This could also make the handling of NNCs easier.

I certainly agree that this patch is messy and prone to inconsistencies, but in practice your suggestion implies moving the `thresholdPressures()` functions out of opm-core into opm-autodiff and -- probably more relevantly -- as far as I can see the duplication would not vanish but just move to a different module. (because gravity correction is "hard coded" into the assembly process and thus not directly available for the initialization code.)

Finally, the code duplication and consistency arguments equally apply to the `initStateEquil()` functions in opm-core?

atgeirr:
Thanks for your explanations, @andlaus. The discussion of reservoir state is interesting, I'll not continue that here however. Your comment about initStateEquil() also being a consistency risk makes me think that perhaps the right thing to do in the long term is to move all these things into the model. With different models, the equations are different and everything that depends on the equations (including initialization) should therefore be part of the model (system, if not class).

I'll review the code, with intention to merge.

andlaus:
> Your comment about initStateEquil() also being a consistency risk makes me think that perhaps the right thing to do in the long term is to move all these things into the model.

basically that's another argument of abandoning opm-core ;)

> With different models, the equations are different and everything that depends on the equations (including initialization) should therefore be part of the model (system, if not class).

that's the reason why opm-material has the "fluid state" concept: using this, each model can make whatever assumptions are the most appropriate for it, but generic code does not need to care about these details. maybe you should have a closer look to see if something similar would be beneficial for opm-autodiff.

opm/core/utility/thresholdPressures.hpp
```diff
@@ -57,10 +68,186 @@ namespace Opm
             std::shared_ptr<GridProperty<int>> eqlnum = eclipseState->getIntGridProperty("EQLNUM");
             auto eqlnumData = eqlnum->getData();
 
-            // Set values for each face.
+            int numPhases = initialState.numPhases();
```
atgeirr: I prefer that such variables are const, I've gotten so used to it that when they are *not* const I look for where they are changed!

opm/core/utility/thresholdPressures.hpp
```diff
@@ -57,10 +68,186 @@ namespace Opm
...
+                }
+            }
+
+            // retrieve the PVT region of each cell. note that we need c++ instead of
+            // Fortran indices.
+            const int* gc = UgGridHelpers::globalCell(grid);
+            std::vector<int> pvtRegion(numCells);
+            const auto& cartPvtRegion = eclipseState->getIntGridProperty("PVTNUM")->getData();
+            for (int cellIdx = 0; cellIdx < numCells; ++cellIdx) {
+                int cartCellIdx = gc?gc[cellIdx]:cellIdx;
```
atgeirr: Please make the variable const, and add spaces around `?` and `:`.

opm/core/utility/thresholdPressures.hpp
```diff
@@ -57,10 +68,186 @@ namespace Opm
...
+                pvtRegion[cellIdx] = std::max(0, cartPvtRegion[cartCellIdx] - 1);
+            }
+
+            // compute the initial "phase presence" of each cell (required to calculate
+            // the inverse formation volume factors
+            std::vector<PhasePresence> cond(numCells);
+            for (int cellIdx = 0; cellIdx < numCells; ++cellIdx) {
+                if (pu.phase_used[BlackoilPhases::Aqua] && initialState.saturation()[numPhases*cellIdx + pu.phase_pos[BlackoilPhases::Aqua]] > 0.0) {
```
atgeirr: These lines are unreadably long, so I misunderstood a little at first. A reference alias or twou should make it much easier to read.

opm/core/utility/thresholdPressures.hpp
```diff
@@ -57,10 +68,186 @@ namespace Opm
...
+            }
+
+            for (int cellIdx = 0; cellIdx < numCells; ++ cellIdx) {
+                assert(pu.phase_used[BlackoilPhases::Liquid]); // we currently hard-code the oil phase as the reference phase!
+                int opos = pu.phase_pos[BlackoilPhases::Liquid];
+                phasePressure[opos][cellIdx] = initialState.pressure()[cellIdx];
+
+                if (pu.phase_used[BlackoilPhases::Aqua]) {
+                    int wpos = pu.phase_pos[BlackoilPhases::Aqua];
+                    phasePressure[wpos][cellIdx] = initialState.pressure()[cellIdx] + (capPress[cellIdx*numPhases + opos] - capPress[cellIdx*numPhases + wpos]);
```
atgeirr: Again, very long lines!

opm/core/utility/thresholdPressures.hpp
```diff
@@ -57,10 +68,186 @@ namespace Opm
...
+            // done to always have a positive number for the capillary pressure (as long
+            // as the medium is hydrophilic)
+            std::vector<std::vector<double> > phasePressure(numPhases);
+            for (int phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
+                phasePressure[phaseIdx].resize(numCells);
+            }
+
+            for (int cellIdx = 0; cellIdx < numCells; ++ cellIdx) {
+                assert(pu.phase_used[BlackoilPhases::Liquid]); // we currently hard-code the oil phase as the reference phase!
+                int opos = pu.phase_pos[BlackoilPhases::Liquid];
```
atgeirr: Could (and therefore should) be const.

atgeirr:
While I have made a few nitpicky remarks as usual, I find the code good in general. There is one change I will request though: to put the computation of maxDp into its own function. That will allow the (simple) logic of the thresholdPressure() function to be obvious to the reader (now it is at the bottom of a long function).

andlaus:
ok, all variables have been `const`ified in that functions and long lines have been split.

atgeirr:
Still missing separation of maxDp calculation into its own function.

andlaus:
> Still missing separation of maxDp calculation into its own function.

this took a few minutes longer. now it's done. don't forget to also merge OPM/opm-autodiff#506 and OPM/opm-polymer#151

atgeirr:
Thanks for the work, merging.

PR #910, opened 2015-10-21
Fix twophase sim

atgeirr:
This PR makes two-phase runs with sim_2p_incomp work again as intended.

### DIFF
--- a/examples/sim_2p_incomp.cpp
+++ b/examples/sim_2p_incomp.cpp
@@ -249,15 +249,17 @@ try
         // Use timer for last epoch to obtain total time.
         simtimer.init(timeMap);
         const double total_time = simtimer.totalTime();
-        for (size_t reportStepIdx = 0; reportStepIdx < timeMap->numTimesteps(); ++reportStepIdx) {
+        // for (size_t reportStepIdx = 0; reportStepIdx < timeMap->numTimesteps(); ++reportStepIdx) {
+        size_t reportStepIdx = 0; // Only handle a single, unchanging well setup.
+        {
             // Update the timer.
             simtimer.setCurrentStepNum(step);
             simtimer.setTotalTime(total_time);
 
             // Report on start of report step.
-            std::cout << "\n\n--------------    Starting report step " << reportStepIdx << "    --------------"
-                      << "\n                  (number of time steps: "
-                      << simtimer.numSteps() - step << ")\n\n" << std::flush;
+            // std::cout << "\n\n--------------    Starting report step " << reportStepIdx << "    --------------"
+            //           << "\n                  (number of time steps: "
+            //           << simtimer.numSteps() - step << ")\n\n" << std::flush;
 
             // Create new wells, well_state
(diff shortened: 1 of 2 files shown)


andlaus:
this looks good, but I can neither run my 2p example (SAIGUP) with flow:

```
> ./bin/flow deck_filename=~/tmp/saigup/PRUESAIDAV.DATA
**********************************************************************
*                                                                    *
*                   This is Flow (version 2015.04)                   *
*                                                                    *
* Flow is a simulator for fully implicit three-phase black-oil flow, *
*            and is part of OPM. For more information see:           *
*                       http://opm-project.org                       *
*                                                                    *
**********************************************************************

---------------    Reading parameters     ---------------
deck_filename found at /, value is /home/and/tmp/saigup/PRUESAIDAV.DATA
output not found. Using default value 'true'.
output_dir not found. Using default value 'output'.
pvt_tab_size not found. Using default value '-1'.
threephase_model not found. Using default value 'gwseg'.
Program threw an exception: [/home/and/src/opm-core/opm/core/simulator/initState_impl.hpp:738] initBlackoilSurfvol() is only supported in three-phase simulations.
```

nor with `sim_2p_incomp_ad`:

```
> ./bin/sim_2p_incomp_ad deck_filename=~/tmp/saigup/PRUESAIDAV.DATA

================    Test program for incompressible two-phase flow     ===============

---------------    Reading parameters     ---------------
deck_filename found at /, value is /home/and/tmp/saigup/PRUESAIDAV.DATA
use_pside not found. Using default value 'false'.
linsolver not found. Using default value 'umfpack'.
output not found. Using default value 'true'.
output_dir not found. Using default value 'output'.


================    Starting main simulation loop     ===============


--------------    Starting report step 0    --------------
                  (number of steps left: 100)

transport_solver_type not found. Using default value 'ad'.
use_segregation_split not found. Using default value 'false'.
nl_pressure_residual_tolerance not found. Using default value '0'.
nl_pressure_change_tolerance not found. Using default value '1'.
nl_pressure_maxiter not found. Using default value '10'.
Program threw an exception: [/home/and/src/opm-autodiff/opm/autodiff/SimulatorIncompTwophaseAd.cpp:364] The implicit ad transport solver cannot handle rock compressibility.
terminate called after throwing an instance of 'std::runtime_error'
  what():  [/home/and/src/opm-autodiff/opm/autodiff/SimulatorIncompTwophaseAd.cpp:364] The implicit ad transport solver cannot handle rock compressibility.
Aborted
```

the non-AD version of `sim_2p_incomp` from opm-core seems to work, as do the SPE test cases. since this is not a regression in any case, I'll merge this.

atgeirr:
I think for Flow, we should really make it run with two phases, but that functionality has been dead for a year and a half (or so), there is an issue about it... It seems to me that sim_2p_incomp_ad throws because of rock compressibility, my guess is that it is simply ignored by the others. There are lots of refinements to be made for these simpler simulators if we find the time for it!

PR #917, opened 2015-10-26
[bugfix] Fixes copyOwnerToAll.

blattms:
Previously, we copied owner/overlap to all which is clearly wrong.
Now we copy from owner to all as the function name says.

As this is a bugfix and it would make sense to include it in the release.


atgeirr:
Thanks, will merge.

PR #918, opened 2015-10-26
Allow the rock properties to be zero initialized without a deck.

blattms:
This is needed in parallel runs where a the rock properties will not
be read from the deck but be communicated from a master process. Nevertheless
we need to be able to initialize the data structures with the correct
container size. In addition we need to be able to change the container values
from opm-autodiff's BlackoilPropsDataHandle.

### DIFF
--- a/opm/core/props/rock/RockFromDeck.cpp
+++ b/opm/core/props/rock/RockFromDeck.cpp
@@ -72,6 +72,13 @@ namespace Opm
     {
     }
 
+    RockFromDeck::RockFromDeck(std::size_t number_of_cells)
+        : porosity_(number_of_cells, 0),
+          permeability_(number_of_cells*9, 0.0),
+          permfield_valid_(number_of_cells, false)
+    {
+    }
+
     void RockFromDeck::init(Opm::EclipseStateConstPtr eclState,
                             int number_of_cells, const int* global_cell,
                             const int* cart_dims)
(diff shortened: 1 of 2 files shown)


opm/core/props/rock/RockFromDeck.cpp
```diff
@@ -72,6 +72,13 @@ namespace Opm
     {
     }
 
+    RockFromDeck::RockFromDeck(std::size_t number_of_cells)
+        : porosity_(number_of_cells, 0),
+          permeability_(number_of_cells*9, 0.0),
```
atgeirr: A comment explaining the 9 (room for full tensor in 3D represented as a matrix) would be good.

opm/core/props/rock/RockFromDeck.hpp
```diff
@@ -31,10 +31,16 @@ namespace Opm
...
+        // access to porosity and permeability
+        friend class BlackoilPropsDataHandle;
+
     public:
         /// Default constructor.
         RockFromDeck();
-
+        /// Creates rock properties with zero porosity and permeability
+        /// \param number_of_cells The number of cells
+        RockFromDeck(std::size_t number_of_cells);
```
atgeirr: Should be explicit.

opm/core/props/rock/RockFromDeck.cpp
blattms: Comment added. Pease note that 9 is used throught the file at various other places.

opm/core/props/rock/RockFromDeck.hpp
blattms: I always miss this one. Thanks for the hint. Change applied

atgeirr:
Thanks! Merging.

PR #919, opened 2015-10-27
Fixes well manager for wells crossing into cells that overlap.

blattms:
In the parallel run there are cases where wells perforate cells
that are neighbors of overlap/halo cells. On other process only
parts of the well are seen as perforations. These wells should be
ignored there. While the well was indeed ignored, the perforations
found where mistakenly added to the well found due not clearing the
wellperf_data[well_index]. This commit now does this clearing and
results in the right handling of wells for e.g. SPE9.

### DIFF
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -231,6 +231,7 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
                                  << "process deactivating here." << std::endl;
                         // Mark well as not existent on this process
                         wells_on_proc[wellIter-wells.begin()] = 0;
+                        wellperf_data[well_index].clear();
                         continue;
                     }
                 }
--- a/opm/core/wells/WellsManager_impl.hpp
+++ b/opm/core/wells/WellsManager_impl.hpp
@@ -128,7 +128,11 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
     wellperf_data.resize(wells.size());
     wells_on_proc.resize(wells.size(), 1);
 
-    int well_index = 0;
+    // The well index on the current process.
+    // Note that some wells are deactivated as they live on the interior
+    // domain of another proccess. Therefore this might different from
+    // the index of the well according to the eclipse state
+    int well_index_on_proc = 0;
     for (auto wellIter= wells.begin(); wellIter != wells.end(); ++wellIter) {
         WellConstPtr well = (*wellIter);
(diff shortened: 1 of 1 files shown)


blattms:
Considering that this fixes a serious bug, it might be worthwhile to also include it into the release. Your call.

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -226,11 +226,12 @@ void WellsManager::createWellsFromSpecs(std::vector<WellConstPtr>& wells, size_t
...
                     if ( sum_completions_on_proc < completionSet->size() )
                     {
-                        std::cout<< "Well "<< well->name() << " semms not be in "
+                        std::cout<< "Well "<< well->name() << " does not seem to be"
                                  << "completely in the disjoint partition of "
-                                 << "process deactivating here." << std::endl;
+                                 << "process. Therefore we  deactivate it here." << std::endl;
                         // Mark well as not existent on this process
                         wells_on_proc[wellIter-wells.begin()] = 0;
+                        wellperf_data[well_index].clear();
```
atgeirr: It seems to me that `well_index` and `wellIter - wells.begin()` should be the same number. If so, please unify. If not, please explain.

blattms: On Wed, Oct 28, 2015 at 03:29:46AM -0700, Atgeirr Flø Rasmussen wrote:
> >                          // Mark well as not existent on this process
> >                          wells_on_proc[wellIter-wells.begin()] = 0;
> > +                        wellperf_data[well_index].clear();
> 
> It seems to me that `well_index` and `wellIter - wells.begin()` should be the same number. If so, please unify. If not, please explain.

That is the tricky part: They are not. In a parallel run we manually
deactivate/neglect wells that are not on the process's
domain. well_index counts the wells that are not neglected. Therefore
well_index might be smaller than wellIter-wells.begin().

Maybe we should rename well_index to well_stored_on_proc_index or
similar.

atgeirr: Thanks for the explanation! Please add you explanation to the code. (Best place is perhaps at initialization of well_index.)

blattms: comment is now added.

atgeirr: > comment is now added.

I do not see it?

blattms: > I do not see it?

Well, you couldn't as I missed the push to origin. Did that now. Sorry.

atgeirr:
Excellent, merging!

PR #920, opened 2015-10-30
Change the number of integration points in the initialization

totto82:
The number is changed from 100->2000 to increase accuracy.

According to @osae this is necessary to get match for the new model. 

An alternative fix that would require less integration points would be to compute the span of the integration pr. pillar instead and to make the number of integration points dependent on the size of the integration span.


atgeirr:
I'm fine with this, but a little surprised that our tests (that are mostly regression tests) did not trigger?

PR #921, opened 2015-10-31
AdaptiveTimeStepping: allow for more flexible error computation in the future.

dr-robertk:
This PR revises the TimeStepControl interface such that the computation of the 
error || u^n+ - u^n || / || u^n+1 || is forwarded to the Model class. This allows for more flexible and model dependent implementation. A corresponding PR in opm-autodiff is coming.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -49,9 +49,9 @@ namespace Opm {
             {}
 
             /// return || u^n+1 - u^n || / || u^n+1 ||
-            double timeError() const
+            double relativeChange() const
             {
-                return solver_.model().computeTimeError( previous_, current_ );
+                return solver_.model().relativeChange( previous_, current_ );
             }
         };
     }
@@ -195,11 +195,11 @@ namespace Opm {
 
                 // create object to compute the time error, simply forwards the call to the model
                 detail::SolutionTimeErrorSolverWrapper< Solver, State >
-                    timeError( solver, last_state, state );
+                    relativeChange( solver, last_state, state );
 
                 // compute new time step estimate
                 double dtEstimate =
-                    timeStepControl_->computeTimeStepSize( dt, linearIterations, timeError );
+                    timeStepControl_->computeTimeStepSize( dt, linearIterations, relativeChange );
 
                 // limit the growth of the timestep size by the growth factor
                 dtEstimate = std::min( dtEstimate, double(max_growth_ * dt) );
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -31,11 +31,35 @@
 
 namespace Opm {
 
+    namespace detail
+    {
+        template <class Solver, class State>
+        class SolutionTimeErrorSolverWrapper : public SolutionTimeErrorInterface
+        {
+            const Solver& solver_;
+            const State&  previous_;
+            const State&  current_;
+        public:
+            SolutionTimeErrorSolverWrapper( const Solver& solver,
+                                            const State&  previous,
+                                            const State&  current )
+              : solver_( solver ),
+                previous_( previous ),
+                current_( current )
+            {}
+
+            /// return || u^n+1 - u^n || / || u^n+1 ||
+            double timeError() const
+            {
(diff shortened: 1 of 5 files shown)


opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -31,11 +31,35 @@
...
+            SolutionTimeErrorSolverWrapper( const Solver& solver,
+                                            const State&  previous,
+                                            const State&  current )
+              : solver_( solver ),
+                previous_( previous ),
+                current_( current )
+            {}
+
+            /// return || u^n+1 - u^n || / || u^n+1 ||
+            double timeError() const
```
atgeirr: I think timeError() is a strange name for this, would not relativeChange() or something like that be better? I interpret timeError as being an error in a time variable or something like that.

opm/core/simulator/TimeStepControl.hpp
```diff
@@ -82,60 +82,18 @@ namespace Opm
...
-        mutable std::vector<double> p0_;
-        mutable std::vector<double> sat0_;
-
         const double tol_;
         mutable std::vector< double > errors_;
 
         const bool verbose_;
     private:
-        const boost::any parallel_information_;
+    //    const boost::any parallel_information_;
```
atgeirr: Should be removed entirely.

opm/core/simulator/TimeStepControlInterface.hpp
```diff
@@ -19,31 +19,45 @@
...
+#include <opm/core/simulator/SimulatorState.hpp>
 
 namespace Opm
 {
 
     ///////////////////////////////////////////////////////////////////
     ///
-    ///  TimeStepControlInterface 
-    /// 
+    ///  TimeStepControlInterface
```
atgeirr: Wrong name?

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -172,9 +193,13 @@ namespace Opm {
...
                 ++substepTimer;
 
+                // create object to compute the time error, simply forwards the call to the model
+                detail::SolutionTimeErrorSolverWrapper< Solver, State >
+                    timeError( solver, last_state, state );
+
                 // compute new time step estimate
                 double dtEstimate =
-                    timeStepControl_->computeTimeStepSize( dt, linearIterations, state );
+                    timeStepControl_->computeTimeStepSize( dt, linearIterations, timeError );
```
atgeirr: Since the "time error" is always a double, would it not be simpler if there was no SolutionTimeErrorSolver* classes, that the SolutionTimeErrorSolverWrapper timeError() method content was a small method in this class (AdaptiveTimeStepping) and the computeTimeStepSize() methods took a double instead of the STESInterface?

Then this would look like:
```
const double timeError = computeTimeError(solver, last_state, state); // method call
timeStepControl_->computeTimeStepSize(dt, linearIterations, timeError); // passing a double
```
I am not against using classes and inheritance, but in this case I do not see the benefit.

dr-robertk:
I have addressed the concerns. timeError has been renamed to relativeChange. 
It is not a good idea to pass a simple double, because then the relativeChange has to be computed in any case, whereas with my solution it's only computed when needed.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -31,11 +31,35 @@
 
 namespace Opm {
 
+    namespace detail
+    {
+        template <class Solver, class State>
+        class SolutionTimeErrorSolverWrapper : public SolutionTimeErrorInterface
```
bska: What is the `SolutionTimeErrorInterface`?  It does not appear to exist in any OPM source file.

atgeirr: I do not know why I did not encounter a build error when I tested this, I do get one now in any case! Will fix.

bska: > I do not know why I did not encounter a build error when I tested this,

Did you build-test opm-core in isolation?  I only get a build-failure in downstream module opm-autodiff.

atgeirr: Perhaps, but I also built in opm-autodiff (I think). @dr-robertk has suggested that these parts should be moved to opm-autodiff in any case, and I support that. Fix is easy, coming soon.

PR #922, opened 2015-11-02
Return a constant reference to the mask for ParallelISTLInformation::updateOwnerMask

blattms:
Thus we can actually access the mask in external code. This is for
example needed when calculating averages in RateConverter of opm-autodiff.

### DIFF
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -160,7 +160,7 @@ public:
       communicator.free();
     }
     template<class T>
-    void updateOwnerMask(const T& container) const
+    const std::vector<double>& updateOwnerMask(const T& container) const
     {
         if( ! indexSet_ )
         {
@@ -177,6 +177,7 @@ public:
                 }
             }
         }
+        return ownerMask_;
     }
     /// \brief Compute one or more global reductions.
     ///


opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -160,7 +160,7 @@ class ParallelISTLInformation
       communicator.free();
     }
     template<class T>
-    void updateOwnerMask(const T& container) const
+    const std::vector<double>& updateOwnerMask(const T& container) const
```
atgeirr: When reading this, I asked myself: why is the ownerMask_ mutable? I find it a bit strange to have a const method called updateSomething()...

Would it not be better to add a separate getOwnerMask() method? As it is now, I need to (possibly) change the mask just to look at it.

blattms: The mask is there to keep some internal state and prevent recomputing it every time. Basically, it is used to switch of non-interior degrees of freedom (that other processes compute for) during the global reductions. I guess that is just what mutable was invented for. Unfortunately, the size of this container is not known until the first reduction takes place.

In addition, I now need to switch of non-interior in the global averages over regions in the RateConverter in opm-autodiff (before any other reductions), too. I agree that it is kind of hacky, but it prevents one further iteration over a vector.

atgeirr:
Thanks for clarifying!

PR #928, opened 2015-11-13
compatibility of MINPV and PINCH  

totto82:
- adding of the removed pore volume to the above cell is made optional.
- changes is made in the PinchProcessor to account for the change in the MinpvProcessor.
- a method that return the thressholdpressures for the NNCs are added

Note this is not compatible with #907

### DIFF
--- a/opm/core/grid/PinchProcessor.hpp
+++ b/opm/core/grid/PinchProcessor.hpp
@@ -395,6 +395,7 @@ namespace Opm
                             break;
                         }
                     }
+                    break;
                 }
                 pinFaces.push_back(interface_(grid, topCell, Opm::FaceDir::ZPlus));
                 pinCells.push_back(topCell);
@@ -406,11 +407,14 @@ namespace Opm
                         botCell = getGlobalIndex_(ijk2[0], ijk2[1], botk, dims);
                         if (actnum[botCell]) {
                             break;
+
                         }
                     }
+                    break;
                 }
                 pinFaces.push_back(interface_(grid, botCell, Opm::FaceDir::ZMinus));
                 pinCells.push_back(botCell);
+
             }
         }
 
--- a/opm/core/grid/PinchProcessor.hpp
+++ b/opm/core/grid/PinchProcessor.hpp
@@ -395,7 +395,6 @@ namespace Opm
                             break;
                         }
                     }
-                    break;
                 }
                 pinFaces.push_back(interface_(grid, topCell, Opm::FaceDir::ZPlus));
                 pinCells.push_back(topCell);
@@ -410,7 +409,6 @@ namespace Opm
 
                         }
                     }
-                    break;
                 }
                 pinFaces.push_back(interface_(grid, botCell, Opm::FaceDir::ZMinus));
                 pinCells.push_back(botCell);
--- a/opm/core/grid/PinchProcessor.hpp
+++ b/opm/core/grid/PinchProcessor.hpp
@@ -389,37 +389,27 @@ namespace Opm
                 /// if the original segment's top and bottom is inactive, we need to lookup
                 /// the column until they're found otherwise just ignore this segment.
                 if (!actnum[topCell]) {
-                    seg.insert(seg.begin(), topCell);
                     for (int topk = ijk1[2]-2; topk > 0; --topk) {
                         topCell = getGlobalIndex_(ijk1[0], ijk1[1], topk, dims);
                         if (actnum[topCell]) {
                             break;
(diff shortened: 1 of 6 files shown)


atgeirr:
> Note this is not compatible with #907

I intend to merge #907 soon, will you change this, then?

totto82:
This PR is changed to keep the same behavior as is as long the new minpv-pinch-nnc implementation is not fully tested. To do this opmfil is hardcoded to true.

bska:
> This PR is changed to keep the same behavior as is as long the new minpv-pinch-nnc implementation is not fully tested.

I don't understand the meaning of that sentence.

totto82:
The original intention of this PR was to start using the pinchProcessor to generate connections and transmissibilities over cells removed by the minpvProcessor. Unfortunately, the norne results degrades, for an unknown reason, when such an approach is used. The latest changes to this PR hardcodes the old (and better) behavior, while keeping the changes that ads value to the code base. i.e.  handing of threshold pressure in combination with the nncs and the option of specifying whether cells with pore volumes less then MINPV should be merged to its neighbor or not.

opm/core/grid/PinchProcessor.hpp
```diff
@@ -389,38 +389,30 @@ namespace Opm
...
-                        } else {
-                            auto it = seg.begin();
-                            seg.insert(it, topCell);
                         }
                     }
-                    pinFaces.push_back(interface_(grid, topCell, Opm::FaceDir::ZPlus));
-                } else {
-                    pinFaces.push_back(interface_(grid, topCell, seg.front()));
                 }
+                pinFaces.push_back(interface_(grid, topCell, Opm::FaceDir::ZMinus));
```
atgeirr: I'd appreciate an explanation of why this is changed, especially turning ZPlus into ZMinus.

opm/core/utility/thresholdPressures.hpp
```diff
@@ -355,6 +356,55 @@ void computeMaxDp(std::map<std::pair<int, int>, double>& maxDp,
...
+
+                if (thresholdPressure->hasRegionBarrier(eq1,eq2)) {
+                    if (thresholdPressure->hasThresholdPressure(eq1,eq2)) {
+                        thpres_vals[i] = thresholdPressure->getThresholdPressure(eq1,eq2);
+                    } else {
+                        // set the threshold pressure for NNC of PVT regions where the third item
+                        // has been defaulted to the maximum pressure potential difference between
+                        // these regions
+                        const auto barrierId = std::make_pair(eq1, eq2);
+                        thpres_vals[i] = maxDp.at(barrierId);
```
atgeirr: This is fine as is (and just as in the function above), but it can be written simply as `maxDp.at({eq1, eq2})`.

atgeirr:
Thanks for the patch! This is almost ready to go in, I'll merge once I am satisfied that the changes in the PinchProcessor are correct.

opm/core/grid/PinchProcessor.hpp
totto82: ZPlus and ZMinus should never have been changed. From the topCell point of view it is the ZPlus face.  
According to my understanding of this code. The seg.front() cell is the uppermost cell in the minpv column and thus therefore not exist in the grid.  ``interface_(grid, topCell, seg.front())`` tries to find the common face between topCell and seg.front(). Maybe @qilicun could comment on the original thought behind this.

qilicun: The original thought of this is: if the topCell  is not active, then extend the segment until finding an active cell in the upper, which is seg.front(), then get the lower face of the seg.front() which is the new top cell. Else if the topCell is active, then return the intersection face of topCell and seg.front(). here seg,front stays the same which in the non-extended segment.

qilicun: So, if you don't want to extend the segment, I think you can just use ``pinFaces.push_back(interface_(grid, topCell, seg.front())``

totto82: My concern with ``` interface_(grid, topCell, seg.front() ``` is what getActiveCellIdx_(grid, seg.front()) returns (in side interface(grid, cellIdx1, cellIdx2) in the case where seg_front() does not exist in the unstructured grid.

qilicun: In this case, if TopCell is active, all the cell in segments is active.

atgeirr:
With this, and the updated PR in opm-autodiff, I'll merge the whole chain. Thanks for the effort!

PR #929, opened 2015-11-20
Relperm diagnostics

qilicun:
This PR is just the start of relperm diagnostics. Currently, the diagnostics support features including
- saturation table check
- endpoints consistency check

In the near feature, more features like, endpoint scaling method check and hysteresis check will be added. 

This diagnostics is very far beyond shaped, any comments and suggestions are welcome!


examples/diagnose_relperm.cpp
```diff
@@ -0,0 +1,113 @@
...
+int
+main(int argc, char** argv)
+try
+{
+    using namespace Opm;
+    if (argc <= 1) {
+        usage();
+        exit(1);
+    } 
+    static char* ECLIPSEFILENAME(argv[1]);
```
atgeirr: It is best to use ALL_CAPS for macros only, and I see no reason to use a static variable, for example:
```
const char* eclipseFileName = argv[1];
```

examples/diagnose_relperm.cpp
```diff
@@ -0,0 +1,113 @@
...
+    std::ifstream eclipseFile(ECLIPSEFILENAME, std::ios::in);
+    if (eclipseFile.fail()) {
+        std::cerr << "Error: Filename " << ECLIPSEFILENAME << " not found or not readable." << std::endl;
+        usage();
+        exit(1);
+    }
+    eclipseFile.close();
```
atgeirr: I think the checking above is unnecessary, since it is done (or should be) anyway by parseFile().

examples/diagnose_relperm.cpp
```diff
@@ -0,0 +1,113 @@
...
+    // Read saturation tables.
+    EclipseStateConstPtr eclState; 
+    ParserPtr parser(new Opm::Parser);
+    //ParseMode parseMode;
+    Opm::ParseMode parseMode({{ ParseMode::PARSE_RANDOM_SLASH , InputError::IGNORE }, 
+                              { ParseMode::PARSE_UNKNOWN_KEYWORD, InputError::IGNORE},
+                              { ParseMode::PARSE_RANDOM_TEXT, InputError::IGNORE}
+                             });
+    Opm::DeckConstPtr deck(parser->parseFile(ECLIPSEFILENAME, parseMode));
+    //std::string deck_filename = param.get<std::string>("deck_filename");
```
atgeirr: I assume the intention of the commented-out code was to support running this program with the same command line as you would use for Flow, since that is not supported it is better to remove this entirely.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+
+#include <opm/core/grid.h>
+#include <opm/core/grid/GridManager.hpp>
+#include <opm/core/grid/GridHelpers.hpp>
+#include <opm/core/utility/linearInterpolation.hpp>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/material/fluidmatrixinteractions/EclEpsScalingPoints.hpp>
+
+namespace Opm {
```
atgeirr: The names of these enums may conflict with names elsewhere, I suggest they are moved to the private section of the class since they are not needed in the public interface.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+        NoFamily
+    };
+
+
+    ///This class is intend to be a relpmer diganostics, to detect
+    ///wrong input of relperm table and endpoints.
+    class RelpermDiagnostics 
+    {
+    public:
+        RelpermDiagnostics();
```
atgeirr: I suggest removing the default constructor, as it is unused.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+        RelpermDiagnostics(EclipseStateConstPtr eclstate);
```
atgeirr: This constructor should be `explicit` to avoid unintended implicit conversion.

Also, the constructor and `diagnosis()` must be documented using doxygen notation.

atgeirr: On closer look, I see that none of the constructors do anything, in particular the argument of this constructor is not used. I therefore suggest that you instead delete them both (the compiler-generated default constructor would be the same as the current implemented one).

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+    RelpermDiagnostics::RelpermDiagnostics(Opm::EclipseStateConstPtr eclState)
+    {}
+
+
+    void RelpermDiagnostics::diagnosis(Opm::EclipseStateConstPtr eclState,
+                                       Opm::DeckConstPtr deck,
+                                       const UnstructuredGrid& grid)
+    {
+        std::cout << "***************Relperm Diagnostics***************\n";
+        phaseCheck_(eclState, deck);
```
atgeirr: Since these function have a logical ordering, I suggest that you put both the in-class declarations and the out-of-class definitions in the same order as they are called here.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+        void sgofTableCheck_(const Opm::SgofTable& sgofTables);
+        void slgofTableCheck_(const Opm::SlgofTable& slgofTables);
+        void swfnTableCheck_(const Opm::SwfnTable& swfnTables);
+        void sgfnTableCheck_(const Opm::SgfnTable& sgfnTables);
+        void sof3TableCheck_(const Opm::Sof3Table& sof3Tables);
+        void sof2TableCheck_(const Opm::Sof2Table& sof2Tables);
+        void sgwfnTableCheck_(const Opm::SgwfnTable& sgwfnTables);
+    };
+
+    RelpermDiagnostics::RelpermDiagnostics()
```
atgeirr: In general, I find it much easier to read a file like this if the functions are separated by more whitespace (I like 5 lines).

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+    {}
+
+    RelpermDiagnostics::RelpermDiagnostics(Opm::EclipseStateConstPtr eclState)
+    {}
+
+
+    void RelpermDiagnostics::diagnosis(Opm::EclipseStateConstPtr eclState,
```
atgeirr: Since the class is not a template, nor any of the methods, I suggest putting all the implementation into a .cpp file.

As written, there may be linker issues: when defining (implementing) a function in a header, it must be either:
 - implemented inline in the class,
 - a template, or
 - declared `inline` explicitly.
Either of those will give the function weak linkage, meaning that the linker will handle the fact that it will be duplicated in all compilation units the header is included from. Failing to do so may cause linker errors. In this case I prefer not having the methods implemented in the header.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+
+    private:
+        FluidSystem fluidSystem_;
+        
+        SaturationFunctionFamily satFamily_;
+
+        std::vector<Opm::EclEpsScalingPointsInfo<double> > unscaledEpsInfo_;
+        std::vector<Opm::EclEpsScalingPointsInfo<double> > scaledEpsInfo_;
+
+        std::vector<std::string> messager_;
```
atgeirr: I think `messages_` would be a better name.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+        }
+    }
+
+    void RelpermDiagnostics::swofTableCheck_(const Opm::SwofTable& swofTables)
+    {
+        const auto& sw = swofTables.getSwColumn();
+        const auto& krw = swofTables.getKrwColumn();
+        const auto& krow = swofTables.getKrowColumn();
+        ///Check sw column.
+        if (sw.front()< 0 || sw.back() > 1) {
```
atgeirr: Space around `<` (yes I know this is only a nitpick...)

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+
+    void RelpermDiagnostics::slgofTableCheck_(const Opm::SlgofTable& slgofTables) 
+    {
+        const auto& sl = slgofTables.getSlColumn();
+        const auto& krg = slgofTables.getKrgColumn();
+        const auto& krog = slgofTables.getKrogColumn();
+
+        ///Check sl column.
+        ///TODO first value means sl = swco + sor
+        if (sl.front()< 0 || sl.back() > 1) {
```
atgeirr: Whitespace around `<` again. Also: in general when dealing with doubles I find it clearer if the code uses 0.0 and 1.0 for the literals instead if 0 and 1. While not incorrect, it makes it clear that we are dealing with floating-point numbers.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+
+
+
+    void RelpermDiagnostics::unscaledEndPointsCheck_(DeckConstPtr deck,
+                                                     EclipseStateConstPtr eclState)
+    {
+        // get the number of saturation regions and the number of cells in the deck
+        unsigned numSatRegions = static_cast<unsigned>(deck->getKeyword("TABDIMS")->getRecord(0)->getItem("NTSFUN")->getInt(0));
```
atgeirr: Since the method call return an int, why not use an int for numSatRegions (as well as for the loop counter satnumIdx below)?

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+            std::cout << "relperm: Saturation Family I." << std::endl;
+        } 
+        if (!family1 && family2) {
+            satFamily_ = SaturationFunctionFamily::FamilyII;
+            std::cout << "relperm: Saturation Family II." << std::endl;
+        }
+    }
+
+
+    void RelpermDiagnostics::phaseCheck_(EclipseStateConstPtr eclState,
```
atgeirr: The eclstate is not used by this method, please remove.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,643 @@
...
+                         auto sl = slgofTables.getTable<SlgofTable>(satnumIdx).getSlColumn();
+                         auto krog = slgofTables.getTable<SlgofTable>(satnumIdx).getKrogColumn();
+                         krog_value=Opm::linearInterpolation(sl, krog, unscaledEpsInfo_[satnumIdx].Sgl);                        
+                     }
+                     auto sw = swofTables.getTable<SwofTable>(satnumIdx).getSwColumn();
+                     auto krow = swofTables.getTable<SwofTable>(satnumIdx).getKrowColumn();
+                     krow_value = Opm::linearInterpolation(sw, krow,unscaledEpsInfo_[satnumIdx].Swl);
+                 }
+                 if (satFamily_ == SaturationFunctionFamily::FamilyII) {
+                     assert(!sof3Table.empty());
```
atgeirr: This fails to compile. Should be `sof3Tables` perhaps?

atgeirr:
Thanks for the contribution! I have made comments on the code in the diff, that should be fixed.

Note that I have *not* carefully reviewed the behavior of the program and the actual checks, and since the PR contain no test program it is hard for me to know if it does what it is supposed to do. It should not be too hard to write tests for this though, although you might want to change the interface a little (for example, to make an accessor for the messages so that you could check after the call to diagnose() if any problem was encountered).

I have been convinced (yes @joakim-hove...) that testing must be in from the start or it will not happen. I'd like to see at least a few tests before it is merged.

qilicun:
@atgeirr 
Thanks for your comments. I have fixed lots of them. About the unit tests, I'm still thinking, it seems I need to make some changes to the ``messages_`` not only output messages.

atgeirr:
Thanks for your update, the class looks very good now. Please post a notice when you have fixed the remaining minor issues in the program and added some tests, so I can merge it.

qilicun:
@atgeirr 
A simple tests have been made, and also OpmLog is used to store error messages, that might the first time OpmLog is really used.:)

examples/diagnose_relperm.cpp
```diff
@@ -0,0 +1,102 @@
...
+main(int argc, char** argv)
+try
+{
+    using namespace Opm;
+    if (argc <= 1) {
+        usage();
+        exit(1);
+    } 
+    const char* eclipseFilename = argv[1];
+    std::ifstream eclipseFile(eclipseFilename, std::ios::in);
```
atgeirr: I do not see the need to create this object (and close the file immediately), this should be left for the parseFile() call.

atgeirr:
I am not sure why this is no longer mergeable, probably a conflict in the CMake files.

qilicun:
@atgeirr 
The conflict is solved, the green button comes back.

opm/core/props/satfunc/RelpermDiagnostics.hpp
```diff
@@ -0,0 +1,123 @@
...
+namespace Opm {
+
+    ///This class is intend to be a relpmer diganostics, to detect
+    ///wrong input of relperm table and endpoints.
+    class RelpermDiagnostics 
+    {
+    public:
+
+        ///Constructor for OpmLog.
+        RelpermDiagnostics(std::string& logFile);
```
atgeirr: I was about to merge, then I noticed that the changed constructor is not `explicit`, it should be since otherwise we allow implicit conversion of a string to a RelpermDiagnostics object, which does not make sense.

Sorry for noticing so late!

qilicun: Sorry, I don't know this, you mean adding ``explicit`` to this constructor?

atgeirr: Yes:
```
explicit RelpermDiagnostics(std::string& logFile);
```
Without, the following compiles:
```
void foo(const RelpermDiagnostics& rd);
...
std::string s("Some string");
foo(s);
```

qilicun:
@atgeirr 
Thanks, the PR is updated.

atgeirr:
Great! Thanks for your patience and persistence, merging now!

PR #930, opened 2015-11-24
Use cell depths based on averaging cell corners

totto82:
CellCenterDepth is used instead of z coordinate of the cell centroid. The cellCenterDepth is calculated based on the cell corners as Eclipse.

### DIFF
--- a/opm/core/simulator/initStateEquil_impl.hpp
+++ b/opm/core/simulator/initStateEquil_impl.hpp
@@ -313,7 +313,6 @@ namespace Opm
                    const CellRange&                    cells,
                    std::vector<double>&                p    )
             {
-                const int nd = UgGridHelpers::dimensions(G);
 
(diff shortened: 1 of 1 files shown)


atgeirr:
Looks good, the only question I have is: is there a point in retaining the existing behavior as an option?

totto82:
I agree. The question is how. Would it be convenient with a grid.type (= cornerpoint etc. ) telling how to compute the depth? This could also be used to unify the center and volume calculations in the grid with the ones in EclipseGrid

atgeirr:
> Would it be convenient with a grid.type (= cornerpoint etc. ) telling how to compute the depth? This could also be used to unify the center and volume calculations in the grid with the ones in EclipseGrid

That could work. It would be a bit of a hassle passing a parameter like that through the whole system to where it is used though. It may not be worth it if it complicates the code a lot.

PR #931, opened 2015-11-27
fix build for newish boost versions

andlaus:
my boost is 1.58. once more, boost seemed to have become more picky on
where a semicolon must be placed and where not.

### DIFF
--- a/tests/DuneIstlTestHelpers.hpp
+++ b/tests/DuneIstlTestHelpers.hpp
@@ -61,7 +61,7 @@ struct MPIFixture {
 };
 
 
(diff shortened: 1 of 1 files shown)


bska:

andlaus:
small correction: the boost version which triggered this was 1.59 (which is compiled in my home directory). A bit downstream I got some linker errors due to conflicts with 1.58 (which is provided by the system). I have no idea if this is an issue of the build system or whether it is just my setup, but I don't think it is worth to spend time on investigating this so I'll revert back to the distributor's version.

PR #932, opened 2015-11-30
Added regression test for Norne PVT.

joakim-hove:
The test is based on dumping the internal state of the PvtLiveOil and
PvtLiveGas objects and comparing with a reference state stored on disk.

A significant refactoring of the table code in opm-parser is underway - this is just preparing the crash helmet.


tests/test_norne_pvt.cpp
```diff
@@ -0,0 +1,92 @@
...
+bool compareContent(const std::stringstream& newContent , const std::string& referenceFile) {
+    std::ifstream inputStream(referenceFile);
+    std::stringstream oldContent;
+    oldContent << inputStream.rdbuf();
+    if (oldContent.str() == newContent.str()) {
+        return true;
+    } else {
+        return false;
+    }
+}
```
bska: This pattern is pervasive in opm-parser.  Is there a reason you don't write just
```C++
return oldContent.str() == newContent.str();
```
?

joakim-hove: > This pattern is pervasive in opm-parser. Is there a reason you don't write just

Well - personally I think the form chosen is more readable; however I realize that is not the common conception in this part of the woods - I will push an update tomorrow. 

That said I have felt dirty ever since I created this PR; I really do not like exposing the internals in that way. I'll try to see if I can achieve the same using the public API.

[1]: Funny thing: we all agree that creating readable, easy to maintain code is a goal - however we quite often disagree on exactly what constitutes most readable code.

bska: > we all agree that creating readable, easy to maintain code is a goal - however we quite often disagree on exactly what constitutes most readable code.

Yes, and that's why I asked in the first place.  I find that this particular construction cumbersome because it makes me read the condition twice to see if there's something I'm missing.

andlaus: > [1]: Funny thing: we all agree that creating readable, easy to maintain code is a goal - however we quite often disagree on exactly what constitutes most readable code. 

That's actually true.

What I consider even funnier is that most people (and that definitely includes myself) are inconsistent with themselves on what this means: Depending on the context, the same person usually prefers different things. ;)

joakim-hove:
OK - this has been completely rewritten; now the `PvtLiveOil` instance from norne is tested with the `mu( )` method.

bska:
> OK - this has been completely rewritten

I have not looked into the details of this at all, but I do have one stylistic comment.  Why do the `verify_*()` functions take a `std::shared_ptr<>` parameter (by value) when neither of them engage in lifetime management of the underlying resource?

For instance, function [`verify_norne_oil_pvt_region2()`](https://github.com/OPM/opm-core/pull/932/files#diff-d00b06676ac9fc2d6017b08b9ff2b51eR59) is declared as
```C++
void verify_norne_oil_pvt_region2(std::shared_ptr<const TableManager> tableManager);
```
I think it would be easier to understand if the parameter type was `const TableManager&` instead.

joakim-hove:
>Why do the verify_*() functions take a std::shared_ptr<> parameter (by value) when neither of them engage in lifetime management of the underlying resource?

Because I *love* `std::shared_ptr<>` - I probably overuse it, but quite frankly I think C++ `std::forward`, `&&`, `&` return value optimization(??) and so on is a quagmire. Use of `std::shared_ptr<>` just works&trade;.

joakim-hove:
> I think it would be easier to understand if the parameter type was const TableManager& instead.

I'm all for code readability and I'm happy to change the signature as indicated; however when combined with:

> I have not looked into the details of this at all

I am wondering if we have passed the nitpick limit? As a developer I would like to know if the reviewer intends to review the PR as a whole- Addressing minor stylistic issues like this one without a clear commitment from the reviewer that: 'I will review this PR fully - and eventually merge or close it', makes limited sense.

joakim-hove:
> I'm happy to change the signature as indicated;

done

joakim-hove:
>And for shared_ptr addicts:
http://meetingcpp.com/index.php/br/items/shared_ptr-addiction.html

Thank you - that was an interesting read; to be honest I have also started to see in particular the issue of "interface dictatorship" caused by extensive use of shared pointers in the API:

```html
... it will force you and any other user of this code, to use std or/and boost::shared_ptr...
```

bska:
>> http://meetingcpp.com/index.php/br/items/shared_ptr-addiction.html
>
> Thank you - that was an interesting read;

Interesting indeed.  To me the overarching principle for when to use a smart pointer in an interface, especially as a function parameter, is whether or not the function takes part in the lifetime management of the underlying object.  As a rule of thumb If I just want to look at (read from) an object of class `X` I normally choose a parameter of type `const X&` (or `const X*` if the parameter is somehow optional and `nullptr` is a valid sentinel for "not present").  If I need read-write access to the object I typically choose `X&` or `X*`.  I treat rvalue references (`X&&`) as an optimisation technique and hardly ever use it outside of constructors and I have yet to employ forwarding references at all.

This way I can defer object lifetime management to the caller which is usually trivial through structured scopes.  Only in the few cases where I really want to participate in lifetime management do I use a smart pointer as a function parameter type.

bska:
I've looked at the thing now and run the tests in the following environments
```
{Debug, Release} x { Sequential, MPI } x { Dune 2.3.1, Dune 2.4.0 }
```
All tests pass in all environments so I intend to merge this.  I do have a couple of comments before I merge, however.

1. Strictly speaking you need to `#include <boost/test/floating_point_comparison.hpp>` in order to use the `BOOST_*_CLOSE()` macros.

2. The vector names `*B*` are misleading since you're calling the reciprocal FVF method, `b()`.

3. You're testing Rs values that are within the ranges of each region's PVTO table.  I think it would good to have some kind of idea of how the evaluators behave outside the defined ranges--especially for the case of Rs greater than the maximum table value.

4. Vector's `size()` method does not return a `size_t`--it returns a `std::vector<T>::size_type`--so the loop should be written as either
    ```C++
    for (std::vector<double>::size_type i = 0, n = mu.size(); i < n; ++i) { /* stuff */ }
    ```
    or
    ```C++
    for (decltype(mu.size()) i = 0, n = mu.size(); i < n; ++i) { /* stuff */ }
    ```
    or
    ```C++
    for (auto i = 0*mu.size(), n = mu.size(); i < n; ++i) { /* stuff */ }
    ```
    I don't really like any of these, but I personally prefer the second option.

joakim-hove:
>  I do have a couple of comments before I merge, however.

Good - thank you; I will address your issues 1-3 tomorrow; I must admit I have difficulties with the fourth issue which I maybe feel is a bit over the top?!

bska:
> I must admit I have difficulties with the fourth issue which I maybe feel is a bit over the top?!

I don't necessarily disagree.  On the other hand, if we never start writing this kind of loop correctly we will keep making the same mistakes.  I have seen cases where it actually mattered.

blattms:
On Wed, Dec 02, 2015 at 07:48:09AM -0800, Bård Skaflestad wrote:
>     ```C++
>     for (std::vector<double>::size_type i = 0, n = mu.size(); i < n; ++i) { /* stuff */ }
>     ```
>     or
>     ```C++
>     for (decltype(mu.size()) i = 0, n = mu.size(); i < n; ++i) { /* stuff */ }
>     ```
>     or
>     ```C++
>     for (auto i = 0*mu.size(), n = mu.size(); i < n; ++i) { /* stuff */ }
>     ```
>     I don't really like any of these, but I personally prefer the second option.

For tests if would prefer
```C++
std::equal(mu.begin(), mu.end(), mu_expected.begin(),
           [](double a, double b){BOOST_CHECK_CLOSE( a, b, 1e-5); });
```
Not sure whether this will compile, though. Depends on whether
BOOST_CHECK_CLOSE does return anything.

Or use iterators
```C++
for ( auto mu_it=mu.begin(), mu_end=mu.end(); mu_it != mu_end; ++mu_it) {
      auto i = mu_it - mu.begin()
      BOOST_CHECK_CLOSE( *mu_it , mu0_expected[i] , 1e-5);
      BOOST_CHECK_CLOSE( B0[i] , B0_expected[i] , 1e-5);
}
```

bska:
> Or use iterators
```C++
for ( auto mu_it=mu.begin(), mu_end=mu.end(); mu_it != mu_end; ++mu_it) {
    auto i = mu_it - mu.begin();
    BOOST_CHECK_CLOSE( *mu_it , mu0_expected[i] , 1e-5);
    BOOST_CHECK_CLOSE( B0[i] , B0_expected[i] , 1e-5);
}
```

I would normally use iterators too (or a range-for), but the need to introduce the counter `i` to be able look up the expected results does make this particular case less appealing in my opinion.  This is the downside to the iterator-based solutions.  They work very well when you "only" need to loop over a single range but lead to various ~~kludges~~ workarounds when we need to simultaneously step multiple ranges.

joakim-hove:
> You're testing Rs values that are within the ranges of each region's PVTO table. I think it would good to have some kind of idea of how the evaluators behave outside the defined ranges--especially for the case of Rs greater than the maximum table value.

As I see it the Rs values are like the outermost argument values, and the P values are the 'inner-argument' values; so I would assume the most interesting test was to have P values outside the range of the inner table corresponding to a particular R value? Or am I misunderstanding your request here.

joakim-hove:
Updated addressing issues 1,2 and 4 pushed; awaiting further clarification regarding the out-of-range testing.

bska:
> Updated addressing issues 1,2 and 4 pushed;

Thanks.  Looks good.

> I would assume the most interesting test was to have P values outside the range of the inner table corresponding to a particular R value?

That's interesting too, of course, and will require larger tables.  That said, I did really mean Rs values that exceed the maximum tabulated value on the saturation curve.  In this case, which hopefully is rare in a real simulation, we're supposed to linearly extrapolate the last record--including the undersaturated data--to the higher Rs value and perform regular property evaluation in the extrapolated record.  Maybe using Rs=410 and P={650, 680, 710} for region 1 and Rs=100, P={270, 300, 330} for region 2 could be useful.

joakim-hove:
Observe that negative values for `mu` for the highest `{P, Rs}` values in region 1.

bska:
> Observe that negative values for `mu` for the highest `{P, Rs}` values in region 1.

Thanks, that's ... disappointing.  I sort of feared that we'd see something like this.  It's certainly something to keep in mind for the impending fluid system rewrite.

I'll build and run the new test and merge if nothing else turns up.

bska:
> I'll build and run the new test and merge if nothing else turns up.

No issue.  I'll merge into master.  Thanks a lot for the efforts.

PR #933, opened 2015-12-07
Silence some warnings.

atgeirr:
The thresholdPressures() should probably be changed to not take a deck, but that requires cross-module changes so I'll not do it here.

I consider this trivial enough to self-merge.


atgeirr:
> I consider this trivial enough to self-merge.

A note on that: if you think I overstep what is acceptable to self-merge, please tell me, and we'll eventually develop a good rule-of-thumb.

PR #934, opened 2015-12-09
Add method to access the materialLawManager

totto82:


opm/core/props/satfunc/SaturationPropsFromDeck.hpp
```diff
@@ -129,6 +129,9 @@ namespace Opm
                              const double pcow, 
                              double & swat);
 
+        std::shared_ptr<const MaterialLawManager> materialLawManager() const { return materialLawManager_; }
```
atgeirr: This method should be documented, and rather return a `const MaterialLawManager&`.

totto82: If I understood you correct. You want: 
```
const std::shared_ptr<const MaterialLawManager>& materialLawManager() const { return materialLawManager_; }
```
Then I get "warning: returning reference to temporary [-Wreturn-local-addr]"

The warning disappear when I change it back to how it was before I changed it according to @andlaus comment. 
```
const std::shared_ptr<MaterialLawManager>& materialLawManager() const { return materialLawManager_; }
```

atgeirr: No, what I want is
```
const MaterialLawManager& materialLawManager() const { return *materialLawManager_; }
```

atgeirr:
Thanks, merging.

PR #939, opened 2016-01-05
Table column

joakim-hove:
See: https://github.com/OPM/opm-parser/pull/649

### DIFF
--- a/opm/core/props/pvt/PvtDead.cpp
+++ b/opm/core/props/pvt/PvtDead.cpp
@@ -47,9 +47,9 @@ namespace Opm
             const Opm::PvdoTable& pvdoTable = pvdoTables.getTable<PvdoTable>(regionIdx);
 
             // Copy data
-            const std::vector<double>& press = pvdoTable.getPressureColumn();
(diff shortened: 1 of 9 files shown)


atgeirr:
Looks good, there are a few minor whitespace issues (missing space around `=` etc.) in the pvt classes, but since those are likely to be replaced soon I'll not waste time on it.

Two questions:
 - Has this been verified to run Norne with identical results?
 - Where do I look to find the strings that are the names of the columns? JSON?

joakim-hove:
> Where do I look to find the strings that are the names of the columns? JSON?

That is unfortunately a week spot; those strings are currently hardcoded - and there is no compiler induced type safety. See e.g.: https://github.com/joakim-hove/opm-parser/blob/table-column/opm/parser/eclipse/EclipseState/Tables/ImptvdTable.hpp and https://github.com/joakim-hove/opm-parser/blob/table-column/opm/parser/eclipse/EclipseState/Tables/PvtoTable.hpp

The future goal is clearly to bind closer to JSON configuration, and also to devise a system for compilation type-safety. Those changes should (mostly) be isolated to the parser.

PR #940, opened 2016-01-07
OPM-163: Use EclipseState instead of Deck for PvtInterface

chflo:
Minor refactor - use EclipseState instead of Deck for PvtInterface. 
If PVTNUM is not present, getIntGridProperty() returns a vector with pvtnum 1 for each gridcell/box. 

Due to changes in extractPvtTableIndex() method signature, the PR https://github.com/OPM/opm-autodiff/pull/569 should be merged at the same time.


atgeirr:
Thanks, looks fine, will merge.

PR #942, opened 2016-01-11
fix relperm diagnostics tests issue.

qilicun:
This should fix the relperm diagnostics test issue.


joakim-hove:
Good - works :-)

PR #943, opened 2016-01-18
Update relperm diagnostics

qilicun:
Some changes have been made, main points are:
 - Add saturation region number for output
 - Add cell index for scaled endpoints message output
 - Output error summary for terminal and LOGFILE.

Any feedback arr welcome!


alfbr:
We need to fix some output formatting. Rather than tracking down commits, I will just list the relevant output messages here.

"Number of saturation region" should be plural, that is, "Number of saturation regions".

"should less than" is not correct English, change to "should be less than".

The error summary counts warnings as errors.
 
It seems that quite grave errors in the input deck is given as warnings, but this I suggest we discuss independent of the merging of this PR.

@atgeirr or @bska please consider doing code review and merging.

alfbr:
Forgot one comment, the region count starts with 0, should it not start with 1?

bska:
> @atgeirr or @bska please consider doing code review and merging.

I'll review the PR tonight (CET) if no-one beats me to it.

alfbr:
Great, thanks. Please allow time for fixing the issues I listed before merging.

alfbr:
Sorry for spamming, but I just saw one more that escaped me earlier. In the initial list it reads:
relperm: Saturation Family I
To remain consistent with the other lines, a capital R should be used. Moreover, abbreviations should be avoided. At least it should read:
Relative permeability:
but even better, it should describe exactly what it is, for instance
Relative permeability input format:

bska:
Overall I think this is okay.  I'd prefer if the types were a little tighter with respect to `const` (i.e., anything that **can** be `const` **should** be `const`) and some of the `to_string()` calls could be omitted, but that's not really much of a problem.  This utility is not performance critical in any way.

You also introduce an instance of end-of-line [whitespace](https://github.com/OPM/opm-core/pull/943/files#diff-146c3c10f37c97576f1c92eb8c38cbc7R694).  I'd like to see that be removed before I merge this.

On a slightly unrelated note, the [`compressedToCartesian` array](https://github.com/qilicun/opm-core/blob/update_relperm_diagnostics/opm/core/props/satfunc/RelpermDiagnostics.cpp#L673-L679) should be constructed using the [`Opm::compressedToCartesian()`](https://github.com/OPM/opm-core/blob/master/opm/core/utility/compressedToCartesian.hpp#L28-L36) function of `opm/core/utility/compressedToCartesian.hpp`.  I only mention this because you're in the area and are using the result as part of your new code.

qilicun:
I have made some changes, the comments of dd9b607 should be "as much as possible..."

About warning and error, some of them are very clear in ECLIPSE manual, but some of them are not. This can be a major issue when we output the messages, which should not confuse the user.

alfbr:
Did you fix the summary, are warnings still counted under total number of errors?

bska:
Due to #944 this can no longer be merged automatically.  Please rebase onto the top of current master (there's a small conflict in the include statements).

joakim-hove:
> the region count starts with 0, should it not start with 1?

That is a good question; currently most all integer properties from the Eclipse input deck are immediately shifted with ` -= 1` to facilitate use as an index. When it comes to the coordinates `i, j, k` I guess that makes sense - that tradition is extremely well established. But - when it comes e.g. `SATNUM` values I would prefer that they were treated as ID's and not assumed to be used as indices.

In the parser code there are some values which are left unmodified, but the general rule in opm is to shift everything. Changing this would of course require a big **flag day** - which would actually get my vote; but others might disagree.

atgeirr:
About counting from 0 or 1: if there was a simple rule-of-thumb about when you would count from 1 (i.e. treat numbers as IDs and not indices) that would help. What are some typical items that we do *not* subtract 1 from currently?

joakim-hove:
> What are some typical items that we do not subtract 1 from currently?

The only datastructure I am aware of is the `MULTREGT` handling - that is done fully in the parser, and the convention (i.e. *not* subtract 1) is isolated there: https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Grid/MULTREGTScanner.cpp#L99 

Another field where I would like to stick the input values is for tables, currently the pseudo code for getting e.g. the local relperm table is effectively like:

```
int satnum_value = SATNUM[i,j,k];
auto relperm_table = tableContainer.getTable( satnum_value - 1 );
```

As defined by the reservoir engineer setting up the model the `SATNUM`values are *ID* values, and in particular there is no guarantee that it forms a densely filled set `[1,...N]`.

qilicun:
@alfbr 
``
Did you fix the summary, are warnings still counted under total number of errors?
``

Yes, that's the way eclipse does. At the end of  .PRT file, eclipse output ``Error Summary`` for all the number of errors, warning, bugs, comments , etc. If you think this is not correct, we can choose another name.

qilicun:
@alfbr 

I have made some changes, it's ready to test and review.

bska:
> As defined by the reservoir engineer setting up the model the `SATNUM` values are *ID* values, and in particular there is no guarantee that it forms a densely filled set `[1,...,N]`.

You are of course correct that there's no expectation of a densely filled set but the `SATNUM` array does have to honour a couple of restrictions.  In particular, no entry should exceed `NTSFUN` from `TABDIMS` (item 1, default 1) or be less than 1.  I admit that I can't quite imagine why one would want to use noncontiguous values as this would require raising `NTSFUN` and entering more tabulated saturation functions, but I'll take your word for it.

joakim-hove:
> admit that I can't quite imagine why one would want to use noncontiguous values 

Not initially - but these models often live a loooong time, and noncontiguous SATNUM values could very well be the result of updates and revisions.

bska:
>> admit that I can't quite imagine why one would want to use noncontiguous values
>
> Not initially - but these models often live a loooong time, and noncontiguous SATNUM values could very well be the result of updates and revisions.

Ah, okay.  That makes sense of course.

atgeirr:
> noncontiguous SATNUM values could very well be the result of updates and revisions.

Would one still need to provide SWOF tables (for example) for all possible (contiguous) SATNUM values up to the maximum? If so, then I think it is meaningful to treat it as an index, since there is an object (table) for every index value even if it is not actually used anymore (by occurring in SATNUM).

bska:
>> noncontiguous SATNUM values could very well be the result of updates and revisions.
>
> Would one still need to provide SWOF tables (for example) for all possible (contiguous) SATNUM values up to the maximum?

Yes.  This process is simplified somewhat by the fact that an entire table can be default (consisting of an empty record only), in which case the values will be copied from the table previously entered (the one immediately preceding the current table).

alfbr:
Beautiful, thanks for your efforts @qilicun , @bska please merge at your convenience.

bska:
> @bska please merge at your convenience.

Thanks a lot for the testing effort.  I'll merge this into master.

PR #945, opened 2016-01-25
Removed SimulatorState::init( UnstructureGrid& )

joakim-hove:
When this overload has been removed the `SimulatorState` class has no dependencies to other parts of opm-core, and it can be moved to opm-common.

The overload seemed to be only used in the tutorials.

### DIFF
--- a/opm/core/simulator/SimulatorState.cpp
+++ b/opm/core/simulator/SimulatorState.cpp
@@ -1,6 +1,5 @@
 #include <opm/common/ErrorMacros.hpp>
 #include <opm/core/simulator/SimulatorState.hpp>
-#include <opm/core/grid.h>
 
 #include <cmath>
 #include <cassert>
@@ -41,11 +40,7 @@ SimulatorState::vectorApproxEqual(const std::vector<double>& v1,
     return true;
 }
 
-void
-SimulatorState::init(const UnstructuredGrid& g, int num_phases)
-{
-    init(g.number_of_cells, g.number_of_faces, num_phases);
-}
+
 void
 SimulatorState::init(int number_of_cells, int number_of_faces, int num_phases)
 {
--- a/opm/core/simulator/SimulatorState.hpp
+++ b/opm/core/simulator/SimulatorState.hpp
@@ -8,16 +8,12 @@
 #include <vector>
 #include <string>
 
-// forward declaration
-struct UnstructuredGrid;
 
 namespace Opm
 {
     class SimulatorState
     {
     public:
-
-        virtual void init(const UnstructuredGrid& g, int num_phases);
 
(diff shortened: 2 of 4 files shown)


bska:
I'm a little conflicted about this, to be honest.  On the one hand, it's certainly laudable to move the `SimulatorState` to -common.  On the other, being able to initialise a solution structure that's supposed to be tied to a particular grid by passing the grid to the initialisation method is very convenient.  I think I'd prefer if that ability was not removed outright, but rather replaced by a simple non-member wrapper around the other `init()` method--for instance something along the lines of
```C++
SimulatorState createSimulatorState(const UnstructuredGrid& G, const int nphase)
{
    auto x = SimulatorState{};
    x.init(G.number_of_cells, G.number_of_faces, nphase);

    return x;
}
```

atgeirr:
I don't mind much removing this little-used overload. Having a wrapper like @bska suggests is a nice bonus.

However I think we should start thinking about what should go into opm-common. I do not think we should move the SimulatorState as is, but rather the underlying generic functionality (but not specific members such as `pressure()` for example). Also, that generic functionality should be changed so that it  can easily support the WellState classes, too. How should we do that? By having "cell", "face", "well" and "well-perforation" properties? That may be a bit too heavy-handed. An alternative might be to let the clients of the class define their own categories, but some may think that would be a bit too abstract.

joakim-hove:
> Having a wrapper like @bska suggests is a nice bonus.

I can certainly create the mentioned wrapper;  but that will be orphaned when the `SimulatorState` is moved to opm-common - so frankly I would prefer not.

> [...] grid by passing the grid to the initialisation method is very convenient

I have come to the conclusion that we should not design code for *convenient* or *bonus* - we should implement what we need *here and now*.

andlaus:
I agree with @atgeirr: Besides removing these too specifically named methods, we should IMO also consider calling this class something like `FieldContainer` instead of  `SimulatorState` after the move. (because that's what it realistically is, that the simulator can consider it as its `state` is not too relevant in that context IMO.)

joakim-hove:
> However I think we should start thinking about what should go into opm-common. I do not think we should move the SimulatorState as is, but rather the underlying generic functionality (but not specific members such as pressure() for example)

I would prefer:

1. Merge this.
2. Move current implementation to opm-common.
3. Refactor in opm-common.

???

atgeirr:
> I would prefer:
> 
> Merge this.
> Move current implementation to opm-common.
> Refactor in opm-common.

My suggestion:
1. Merge this.
2. Create `FieldContainer` in opm-common.
3. Refactor SimulatorState family of classes to use `FieldContainer`.

joakim-hove:
OK - I will start with:

> 1. Merge this

since we at least agree there!

PR #947, opened 2016-01-28
PvtLiveGas: do not special case in the extrapolation code

andlaus:
at least, don't special case that much: the  condition seems fishy to me because it only applies for the fist PVT region (i.e., `is == 0`), and the generic code below the if works just fine as well.

This caused a discrepancy with the opm-material PVT relations for Norne after 23 report steps. (which takes an hour or two to get to with debugging options on my machine.) As you can probably imagine, finding this was *a lot* of fun...

### DIFF
--- a/opm/core/props/pvt/PvtLiveGas.cpp
+++ b/opm/core/props/pvt/PvtLiveGas.cpp
@@ -574,12 +574,6 @@ namespace Opm
                                                  press);
             } else {  // Undersaturated case
                 int is = tableIndex(saturatedGasTable[0], press);
-                // Extrapolate from first table section
-                if (is == 0 && press < saturatedGasTable[0][0]) {
-                    return linearInterpolation(undersatGasTables[0][0],
-                                                     undersatGasTables[0][item],
-                                                     r);
-                }
 
                 // Extrapolate from last table section
                 //int ltp = saturatedGasTable[0].size() - 1;


joakim-hove:
Not clear to me: Is it a bug in the current opm-core code?

andlaus:
I guess it is, but I'm scratching my head a bit about why this was added in the first place: The PVT class for live oil does not seem to exhibit the analogous code. Anyway, this is what delayed OPM/opm-autodiff#576 for more than a week.

atgeirr:
Since we have merged the opm-material pvt things now, I guess this should also be merged to have the same behaviour? Unless the old pvt classes are about to die immediately, which I do not think is the case (yet)?

andlaus:
Right now I'm working on replacing the old PVT classes by the new ones. (currently my main problem is that it's nowhere explained what the 'z' parameter is, but I'll figure this out.) That said, I'd recommend to merge this PR anyway in order for the new and old implementations be equivalent (at least as long as no additional features are added to the new one).

atgeirr:
Fine, I will merge!

The z should be interpreted as "surface volume density" in most places I think: z = R * inv(B) * s, where (matlab-like notation)
```
R = [ 1, 0, 0;   0, 1, Rv;   0, Rs, 1]
B = [ Bw, 0, 0;   0, Bo, 0;   0, 0, Bg]
```
(I may have mixed up Rs and Rv above).

andlaus:
> The z should be interpreted as "surface volume density" in most places I think

ok. the current opm-core EQUIL code (that's the only part which is relevant to flow as far as I can see=  seems to use it more like a flag for whether the fluid saturated or not (it seems like it only uses 0, 1, 1e100 as values for z)

atgeirr:
> use it more like a flag for whether the fluid saturated or not

That should be thought of as a hack. It works but is not really nice since it can be confusing. Looking forward to eventually getting rid of it.

andlaus:
> That should be thought of as a hack. It works but is not really nice since it can be confusing. Looking forward to eventually getting rid of it.

Ok, I thought that this was a kludge. Be aware that I'm probably not going to the remove this with the upcoming PVT PR, though. (because the hack is in the EQUIL code which for now I'd like not to touch.)

bska:
The equilibration code does not really use surface volumes at all.  The `z` quantity in this case refers to depth.

andlaus:
> The equilibration code does not really use surface volumes at all. The z quantity in this case refers to depth.

hu? but I hope that it passes something more meaningful as the `z` parameter of the PVT relations?

bska:
>> The equilibration code does not really use surface volumes at all. The `z` quantity in this case refers to depth. 
> 
> hu

The equilibration code needs to compute the pressure distribution as a function of depth.  That's what it is set up to do.  I admit that it may be confusing for the symbol `z` to double as depth, but that's what it does mean in this case.

> I hope that it passes something more meaningful as the z parameter of the PVT relations?

It does.  It derives the requisite surface-volume ratios from direct inspection of the `Rs` and `Rv` functions tabulated as functions of depth derived from the `EQUIL` keyword data.  See, for instance, the [Oil Density](https://github.com/bska/opm-core/blob/master/opm/core/simulator/initStateEquil_impl.hpp#L186-L196) calculation in [`initStateEquil_impl.hpp`](https://github.com/bska/opm-core/blob/master/opm/core/simulator/initStateEquil_impl.hpp)

andlaus:
sounds to me like we're talking about two completely orthogonal things: I was wondering about the parameter called `z` which is passed to some of the methods of the PVT classes. I'm pretty sure that this is not depth ;) (and the equilibration code seems like it only passes 0, 1 and 1e100 for it...) @atgeirr's answer was what I needed, btw: Rs = z_g/z_o and R_v = z_o/z_g

bska:
> sounds to me like we're talking about two completely orthogonal things

Apparently yes.  You're not talking about the equilibration code at all, but rather the bit that attempts to do equilibration based on guesswork.

andlaus:
> Apparently yes. You're not talking about the equilibration code at all, but rather the bit that attempts to do equilibration based on guesswork.

at least we agree on what we're not agreeing: actually I'm not talking on the equilibration code at all, but just about the PVT code. (which just happens to be used by the equilibration code.) I won't touch the equilibration code in my next PR. promised ;)

PR #948, opened 2016-02-03
Simulatorstate remove unused

joakim-hove:
As preparation for move/refactor this PR removes unused methods from SimulatorState.


bska:
> Removed unused `SimulatorState::setFirstSat()`

Thank You!

joakim-hove:
I'll merge this later today?

bska:
> I'll merge this later today?

Go ahead.

bska:
>> I'll merge this later today?
>  
> Go ahead.

I guess we jumped the gun a little bit on this as it breaks the build of opm-core.  In particular, the Jenkins CI system now fails when compiling [`BlackoilState.cpp`](https://github.com/OPM/opm-core/blob/master/opm/core/simulator/BlackoilState.cpp).  This file references both the `ExtremalSat` enumeration and the `setFirstSat()` method.

Do you have any local changes that were not included in this PR?

joakim-hove:
> I guess we jumped the gun a little bit on this as it breaks the build of opm-core.

Yes; I don't understand what happened here -  I can only make the lame excuse: "*it works for me*" - investigating now.

PR #950, opened 2016-02-06
Removed opm_memcmp_double() - use cmp from common.

joakim-hove:
Using the the double compare code from: https://github.com/OPM/opm-common/pull/74 

The current PR is a - not really required - step towards replacing the `SimulatorState`class with `SimulationDataContainer` from opm-common. As I see it the main missing functionality now is the `SimulatorState::setFirstSat()` method. A comment says that the method should no longer be used; it is used in the sense that it is required to build opm-core, but after a cursory glance it is not clear to me if it is really needed?


bska:
I like it.  I'm nevertheless surprised you don't need to guard the implementation in the `.cpp` file by
```C++
extern "C" {
    // stuff
}
```

joakim-hove:
> I'm nevertheless surprised you don't need to guard the implementation in the .cpp file [...]

Well - to be honest I was a bit surprised myself ;-). I guess the language layer explanation would be something like:

1. The header file `grid.h` has an `extern "C" { ... }` guard meaning that the symbol `grid_equal( )` will have unmangeled C linkage.

2. The implemention file `grid_equal.cpp` is compiled with a C++ compiler; and therefor has access to C++ features like namespaces - which was the sole reason for these hoops. But - even though a C++ compiler is used - the symbol `grid_equal()` will be unmangeld.

Iff this guess is close to correct that implies that a `grid_equal( )` overload would give rise to both(?) compiler and linker errors - would be interesting to try out.

joakim-hove:
Forcing  Travis rebuild.

joakim-hove:
I'll merge this tomorrow.

bska:
> I'll merge this tomorrow.

Go ahead.  I am a little concerned about the Travis diagnostic
```
/home/travis/build/OPM/opm-core/opm/core/grid/grid.c: In function ‘input_error’:
/home/travis/build/OPM/opm-core/opm/core/grid/grid.c:167:9: warning: implicit declaration of function ‘strerror’ [-Wimplicit-function-declaration]
         fprintf(stderr, "%s: %s\n", err, strerror(save_errno));
         ^
/home/travis/build/OPM/opm-core/opm/core/grid/grid.c:167:9: warning: format ‘%s’ expects argument of type ‘char *’, but argument 4 has type ‘int’ [-Wformat=]
```
We definitely
```C
#include <string.h>
```
i.e., [the requisite header for `strerror()`](http://en.cppreference.com/w/c/string/byte/strerror) in [opm/core/grid/grid.c](https://github.com/OPM/opm-core/blob/master/opm/core/grid/grid.c#L28) so is this some issue on the Travis server?

joakim-hove:
>We definitely

```C++
#include <string.h>
```

ehhh - no; I had mistakenly removed that line from the patch. Back in now.

bska:
> I had mistakenly removed that line from the patch. Back in now.

Ah, thank you for the explanation.  Look forward to centralising this feature to opm-common.

PR #953, opened 2016-02-15
fix the build

andlaus:
it broke because EclipseState::get*GridProperty() now returns a
pointer to a constant object.

this patch us thus required as a follow-up of OPM/opm-parser#687

### DIFF
--- a/opm/core/utility/CompressedPropertyAccess.hpp
+++ b/opm/core/utility/CompressedPropertyAccess.hpp
@@ -205,13 +205,13 @@ namespace Opm {
                      * \return Data values for property \c kw.
                      */
                     template <class PropertyContainer>
-                    static std::shared_ptr< GridProperty<double> >
+                    static std::shared_ptr<const GridProperty<double> >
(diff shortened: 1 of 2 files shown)


atgeirr:
I do not see that the build is broken, but I agree that it would be a required follow-up to the opm-parser PR. It looks clean to me, so please @joakim-hove merge this when/if you merge OPM/opm-parser#687.

joakim-hove:
> I do not see that the build is broken,

I think the build *will* break when OPM/opm-parser#687 is merged. I will merge both coherently.

atgeirr:
> I think the build will break

OK, I was just confused by the PR title.

andlaus:
> OK, I was just confused by the PR title.

yeah, it was quite misleading, sorry. as far as I can see, this PR can be savely merged before OPM/opm-parser#687, though? (the advantage would be that OPM/opm-parser#687 would become stand-alone.)

atgeirr:
I agree it seems to be perfectly mergable standalone, will merge now.

PR #954, opened 2016-02-17
Changes in SimulatorState:

joakim-hove:
Have removed the `setFirstSat()` method, which had a dependency on `template <class Props>` and instead moved the property lookup to the `initState_impl.hpp` file. The initialisation of saturations is now based on a genereric `SimulatorState::setCellDataComponent()` function.

This is not necessarily a perfect API - but with this in place it should be possible to use the `SimulationDataContainer` from opm-common as drop in replacement for `SimulatorState`.

Currently this PR and: https://github.com/OPM/opm-common/pull/76 can be merged independently.

### DIFF
--- a/opm/core/simulator/SimulatorState.cpp
+++ b/opm/core/simulator/SimulatorState.cpp
@@ -2,6 +2,7 @@
 #include <opm/common/util/numeric/cmp.hpp>
 #include <opm/core/simulator/SimulatorState.hpp>
 
+#include <algorithm>
 #include <cmath>
 #include <cassert>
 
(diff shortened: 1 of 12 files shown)


joakim-hove:
Tickle travis.

atgeirr:
I like this direction (generic setCellDataComponent() instead of setFirstSat()).

I'll try to review properly later.

joakim-hove:
> I like this direction 

I'm glad to hear; observe that there is a fallout in opm-autodiff waiting - it is on my list.

opm/core/simulator/SimulatorState.cpp
```diff
@@ -97,42 +98,35 @@ SimulatorState::registerFaceData( const std::string& name, const int components,
...
-    std::vector<double> smin(num_phases_*n);
-    std::vector<double> smax(num_phases_*n);
-    props.satRange(n, &cells[0], &smin[0], &smax[0]);
-    std::vector< double >& sat = saturation();
-    const double* svals = (es == MinSat) ? &smin[0] : &smax[0];
-    for (int ci = 0; ci < n; ++ci) {
-        const int cell = cells[ci];
-        sat[num_phases_*cell] = svals[num_phases_*ci];
-        sat[num_phases_*cell + 1] = 1.0 - sat[num_phases_*cell];
+void SimulatorState::setCellDataComponent( const std::string& name , size_t component , const std::vector<int>& cells , const std::vector<double>& values) {
```
atgeirr: To see if I get this: this function is meant to do the same as the opm-common implementation, so that we can easily switch later?

I have the same minor issues with this method as its clone in opm-common, however it is less important to fix them here, since this implementation will die soon.

tutorials/tutorial3.cpp
```diff
@@ -267,7 +267,19 @@ try
     /// \internal [two-phase state]
     TwophaseState state;
     state.init(grid.number_of_cells , grid.number_of_faces, 2);
-    state.setFirstSat(allcells, props, TwophaseState::MinSat);
+    {
```
atgeirr: Would it be better to make the initSaturation() function available and use it, rather than (almost) duplicating it here?

atgeirr:
Overall looks good, the initSaturation() function encapsulates better the logic that used to be spread around a bit, I think perhaps it should be promoted to a public function (with documentation...)

tutorials/tutorial3.cpp
joakim-hove: > Would it be better to make the initSaturation() function available and use it, rather than (almost) duplicating it here?

Yes - I guess that would be better; but then that function should live on. Please advise on where you would it implemented in that case.

opm/core/simulator/SimulatorState.cpp
joakim-hove: > [...] this function is meant to do the same as the opm-common implementation, so that we can easily switch later?

Yes - that is correct. I'll leave the small stylistic issues - as this should die quite soon anyway.

joakim-hove:
>  ... (with documentation...)

I have taken that pill now :smile:

tutorials/tutorial3.cpp
atgeirr: > Please advise on where you would wist it implemented in that case.

It could stay where it is, just adding the declaration and doc to initState.hpp should be sufficient.

joakim-hove:
OK - this is now ready to be merged, along with: https://github.com/OPM/opm-autodiff/pull/599. Observe that there have been some updates since the initial review by @atgeirr.

bska:
Is there a reason there's no Travis build for this?

jokva:
@bska could be because it's prefix'd by WIP:

bska:
> could be because it's prefix'd by WIP:

Certainly possible, but the audit trail does hint at Travis being available earlier.

joakim-hove:
> Certainly possible, but the audit trail does hint at Travis being available earlier.

yes - I also wondered of this; the PR might have been created without the 'WIP' initially. Anyway - have removed the 'WIP' now; and doing the close / open dance.

bska:
Okay, save for details of implementation this still has the same semantic problem as the `setFirstSat()` it replaces.  Specifically, it has no meaning in the case of a three-phase simulation because there is then no guarantee that the saturations sum to one.

joakim-hove:
>  this still has the same semantic problem as the setFirstSat() it replaces

Absolutely - the ambition for this PR was to pave the way to make the `SimulationDataContainer` from opm-common a drop-in replacement for `SimulatorState`. When(?) this is merged my plan is to submit new PR's where the `s/SimulatorState/SimulationDataContainer/` has been performed, and *then* we can actually start improving the state implementation.

bska:
> When(?) this is merged my plan is to submit new PR's where the `s/SimulatorState/SimulationDataContainer/` has been performed, and then we can actually start improving the state implementation.

Okay, that's a good plan.  Sorry for being dense earlier.  I'll merge into master.

PR #955, opened 2016-02-17
replace the blackoil PVT classes by the ones of opm-material

andlaus:
the opm-material classes are the ones which are now used by
opm-autodiff and this patch makes it much easier to keep the opm-core
and opm-autodiff results consistent. Also, the opm-material classes
seem to be a bit faster than the opm-core ones (see
https://github.com/OPM/opm-autodiff/pull/576)

I ran the usual array of tests with `flow`: SPE1, SPE3, SPE9 and Norne
all produce the same results (for Norne I could spot some tiny differences
in the curves, but certainly nothing which cannot be explained by the numerical
 noise) at the identical runtime (modulo noise) and also "Model 2" seems to work.

note that this patch is quite huge, but most of it is the removal of the old PVT classes (which results in a net removal of almost 3900 lines of code) and a re-generation of the lists in `CMakeLists_files.txt`. without these two changes, only 750 lines have been modified.

also note that this patch requires minor follow-up  changes in opm-autodiff.


andlaus:
hm, the travis error seems to be due to opm-parser:

```
/home/travis/build/OPM/opm-core/opm/core/io/eclipse/EclipseWriter.cpp:41:58: fatal error: opm/parser/eclipse/Utility/SpecgridWrapper.hpp: No such file or directory
```

joakim-hove:
> hm, the travis error seems to be due to opm-parser:

yes - that is true; fix in progress.

joakim-hove:
> fix in progress.

Though - great difficulties getting things pushed to GitHub today.

opm/core/utility/extractPvtTableIndex.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <vector>
+
+namespace Opm {
+
+void extractPvtTableIndex(std::vector<int> &pvtTableIdx,
+                          Opm::EclipseStateConstPtr eclState,
+                          size_t numCompressed,
+                          const int *compressedToCartesianCellIdx)
+{
+    //Get the PVTNUM data
```
atgeirr: I have a distinct feeling of deja vu here. Have we not implemented this kind of extraction before? I fail to remember where though.

opm/core/props/IncompPropertiesFromDeck.hpp
```diff
@@ -23,8 +23,8 @@
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <opm/core/props/IncompPropertiesInterface.hpp>
-#include <opm/core/props/rock/RockFromDeck.hpp>
 #include <opm/core/props/pvt/PvtPropertiesIncompFromDeck.hpp>
+#include <opm/core/props/rock/RockFromDeck.hpp>
```
atgeirr: I may be a little blind, but is this a change?

atgeirr:
Overall this is a good effort, it will be good to be rid of the duplicate implementations. Some things seem to me to be a bit less than elegantly implemented (with some duplication of code), but that is OK, given that this supports a legacy interface that may not be maintained indefinitely.

The major place this is used in the Flow chain of components is for intialization and threshold pressure calculations (threshold pressure calcs not exercised by Norne though). I would appreciate if @osae for example could verify that this does not change initialization results. I am a little worried about the differences in well curves, since this code should not be used at all after the init phase, so perhaps the init results are changed more than they should be?

andlaus:
> I am a little worried about the differences in well curves, since this code should not be used at all after the init phase, so perhaps the init results are changed more than they should be?

you're right: it should not be used after the init, and I think I even verified with gdb that it isn't. I suspect the differences of the well curves are because the last few bits of many values of the initial initial condition are bound to be different and it seems like the wells code is sensitive to this. I agree that this PR needs be thoroughly tested and looked at, though.

opm/core/utility/extractPvtTableIndex.cpp
andlaus: yeah, it was in `PvtInterface.[ch]pp` before, bit the class got removed so I figured it would be better to rename the file...

for the deja-vu: I have the same feeling, but the reason for is that there are so many kinds of regions (SATNUM, IMBNUM, EQLNUM, FLUXNUM, PVTNUM just to mention the ones which immediately came to my mind) and each of them basically needs the same code. probably it can be unified passing the name of the region keyword as an argument.

opm/core/props/IncompPropertiesFromDeck.hpp
andlaus: Either it's a artifact from development or `IncompPropertiesFromDeck` uses `RockFromDeck` and the file was previously included by transitivity. I'll check.

andlaus:
ok. I've rebased this PR on the current master and fixed the fallout which OPM/opm-parser#677 caused in it.

opm/core/props/IncompPropertiesFromDeck.hpp
andlaus: ok, it is the transivity: Without the include I get

```
In file included from /home/and/src/opm-core/opm/core/props/IncompPropertiesFromDeck.cpp:22:
/home/and/src/opm-core/opm/core/props/IncompPropertiesFromDeck.hpp:139:9: error: unknown type name 'RockFromDeck'
        RockFromDeck rock_;
```

andlaus:
are there any news on this? (it slowly grows a beard. ;))

andlaus:
Ok, I've rebased this PR, so it can now be merged without conflicts again.

totto82:
Good, I will test it and report back.

totto82:
There is a merge conflict in OPM/opm-autodiff#588. @andlaus  can you fix that so I can test this PR.

andlaus:
ok, OPM/opm-autodiff#588 has been fixed...

totto82:
I have looked at the initialization. The difference between opm-material and opm-core << the difference between opm and ecl.

andlaus:
sorry? (your comment seems to be scrambled.)

totto82:
<< much less

andlaus:
ah, I thought this were some kind of quotation characters. stupid me. Thanks for testing!

atgeirr:
Great! Thanks for the patch @andlaus and testing @totto82. Merging now.

PR #958, opened 2016-02-19
use standard filename for output log file.

qilicun:
If the deck file is **XXX.DATA**, output log file will be **XXX.LOG** in the same directory.

### DIFF
--- a/examples/diagnose_relperm.cpp
+++ b/examples/diagnose_relperm.cpp
@@ -86,7 +86,7 @@ try
         baseName = path(fpath.filename()).string();
     }
 
-    std::string logFile = baseName + ".LOG";
+    std::string logFile = baseName + ".SATFUNCLOG";
     Opm::time::StopWatch timer;
     timer.start();
     RelpermDiagnostics diagnostic(logFile);
--- a/examples/diagnose_relperm.cpp
+++ b/examples/diagnose_relperm.cpp
@@ -36,6 +36,7 @@
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 
 #include <boost/filesystem.hpp>
+#include <boost/algorithm/string/case_conv.hpp>
 #include <memory>
 #include <algorithm>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


atgeirr:
Seems reasonable, I just have one worry: should not the XXX.LOG filename be reserved for full simulator output log, so that we risk overwriting it with this diagnostic tool? What about XXX.RELPERMLOG?

qilicun:
>so that we risk overwriting it with this diagnostic tool? 
That's true.

>What about XXX.RELPERMLOG?
But this is a standard-alone binary,  I can suppose that the user just use it for relperm diagnostics.

For the simulator output, it should be XXX.PRT I guess.

alfbr:
Yes please, the output file should be named XXX.PRT. The log is basically the terminal output, while more exhaustive details are as expected in the PRT file. We want to overwrite the PRT file, this is the beginning of OPM's PRT and LOG files :-)

alfbr:
My apologies, I mixed this pull request with the one for integrating the logging in flow. Since this is a stand-alone code, only intended for relperm diagnostics, I do agree with Atgeirr, we should name it accordingly. I suggest naming SATFUNCDIAGNOSIS.LOG. No strong opinions though, feel free to use another descriptive name of the log file.

For the PRT and LOG file in flow, my comments above apply though.

qilicun:
@alfbr 

``the output file should be named XXX.PRT``

Yes, so that means ``.LOG`` file will not used by flow, we can keep this for relperm diagnostics. If it's so, @atgeirr will not worry that will be overwrite by flow. If you all agree, I think this PR should go in. :)

atgeirr:
> Yes, so that means .LOG file will not used by flow, we can keep this for relperm diagnostics

I think this is wrong, that both .PRT and .LOG names should eventually be used by Flow, if I understood @alfbr correctly.

Therefore the file name used by this code should have a different ending (especially if you want to use the casename as base), for example .RELPERMLOG as I proposed earlier.

alfbr:
I see, @atgeirr proposes to use the casename as prefix also for the relpermdiagnostics log-file, and extend the suffix to make it clear that the log-file is only from the relpermdiagnostics (and not flow). That makes sense I guess. I suggest using CASENAME.SATFUNCLOG instead though, capillary pressures and J-functions are quickly part of the diagnostics too. What do you think @qilicun ?

qilicun:
I have no strong opinion on this, I think ``.SATFUNCLOG`` is a good name, will update this PR.

atgeirr:
Thanks for updating, will merge!

PR #959, opened 2016-02-21
Fix build

andlaus:
this is some fallout of OPM/opm-parser#687. the first patch fixes the deprecation warnings whilst the second one fixes a build issue.

While the second patch is quite trivial (some forgotten 'const'), the havoc was caused because I usually configure my modules with --disable-tests (to get much better turn-around times when switching all modules from debug to optimization flags) and the usual way to force them to compile ('make test-suite') does not work for opm-core...


bska:

PR #962, opened 2016-02-25
Replaced SimulatorState -> SimulationDatacontainer

joakim-hove:
Have removed the `SimulatorState` base class, and instead replaced with the `SimulationDatacontainer` class from opm-common. The `SimulatorState` objects were typcially created with a default constructor, and then explicitly initialized with a `SimulatorState::init()` method. For the `SimulationDataContainer` RAII is employed; the `init( )` has been removed - and there is no default constructor.

Depends on:https://github.com/OPM/opm-common/pulls
Downstream PR in opm-autodiff in progress.


bska:
Any idea what's causing the test failure
```
2 - writeReadRestartFile (Failed)
```
in Travis?

joakim-hove:
> Any idea what's causing the test failure

Hmm - no; It works on my machine&trade;. Will look into it.

bska:
>> Any idea what's causing the test failure
> 
> Hmm - no; It works on my machine™. Will look into it.

I'll run the tests on my computer too.

joakim-hove:
> I'll run the tests on my computer too.

Don't spend too much effort; as indicated this PR *is* work in progress.

bska:
>> I'll run the tests on my computer too.
> 
> Don't spend too much effort; as indicated this PR is work in progress.

I'll just run a build and the associate tests and report back.  That's very close to zero effort :wink:

bska:
> I'll just run a build and the associate tests and report back

For what it's worth I see the same symptoms in both debug and release modes in a local build targeting Dune 2.3.1 (Ubuntu 15.10, GCC 5.2.1, Boost 1.58.0).

joakim-hove:
OK - I have fixed the failing test (it failed at my place as well ....)

atgeirr:
Just a thought (since this is a work in progress): would it not be better at least as a first step, to put a
```
using SimulatorState = SimulationDataContainer;
```
in SimulatorState.hpp? That way we'd not have to deal with renaming (and I think SimulatorState is still a usable name) at the same time as the behaviour change (init->constructors).

joakim-hove:
> Just a thought (since this is a work in progress): would it not be better at least as a first step, to put a

I guess that was a good idea - but it is not so much work in progress; on the contrary I know think it is getting ready. Corresponding PR in autodiff: https://github.com/OPM/opm-autodiff/pull/602

joakim-hove:
> and I think SimulatorState is still a usable name

That I agree - I would be positive to a all in rename `s/SimulationDataContainer/SimulatorState/`.

joakim-hove:
Ahhhh - sorry; I have made some git f... up here. Will update the branches later today.

PR #965, opened 2016-02-29
Remove unused 'restart' parameter from SimulatorTimer::init().

atgeirr:
Requires follow-up PR in opm-autodiff.


atgeirr:
No comments, will self-merge this and the follow-up PR.

PR #968, opened 2016-03-04
make grid as template parameter.

qilicun:
See comments in https://github.com/OPM/opm-autodiff/pull/593


atgeirr:
Thanks! I'll merge this and trigger a Travis rebuild of the autodiff PR.

atgeirr:
I was a little too quick merging this, sorry for that. The main header needs to include the _impl.hpp header at the end. I'll submit and fix this now.

PR #969, opened 2016-03-04
Fix minor issues with RelpermDiagnostics

atgeirr:
This fixes a few issues resulting from the change to make two of the methods templates.


atgeirr:
Will self-merge since this is a necessary follow-up to an already-merged PR.

PR #970, opened 2016-03-08
BlackoilPropertiesFromDeck: properly calculate the Rs and Rv values for viscosity

andlaus:
the dissolution factors used for the viscosities were always zero so
far. this was not discovered earlier because flow is completely
unaffected by this since the only place where this class is in flow is
the equilibration code and the equilibration code does not need the
phase viscosities.

thanks to @atgeirr for finding this.


atgeirr:
Thanks for a quick fix! Merging.

PR #971, opened 2016-03-08
threshold pressures: consider the saturated case and fix a typo

andlaus:
the typo was caused the surface density of the oil phase to be used
instead of the one of gas. This caused the density to be off by a
factor of typically about 900.

using saturated FVFs does not change much, but it does not hurt
because it is also done that way in the simulator.

This makes the defaults for the threshold pressures reasonable again,
but for some reason they are not exactly the same as in the old
implementation. (although the differences are very tolerable.)

On the question why only "Model 2" is affected by this: the other
decks don't use threshold pressures (SPE-X) or do not default any
values (Norne).


totto82:
Thanks. I will test it.

totto82:
With is PR the results are back to normal for model 2.

atgeirr:
Code looks good to me, all checks and tests say go, so I'll merge. Thanks for the fix!

PR #973, opened 2016-03-14
Fix PERM[XYZ] output

andlaus:
this *might* fix the PERM[XYZ] fields in the ECL restart files. notice the emphasis on the *might*: this is because I could not properly test the patch because ResInsight seems to have issues with the restart file produced by `flow`. (I haven't found a way to make it display the PERM[XYZ] fields. I have tested that this is the case for both this PR's branch as well as well as the current opm-core and opm-autodiff master versions.) So:

Before you merge this, please test carefully
==

### DIFF
--- a/opm/core/io/eclipse/EclipseWriter.cpp
+++ b/opm/core/io/eclipse/EclipseWriter.cpp
@@ -1184,16 +1184,19 @@ void EclipseWriter::writeInit(const SimulatorTimerInterface &timer)
         if (eclipseState_->hasDeckDoubleGridProperty("PERMX")) {
             auto data = eclipseState_->getDoubleGridProperty("PERMX")->getData();
             EclipseWriterDetails::convertFromSiTo(data, Opm::prefix::milli * Opm::unit::darcy);
+            EclipseWriterDetails::restrictAndReorderToActiveCells(data, gridToEclipseIdx_.size(), gridToEclipseIdx_.data());
             fortio.writeKeyword("PERMX", data);
         }
         if (eclipseState_->hasDeckDoubleGridProperty("PERMY")) {
             auto data = eclipseState_->getDoubleGridProperty("PERMY")->getData();
             EclipseWriterDetails::convertFromSiTo(data, Opm::prefix::milli * Opm::unit::darcy);
+            EclipseWriterDetails::restrictAndReorderToActiveCells(data, gridToEclipseIdx_.size(), gridToEclipseIdx_.data());
             fortio.writeKeyword("PERMY", data);
         }
         if (eclipseState_->hasDeckDoubleGridProperty("PERMZ")) {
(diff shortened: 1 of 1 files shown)


joakim-hove:
I agree that the problem is that OPM version writes all cells, whereas eclipse only writes the active cells; so this fix should be OK. That said I really did not like the `swap( )` based implementation of `EclipseWriterDetails::restrictAndReorderToActiveCells()` - I strongly prefer that function to *return* the restricted values - although I realize that was not part of this PR.

> I could not properly test the patch because ResInsight seems to have issues with the restart file produced by

That might be just a keyword slip on your part? But at least for the code we are talking about here it is the `INIT` file which should be loaded - *not* the restart file.

joakim-hove:
If you agree on the `EclipseWriterDetails::restrictAndReorderToActiveCells()` issue you can fix; otherwise I am ready to merge this.

andlaus:
> If you agree on the EclipseWriterDetails::restrictAndReorderToActiveCells() issue you can fix; otherwise I am ready to merge this.

I can (and are even willing to ;)) make it work that way, but I think that's quite out of scope for the present PR?

>>   I could not properly test the patch because ResInsight seems to have issues with the restart file produced by
>
> That might be just a keyword slip on your part? But at least for the code we are talking about here it is the INIT file which should be loaded - not the restart file.

true, these fields are in the init file. What I meant is that I used `ResInsight --case NORNE_ATW2013` to load the output produced by `flow` and did not see any PERM fields to inspect. That said, maybe the case must be loaded differently. (Of course, another possibility is that I'm just too stupid to push the right buttons in the UI.)

joakim-hove:
> I can (and are even willing to ;)) make it work that way, but I think that's quite out of scope for the present PR?

Oh absolutely outside the scope of current PR - but still a clear improvement?

> What I meant is that I used ResInsight --case NORNE_ATW2013 to load the output produced by flow and did not see any PERM fields to inspect.

OK - been long time since I gave Resinsight a swing; will do tomorrow.

andlaus:
> Oh absolutely outside the scope of current PR - but still a clear improvement?

sure. I think the reason why it is implemented as it is is to save some memory, but given the fact that I'm not the original author of this functionality I can only speculate...

> OK - been long time since I gave Resinsight a swing; will do tomorrow.

if the Eclipse visualization tools are happy, we're golden IMO... (I can't test this for obvious reasons.)

joakim-hove:
I'll merge it; seems ok.

joakim-hove:
> (Of course, another possibility is that I'm just too stupid to push the right buttons in the UI.)

Hehe - I actually think that "is the case" - at least I could select "static properties" in the dropdown in the top right corner and choose `PERMX`. I assume that Resinsight will handle arrays with `nactive` or `nx*ny*nz` elements transparently.

andlaus:
ok, thanks for the hint.

PR #974, opened 2016-03-15
AdaptiveTimeStepping: also write the initial solution

andlaus:
The AdaptiveTimeStepping is the only class which has a complete enough
picture to write result files and because that functionallity should
not be scattered any further in the code, it should also write the
result file for the initial solution.

IMO this is pretty hacky (we pretend to write a time step of length
0), but with the current APIs it is the only way to do this which I
could come up with. (And given the prospect that the output writing
code is soon to be refactored within new opm-output module, I advice
not to do major API changes in that code anymore.)

this should fix the travis errors encountered by OPM/opm-autodiff#613.


atgeirr:
Would not this give different output (including the initial step or not) depending on whether the adaptive time-stepper is used?

andlaus:
yes it would. but isn't that the case already? (not sure)

andlaus:
I just had a look: I could not find any time stepping scheme (actually: code which calls `*solver*.step()`) in opm-core

atgeirr:
> I could not find any time stepping scheme (actually: code which calls *solver*.step()) in opm-core

That call would not be there but in SimulatorBase_impl.hpp line 176 (in opm-autodiff).

andlaus:
> That call would not be there but in SimulatorBase_impl.hpp line 176 (in opm-autodiff).

with  OPM/opm-autodiff#613 this would not write any output at all anymore? I'll look into this mess...

akva2:
Can one of the admins verify this patch?

andlaus:
OPM/opm-output #12 seems to be a more appropriate fix. closing.

PR #975, opened 2016-03-16
Remove fast access methods state.pressure() +++

joakim-hove:
For discussion:


atgeirr:
I think this is not needed anymore, so I suggest it is closed. Opinion @joakim-hove?

joakim-hove:
> so I suggest it is closed.

Sure.

PR #977, opened 2016-03-17
Remove output writer

kjellkongsvik:
This PR temporarily removes functionality as per the 2 commits with commented out usage of OutputWriter.

This PR depends on 2 other PRs in opm-common and opm-autodiff that will very soon be made

The functionality removed in the 2 first commits should be readded when the dependencies have been fixed


kjellkongsvik:
Ref: https://github.com/OPM/opm-autodiff/pull/619 and https://github.com/OPM/opm-common/pull/89

PR #982, opened 2016-03-29
Allow SimulatorTimers to be copied.

dr-robertk:
Since we are dealing with a virtual interface here we needed to add a method clone that creates a copy of the real object. This is needed in an upcoming PR in opm-autodiff.

### DIFF
--- a/opm/core/simulator/SimulatorTimerInterface.hpp
+++ b/opm/core/simulator/SimulatorTimerInterface.hpp
@@ -20,6 +20,8 @@
 #ifndef OPM_SIMULATORTIMERINTERFACE_HEADER_INCLUDED
 #define OPM_SIMULATORTIMERINTERFACE_HEADER_INCLUDED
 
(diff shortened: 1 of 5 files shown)


opm/core/simulator/SimulatorTimerInterface.hpp
```diff
@@ -98,6 +98,9 @@ namespace Opm
             tm t = boost::posix_time::to_tm(currentDateTime());
             return std::mktime(&t);
         }
+
+        /// return copy of current timer instance
+        virtual std::unique_ptr< SimulatorTimerInterface > clone () const = 0;
```
atgeirr: Should include <memory> for unique_ptr.

dr-robertk: Fixed.

atgeirr:
Looks good, will merge (after Travis is done). Thanks for the patch!

PR #983, opened 2016-03-31
remove const of ParseContext.

qilicun:
Depend on https://github.com/OPM/opm-parser/pull/738


atgeirr:
Since it is a follow-up to a closed PR and we are going to do things in a different way I'll close this.

PR #984, opened 2016-03-31
BUGFIX. Call setupGuideRates before groupcontrol is applied

totto82:
With this fix guide rates for group control can be specified correctly using WRUPCON


atgeirr:
Thanks for the fix!

PR #988, opened 2016-04-01
Calculate default guide rates using the well potentials

totto82:
The default guide rates are calculated using the well potentials.
The well potentials are calculated in the simulator and given as input
to the wellsManager.

### DIFF
--- a/opm/core/wells/WellsManager.cpp
+++ b/opm/core/wells/WellsManager.cpp
@@ -1,6 +1,6 @@
 /*
   Copyright 2012 SINTEF ICT, Applied Mathematics.
-  Copyright 2015 IRIS AS
+  Copyright 2016 IRIS AS
 
   This file is part of the Open Porous Media project (OPM).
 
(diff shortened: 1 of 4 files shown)


opm/core/wells/WellsManager.cpp
```diff
@@ -330,11 +331,12 @@ namespace Opm
                                const double* permeability)
         : w_(0), is_parallel_run_(false)
     {
+        std::vector<double> well_potensials;
```
atgeirr: Typo: potensials -> potentials

opm/core/wells/WellsManager.hpp
```diff
@@ -86,7 +86,8 @@ namespace Opm
                      const F2C& f2c,
                      FC begin_face_centroids,
                      const double* permeability,
-                     bool is_parallel_run=false);
+                     bool is_parallel_run=false,
+                     const std::vector<double> well_potentials={});
```
atgeirr: Should be a reference argument (also in other methods).

atgeirr:
Looks good, will merge!

totto82:
Doesn't this PR need to be merged together with OPM/opm-autodiff#627. The wells_manager interface is changed.

atgeirr:
> Doesn't this PR need to be merged together with OPM/opm-autodiff#627.

I think the default parameter allows for this to be merged separately. In case I failed I intend to merge the other PR after a build check very soon.

totto82:
You  are right, I forgot I added a default.

PR #991, opened 2016-04-05
If opm failed to create a grid, record it to Log file.

qilicun:
1. GridManager is related to deck file, if insufficient data is provided, the error message should go into log file. 
2. Because the stop criterion is not well-desinged with OpmLog, so we don't need to delete the THROW at this time.

### DIFF
--- a/opm/core/grid/GridManager.cpp
+++ b/opm/core/grid/GridManager.cpp
@@ -113,7 +113,7 @@ namespace Opm
     {
         ug_ = read_grid(input_filename.c_str());
         if (!ug_) {
-            OpmLog::error("Failed to read grid from file"+input_filename);
+            OpmLog::error("Failed to read grid from file" + input_filename);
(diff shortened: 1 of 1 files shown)


opm/core/grid/GridManager.cpp
```diff
@@ -111,6 +113,7 @@ namespace Opm
     {
         ug_ = read_grid(input_filename.c_str());
         if (!ug_) {
+            OpmLog::error("Failed to read grid from file"+input_filename);
```
atgeirr: Use space around binary operators like `+`.

atgeirr:
This is fine, and will go in when the tiny nitpick is addressed. I agree with keeping the throws. We should figure out how stopping should relate to the logging, if at all.

atgeirr:
I was originally going to merge this, but after the discussion in OPM/opm-autodiff#637 I think we should rather move forward with that idea (putting OpmLog::error() in the OPM_THROW macro).

PR #993, opened 2016-04-07
Removed SimulatorOutput class from this module.

atgeirr:
This class is only used from opm-verteq and will be moved to there.


atgeirr:
I will self-merge this as well as the opm-verteq PR as it should have no effect on other parts.

PR #994, opened 2016-04-11
Use Eclipse3DProperties

pgdr:
This pull request uses the Eclipse3DProperties API from opm-parser.

CompressedPropertyAccess.hpp needs a thorough read-through before merge.

Ref OPM/opm-parser#751

### DIFF
--- a/opm/core/props/phaseUsageFromDeck.hpp
+++ b/opm/core/props/phaseUsageFromDeck.hpp
@@ -25,6 +25,7 @@
 
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/TableManager.hpp>
 
 
 namespace Opm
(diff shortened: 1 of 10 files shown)


joakim-hove:
jenkins build this opm-parser=751 opm-material=137 please

bska:
jenkins build this opm-parser=751 opm-material=137 please

pgdr:
jenkins build this opm-parser=751 opm-material=137 opm-output=14 please

bska:
jenkins build this opm-parser=751 opm-material=137 opm-output=14 please

jokva:
@bska Is this ok for merge now?

bska:
> Is this ok for merge now?

I need to look more closely at the `CompressedPropertyAccess` changes before I can sign off on that.

jokva:
Alright, take your time. Once you give the green light this will be merged upstream.

bska:
Okay, I've looked at this.  The changes to `CompressedPropertyAccess.hpp` are okay if (and only if)
```
&ecl->get3DProperties().get [TYPE] GridProperty(kw)
```
is valid for the entire lifetime of a `Compressed<>` object that is backed by an array which we `ExtractFromDeck<>`.  In this context `kw` is a keyword string that identifies a property which is known to exist in the input deck.  I specifically require that `get*Property()` does **not** return a property which is vivified and discarded on demand.

jokva:
As long as you keep the `EclipseState` instance alive this shouldn't be a problem. Once a property is created it lives for as long as state does, and subsequent calls will return the same instance. Properties are a static set once EclipseState is constructed.

I'll have a second look over possible pointer invalidations just to be sure.

jokva:
@bksa @pgdr So I just had a look. There is an `addKeyword` feature that may break this guarantee (well actually, currently the instances are stored with `shared_ptr`s so now it's  fine, but this is likely to change) on `GridProperties` which is used in `get`. This means that we *currently* can guarantee it, but not for the future. I'll think a bit on this and discuss with @pgdr tomrrow and figure out what to do.

I'll merge once we can guarantee the property you seek.

joakim-hove:
> does not return a property which is vivified on demand and subsequently discarded.

Well - then you will be happy; because that is what is supplied. Previously the `GridProperty<T>` container had public autovivification; i.e. you could do e.g. getKeyword("FIPNUM")` and if the `FIPNUM` keyword was not present it would be autocreated. Not so any longer - current implementation will scan through the input deck in the constructor and create and initialize all keywords found in the input deck; after that the container is immutable.

The lifetime of the keyword is the same as the `Ecl3DProperties` - which is the same as the `EclipseState`; which *really* should have application lifetime.

With this PR the creation/initialization of properties in the parser has improved tremendously.

PR #997, opened 2016-04-12
Bugfix in setup guiderates

totto82:
- Handle shut wells
- Use the groups control type to determine which phase to calculate
the guide rates from. i.e for a ORAT controlled group, calculate the
guide rates from the oil phase well potentials etc.

### DIFF
--- a/opm/core/wells/WellCollection.cpp
+++ b/opm/core/wells/WellCollection.cpp
@@ -61,6 +61,11 @@ namespace Opm
     }
 
     void WellCollection::addWell(WellConstPtr wellChild, size_t timeStep, const PhaseUsage& phaseUsage) {
+        if (wellChild->getStatus(timeStep) == WellCommon::SHUT) {
+            //SHUT wells are not added to the well collection
+            return;
+        }
+
         WellsGroupInterface* parent = findNode(wellChild->getGroupName(timeStep));
         if (!parent) {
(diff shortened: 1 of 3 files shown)


atgeirr:
Travis reports the following error:
```
30/38 Test #30: wellcollection ...................***Failed    0.02 sec
Running 1 test case...
ERROR when handling keyword for well INJ1: Cannot open a well where all completions are shut
ERROR when handling keyword for well INJ2: Cannot open a well where all completions are shut
ERROR when handling keyword for well PROD1: Cannot open a well where all completions are shut
ERROR when handling keyword for well PROD2: Cannot open a well where all completions are shut
unknown location(0): fatal error in "AddWellsAndGroupToCollection": memory access violation at address: 0x00000008: no mapping at fault address
/home/travis/build/OPM/opm-core/tests/test_wellcollection.cpp(89): last checkpoint
*** 1 failure detected in test suite "WellCollectionTest"
```

Do you have insight into what happens here? I do not think it should be merged with the test failures.

Also: is this independent of OPM/opm-simulators#648?

totto82:
COMPDAT was missing in the deck. No completion -> shut well -> no well in the collection. 

This is independent of OPM/opm-simulators#648.

atgeirr:
Thanks for the fix, will merge!

PR #1000, opened 2016-04-18
Updates for 16.04 packaging

akva2:
- update redhat packaging. use boost 1.48 only on rhel6, rhel7 supplies a newer boost by default.
- update debian packaging. build with MPI enabled and add related dependencies
- generic: README file was renamed.

note redhat is not yet MPI enabled. i haven't fought through packaging ptscotch / zoltan yet and didn't want that to hold up the branching.

@atgeirr please hit the green unless there are objections.


atgeirr:
I did not notice yesterday, but of course the big Pullitzer Request Prize goes to @akva2 for creating pull request number 1000 in this repo!

bska:
> pull request number 1000 in this repo!

Really, we're going with base 10?-)

atgeirr:
> Really, we're going with base 10?-)

Don't worry, and keep an eye on 1024...

Issue #1001, opened 2016-04-19
Some tests break after OPM/opm-parser#768


pgdr:
I broke three tests (and probably some examples) in opm-core with OPM/opm-parser#768.

The deal is the following:  EclipseGrid will no longer respect the ACTNUM keyword, and we must give EclipseGrid the actnum vector explicitly.  This was done to accept BOXed ACTNUM.

I'm working on the issue, and don't know why jenkins didn't catch this prior to merge.

pgdr:
Fixed and merged in OPM/opm-parser#1002.

PR #1002, opened 2016-04-19
 Using getInputGrid API from Parser

pgdr:
This PR adds two fixes
* First, now we are using the new EclipseState.getInputGrid API in Parser
* Second, this PR fixes a bug introduced in ACTNUM PR OPM/opm-parser#768 (see issue #1001)
* Third, fixes arbitrary indentation in test file

See OPM/opm-parser#769

### DIFF
--- a/tests/test_blackoilstate.cpp
+++ b/tests/test_blackoilstate.cpp
@@ -28,6 +28,13 @@ using namespace Opm;
 using namespace std;
 
 
+// ACTNUM 1 998*2 3
+std::vector<int> get_testBlackoilStateActnum() {
+    std::vector<int> actnum(10 * 10 * 10, 2);
+    actnum.front() = 1;
+    actnum.back() = 3;
+    return actnum;
+}
 
 BOOST_AUTO_TEST_CASE(EqualsDifferentDeckReturnFalse) {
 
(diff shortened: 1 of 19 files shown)


joakim-hove:
jenkins build this opm-parser=769

tests/test_blackoilstate.cpp
```diff
@@ -28,19 +28,38 @@ using namespace Opm;
 using namespace std;
 
 
+// ACTNUM 1 998*2 3
+static std::vector<int> get_testBlackoilStateActnum() {
+    std::vector<int> actnum;
+    for (size_t i = 1; i <= (10 * 10 * 10); i++)
```
bska: Is there a reason for not writing this as
```C++
TYPE actnum(10 * 10 * 10, 2);
actnum.front() = 1;
actnum.back() = 3;
```
?

bska: And as for `static`, this is C++ so we should use anonymous namespaces to achieve the same effect.

bska:
jenkins build this opm-parser=769 opm-material=141 please

tests/test_blackoilstate.cpp
```diff
@@ -28,19 +28,34 @@ using namespace Opm;
 using namespace std;
 
 
+// ACTNUM 1 998*2 3
+std::vector<int> get_testBlackoilStateActnum() {
+    std::vector<int> actnum(10 * 10 * 10, 2);
```
pgdr: @bska The answer to you question "why" is simply because I didn't think of it.  Fixed.

tests/test_blackoilstate.cpp
```diff
@@ -28,19 +28,34 @@ using namespace Opm;
...
+    actnum.front() = 1;
+    actnum.back() = 3;
```
atgeirr: I do not understand the purpose of this, what is the meaning of actnum other than 0 or 1?

bska: > what is the meaning of actnum other than 0 or 1?

Doubling as `SATNUM`.

atgeirr: OK, a bit confusing but I get it.

pgdr: This is simply an internalization of `ACTNUM 1 998*2 3` as it is set in the file `testBlackoilState1.DATA`.  If it is confusing, it is because the test file is.

joakim-hove: > I do not understand the purpose of this, what is the meaning of actnum other than 0 or 1?

This particular test aside `ACTNUM` can actually have values different from 0 and 1; in dual porosity models `ACTNUM == 2` denotes a *fracture cell* - and there is also some use for `ACTNUM > 1` in ECLIPSE300.

PR #1003, opened 2016-04-22
Two phase bugfixes

babrodtk:
This PR enables running a two phase test case with plausible results

This PR should have no side effects for running existing three phase simulations. See also the twin PR in opm-simulators which should be merged simultaneously with this one


GitPaean:
One general question, not necessarily related to this PR. This PR looks like for oil and gas two-phase simulation. 

Do we need oil and water two-phase simulation? or even if there is any requirement for water and gas two-phase simulation?

pgdr:
parser or simulators?

babrodtk:
The twin PR is in opm-simulators, thanks spotting!

PR #1005, opened 2016-04-29
added: downstream build support in the jenkins trigger

akva2:
use 'jenkins build this with downstreams please'.

this will build all downstreams against the PR, and
execute their tests. PR's for upstream and downstream modules
can be specified.

the aggregated test results are attached to the job result on jenkins.

depends on https://github.com/OPM/opm-common/pull/111


akva2:
@bska @atgeirr ready for review

PR #1009, opened 2016-05-06
Prevent dereferencing the end iterator.

### LINKED ISSUE #1008: Dereferencing of end iterator happens in initStateEquil.hpp
InitialStateComputer OPM/opm-core/blob/131e900cf32cb3b55d8c4a83dc3a1a8904f70a8c/opm/core/simulator/initStateEquil.hpp#L267 has a problem for empty ranges. For one of testcases that I have received this is a problem as `eqlmap.cells(i).begin()==eqlmap.cells(i).end()` holds.

Even if it does not crash it will add bogus entries to `rs_func_`

blattms:
This happened for empty regions and surfaced when compiling
with "-D_GLIBCXX_DEBUG  -DDEBUG -DGLIBCXX_FORCE_NEW".

This closes #1008, but I am not sure whether pushing nullpointers to rs_func_ and rv_func_ is a good idea.

### DIFF
--- a/opm/core/simulator/initStateEquil.hpp
+++ b/opm/core/simulator/initStateEquil.hpp
@@ -264,6 +264,11 @@ namespace Opm
                     if (deck->hasKeyword("DISGAS")) {                    
                         const TableContainer& rsvdTables = tables.getRsvdTables();
                         for (size_t i = 0; i < rec.size(); ++i) {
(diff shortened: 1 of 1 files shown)


atgeirr:
I think this is a good workaround, a better solution might involve mapping only the actually used regions but that would require a more significant rewrite. Thanks for the patch, merging!

PR #1010, opened 2016-05-09
Log timer messages

qilicun:
Log timer messages to log file.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -245,12 +245,12 @@ namespace Opm {
                 const double newTimeStep = restart_factor_ * dt;
                 // we need to revise this
                 substepTimer.provideTimeStepEstimate( newTimeStep );
-                if( solver_verbose_ )
+                if( solver_verbose_ ) {
                     std::string msg;
                     msg = "Solver convergence failed, restarting solver with new time step ("
                         + std::to_string(unit::convert::to( newTimeStep, unit::day )) + " days).\n";
                     OpmLog::error(msg);
-
+		}
                 // reset states
                 state      = last_state;
                 well_state = last_well_state;
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -150,15 +150,16 @@ namespace Opm {
         int restarts = 0;
 
         // sub step time loop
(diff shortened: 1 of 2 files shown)


opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -258,13 +259,15 @@ namespace Opm {
...
-            substepTimer.report( std::cout );
-            std::cout << "Suggested next step size = " << unit::convert::to( suggested_next_timestep_, unit::day ) << " (days)" << std::endl;
+            substepTimer.report(ss);
+            ss << "Suggested next step size = " << unit::convert::to( suggested_next_timestep_, unit::day ) << " (days)" << std::endl;
         }
 
         if( ! std::isfinite( suggested_next_timestep_ ) ) { // check for NaN
             suggested_next_timestep_ = timestep;
         }
+        std::cout << ss.str();
```
joakim-hove: Instead of writing to `std::cout` directly I would add a  `StreamLog( std::cout )` instance in `flow`.

qilicun: >Instead of writing to std::cout directly I would add a StreamLog( std::cout ) instance in flow

I agree, that improve the readability of the codes.

atgeirr: I think it is not correct to collect all the reporting in a string here and then log (or print) it, we should rather use the OpmLog facility for each single message.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -242,8 +246,10 @@ namespace Opm {
                 // we need to revise this
                 substepTimer.provideTimeStepEstimate( newTimeStep );
                 if( solver_verbose_ )
```
atgeirr: Missing braces, must be added.

In this case, it did not compile on my system so I caught it. (Why does this even compile on other systems? There is nothing else called msg in scope as far as I can see?).

To @joakim-hove @bska and @pgdr: this is a typical example of the danger of not requiring braces: manual or automatic refactoring changes one statement to more.

joakim-hove: Hehe - it is a typical example of missing tests. Strict disciplin on code layout can *never* give you the safety of tests.

qilicun:
Does this can be merged?

atgeirr:
> Does this can be merged?

Sure!

PR #1011, opened 2016-05-10
Unify output messages format.

qilicun:
With https://github.com/OPM/opm-common/pull/115, the diagnostics messages will not output prefix twice.

### DIFF
--- a/opm/core/props/satfunc/RelpermDiagnostics.cpp
+++ b/opm/core/props/satfunc/RelpermDiagnostics.cpp
@@ -107,14 +107,14 @@ namespace Opm{
         bool family2 = ((!swfnTables.empty() && !sgfnTables.empty()) || !sgwfnTables.empty()) && (!sof3Tables.empty() || !sof2Tables.empty());
 
         if (family1 && family2) {
-            const std::string msg = "-- Error:   Saturation families should not be mixed.\n Use either SGOF and SWOF or SGFN, SWFN and SOF3.";
+            const std::string msg = "Error:   Saturation families should not be mixed.\n Use either SGOF and SWOF or SGFN, SWFN and SOF3.";
             messages_.push_back(msg);
             OpmLog::error(msg);
             counter_.error += 1;
         }
 
         if (!family1 && !family2) {
-            const std::string msg = "-- Error,   Saturations function must be specified using either \n \
+            const std::string msg = "Error,   Saturations function must be specified using either \n \
                              family 1 or family 2 keywords \n \
                              Use either SGOF and SWOF or SGFN, SWFN and SOF3.";
             messages_.push_back(msg);
@@ -217,7 +217,7 @@ namespace Opm{
         const std::string regionIdx = std::to_string(satnumIdx);
         //Check sw column.
         if (sw.front() < 0.0 || sw.back() > 1.0) {
-            const std::string msg = "-- Error:   In SWOF table SATNUM = "+ regionIdx + ", saturation should be in range [0,1].";
+            const std::string msg = "Error:   In SWOF table SATNUM = "+ regionIdx + ", saturation should be in range [0,1].";
             messages_.push_back(msg);
             OpmLog::error(msg);
             counter_.error += 1;
@@ -225,13 +225,13 @@ namespace Opm{
         //TODO check endpoint sw.back() == 1. - Sor.
         //Check krw column.
         if (krw.front() != 0.0) {
-            const std::string msg = "-- Error:   In SWOF table SATNUM = " + regionIdx + ", first value of krw should be 0.";
+            const std::string msg = "Error:   In SWOF table SATNUM = " + regionIdx + ", first value of krw should be 0.";
             messages_.push_back(msg);
             OpmLog::error(msg);
             counter_.error += 1;
         }
         if (krw.front() < 0.0 || krw.back() > 1.0) {
-            const std::string msg = "-- Error:   In SWOF table SATNUM = " + regionIdx + ", krw should be in range [0,1].";
+            const std::string msg = "Error:   In SWOF table SATNUM = " + regionIdx + ", krw should be in range [0,1].";
             messages_.push_back(msg);
             OpmLog::error(msg);
             counter_.error += 1;
@@ -239,7 +239,7 @@ namespace Opm{
 
         ///Check krow column.
         if (krow.front() > 1.0 || krow.back() < 0.0) {
-            const std::string msg = "-- Error:   In SWOF table SATNUM = "+ regionIdx + ", krow should be in range [0, 1].";
+            const std::string msg = "Error:   In SWOF table SATNUM = "+ regionIdx + ", krow should be in range [0, 1].";
             messages_.push_back(msg);
             OpmLog::error(msg);
             counter_.error += 1;
@@ -260,13 +260,13 @@ namespace Opm{
         const std::string regionIdx = std::to_string(satnumIdx);
         //Check sw column.
         if (sg.front() < 0.0 || sg.back() > 1.0) {
-            const std::string msg = "-- Error:   In SGOF table SATNUM = " + regionIdx + ", saturation should be in range [0,1].";
+            const std::string msg = "Error:   In SGOF table SATNUM = " + regionIdx + ", saturation should be in range [0,1].";
             messages_.push_back(msg);
             OpmLog::error(msg);
             counter_.error += 1;
         }
         if (sg.front() != 0.0) {
-            const std::string msg = "-- Error:   In SGOF table SATNUM = " + regionIdx + ", first value of sg should be 0.";
+            const std::string msg = "Error:   In SGOF table SATNUM = " + regionIdx + ", first value of sg should be 0.";
             messages_.push_back(msg);
             OpmLog::error(msg);
             counter_.error += 1;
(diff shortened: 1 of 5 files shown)


atgeirr:
The messages still have "Error: " or "Warning: " in them, I think that should be left to the decoration chosen for the backends in use, but of course that depends on OPM/opm-common#115 being accepted first.

qilicun:
The reason to keep the prefix here is that if you choose ``StreamLog`` to output put to the terminal with current implementation inside ``OpmLog``, all the messages will be output. In the diagnostics, we think it's better to output at most first 10 messages and warn the user to check the log file. To solve this issue, I can see two ways:
- For every message, store it with and without prefix into ``messages`` and ``OpmLog``separately.
- Let ``StreamLog`` have the ability to setting output limits.

The first one is the easiest, but it looks ugly.

atgeirr:
I agree that output of thousands of messages is problematic, especially when they are all essentially the same message, just for different cells. Perhaps this should be handled in a different way than logging a warning for each cell in the first place? Output limits are another way to deal with it, but a bit coarse (i.e. we want to suppress warnings after the 10 first, but only of the same kind, we do not want to suppress different warnings).

qilicun:
> Perhaps this should be handled in a different way than logging a warning for each cell in the first place

This is a possible way, but I have no idea right now.

qilicun:
Since  relperm diagnostics has a internal messages container to output messages to the terminal, is there any possible to shut down the streamLog output for relperm diagnostics?

atgeirr:
I think we should be able to use the OpmLog::warning() function (and the others) also for situations like here, where we expect a lot of messages. We just need to implement message limits in some way. However the difficulty lies in deciding what constitutes the same or similar messages. With a limit of for example 10 messages, we want only 10 of these:
```
Warning: For scaled endpoints input, cell 12345 SATNUM = 4, SOGCR + SGCR + SWL exceed 1.0
```
However if there is a different problem (with relperm for example), we should count those messages separately. I can see three ways to realize this.
 - Add a string with a tag indicating the warning (or error etc.) in question, and pass it to the OpmLog::warning() function. For the above warning it could be the string "Scaled endpoints sat over 1" for example. The string should not be the same as the message, because we want to collect similar messages with different data (different cells in the above).
 - Add an enum. As above, but we create a (big) registry of all warnings and messages in an enum kept somewhere central (opm-common). On the pro side this can serve as a documentation of possible errors and warnings, but it does require frequent updates to opm-common when adding new stuff to other modules (which produce new messages).
 - Add the ability to reset message counters somehow, and use it after each type of message is done with in the relperm diagnostics.

The first two are the most sophisticated variants, and require more significant implementation effort than the last suggestion, but perhaps they are the most valuable long term. For example MATLAB supports tags in its error() function: error('mySolver:didNotConverge', 'The solver did not converge because of bla bla bla.'). The tag is the first argument.

joakim-hove:
>Add an enum. As above, but we create a (big) registry of all warnings and messages in an enum kept somewhere central (opm-common). 

My suggestion would be a variety of this; we just use a simple string ID which is 100% controlled by the calling scope - i.e. *no central registry*. Then the decision: "Is this really the same warning" is controlled by the module invoking the logging, and within limit scope correct use / handling of message ID's should be manageable:

```C++
// Relperm example
std::string EXCEED_ONE = "Exceed_ONE";
std::string NEGATIVE_SGCR = "Neagative SGCR";

for (cell : grid ) {
    if ((SOGCR + SGCR + SWL) > 1.0)
       OpmLog::warning("For scaled endpoints input, cell 12345 SATNUM = 4, SOGCR + SGCR + SWL exceed 1.0" , EXCEED_ONE);

    if (SGCR < 0)
       OpmLog::warning("SGCR for cell 12345" , NEGATIVE_SGCR);
}
```

This should require roughly these changes to the `LogBackend` (or derived class?):

```C++
...
private:
    std::map<std::string , int> max_messages;
    std::map<std::string, int> message_count;
}
...

void configureMaxmessages( const std::string& messageTag, int max) {

}

```

And the actual logging backend must be configured / initialized in `main`scope before starting the run.

qilicun:
>Add an enum. As above, but we create a (big) registry of all warnings and messages in an enum kept somewhere central (opm-common). On the pro side this can serve as a documentation of possible errors and warnings, but it does require frequent updates to opm-common when adding new stuff to other modules (which produce new messages).

This sounds a good idea, for relperm diagnostics perspective, there are about 15 rules for checking. To achieve this,  I think more fundamental functions in OpmLog need to be done for not only output but also counting and logging, or furthermore: a good stop criterion for the simulation.

atgeirr:
To summarize: @qilicun prefers the enum solution, @joakim-hove prefers the string solution (I could not see that there were any significant differences between my first suggestion and @joakim-hove's, what are the differences?). If it came to a vote I would vote for a string based message tag solution as well.

joakim-hove:
> I could not see that there were any significant differences between my first suggestion and @joakim-hove's, what are the differences?

You are absolutely right - I just jumped the gun completely when I saw the solution based on central enum and failed to recognize that I essentially reiterated your first alternative.

**Enum based solution:**

1. Compile time safety.
2. *Strong* coupling between the various modules and the enum definitions in opm-common, unnecessary cumbersome to update. I fear that result will be that a system of enum based message tags will be under-used. 

For the string based solution I see the advantages and disadvantages as opposite to those which apply to the enum based solution; i.e. no compile time safety and much greater flexibility. In particular the tags relevant for a particular peace of code can be defined, configured and used locally in the code.

atgeirr:
Continuing this line of thought, I propose we create a proof-of-concept with the string tag idea. I volunteer to do that in opm-common. There are some things that are not nailed down, so I propose:
 - You can still use untagged messages, that is equivalent to an empty string tag.
 - An empty or missing tag means the message is not counted towards any tag limit.
 - Each log backend object has an object keeping count/limits, similar to the decoration feature.
 - Count limits are not stored per tag, but for all tags (unlike @joakim-hove's code example above), because I cannot see where this could be used: the place (such as the diagnostics) that knows its tags does not have access to any particular backend.
 - Count limits apply by tag, and do not care about message type (warning, error, info etc.).
 - Tags should be human readable, so that when the system writes: `Message limit reached (<num> messages) for tag <tag>.` (or similar) then it makes sense.

I'll start the proof-of-concept based on the above, so please chime in if you disagree or have more points to add.

joakim-hove:
> I volunteer to do that in opm-common.

Good plan.

> Count limits are not stored per tag, but for all tags (unlike @joakim-hove's code example above), because I cannot see where this could be used: the place (such as the diagnostics) that knows its tags does not have access to any particular backend.

I do not really understand what you mean here? As I see it the tag limit is consulted in the `LogBackend::includeMessage()` function.

atgeirr:
What I mean is: there is no place where the
```
void configureMaxmessages( const std::string& messageTag, int max);
```
function could be called from. The tags are only known locally where used, and that context does not know anything about any particular backends in use and therefore cannot configure them.

joakim-hove:
>The tags are only known locally where used, and that context does not know anything about any particular backends in use and therefore cannot configure them.

OK - I see; I thought a small function could be added to the same scope:

```C++
configureMaxRelpermMessages( Opm::LogBackend& backend ) {
     backend.addMaxMessage( "STRING1" , 100 );
     backend.addMaxMessage( "STRING2" , 200 );
}
```
It would then be responsability of application scope - i.e. typically `flow` to call this method, and related methods for other subsystems, after the backend has been instantiated and added to the main logger.

atgeirr:
I see, that is a good idea. In that way the Flow scope retains control of the backends and the local scope of the tags, and the function you suggest would connect the dots nicely. It is not in my current branch though (could be added later I think).

atgeirr:
I have submitted for review an implementation with message tags in OPM/opm-common#119.

To really test it, one should try using tags in the relperm diagnostics, and configure the terminal output loggger to have a limit (but not the PRT file logger), I have not yet done so (but the feature is unit tested in opm-common).

qilicun:
Just a comment: logfile should log all the messages, but terminal output can be limited.

qilicun:
According to https://github.com/OPM/opm-common/pull/119, I will update the PR.

qilicun:
Depend on https://github.com/OPM/opm-common/pull/119/

atgeirr:
It seems to me that you did not add tags to the error() and warning() calls in the .cpp file?

qilicun:
@atgeirr 

Yes, because in ``.cpp`` file, the ``error`` and ``warning`` are all about saturation functions, if they have errors, at most there are just 4-5, which has no relation with grid cells. Of cause, if there are lots of regions, the situation will be different, but I suppose that this is not usual in reservoir simulation?

opm/core/props/satfunc/RelpermDiagnostics.cpp
```diff
@@ -385,24 +316,18 @@ namespace Opm{
         const std::string regionIdx = std::to_string(satnumIdx);
         //Check sg column.
         if (sg.front() < 0.0 || sg.back() > 1.0) {
-            const std::string msg = "-- Error:   In SGFN table SATNUM = " + regionIdx + ", saturation should be in range [0,1].";
-            messages_.push_back(msg);
+            const std::string msg = "Error:   In SGFN table SATNUM = " + regionIdx + ", saturation should be in range [0,1].";
```
atgeirr: It seems these three messages retained the Error: (potentially extra) prefix.

atgeirr:
I'll merge this now even though there is a little nitpick about the Error: prefix that was not removed from all places (that can be fixed later). Thanks for the patch!

PR #1012, opened 2016-05-10
Remove the check for MISC

totto82:
Misc controls the transition from miscible to immiscble relperm models.
One table should be specified pr miscibility region (MISCNUM), and not
pr saturation region (SATNUM).

I suggest to remove this test for now as all other tests are based on SATNUM.  What do you think @qilicun?


atgeirr:
This is not currently mergeable, please update if you want it in. Also @qilicun please indicate if you think it should be merged.

atgeirr:
Ping? Please tell if you want this in, @qilicun and @totto82.

atgeirr:
No activity, will close. Reopen if still wanted.

PR #1014, opened 2016-05-10
PinchProcessor no longer does anything with NNC

pgdr:
Removed logic regarding `NNC` in PinchProcessor

* Updated parameters to methods `process` and `transTopbot_`
* Removed testing of NNC behavior in `test_Pinchprocessor`

We should probably add the test in **simulators**.

See also OPM/opm-simulators#680.


atgeirr:
I do not understand what this change wants to do. The only purpose of the PinchProcessor is to build the NNC connections across pinched-out layers, and removing this makes this class do nothing (the only way it changes the state of the program is through the mutable reference nnc argument)!

pgdr:
This PR should have been closed, my apologies.

atgeirr:
No problem!

PR #1015, opened 2016-05-12
Add member hydroCarbonState to the BlackoilState

totto82:
The hydroCarbonState is used to store the hydroCarbonState
  State 1: Gas only
  State 2: Gas and Oil
  State 3: Oil only
An empty vector is return at initialization as
no default values are provided by the blackoilstate.

### DIFF
--- a/opm/core/simulator/BlackoilState.cpp
+++ b/opm/core/simulator/BlackoilState.cpp
@@ -21,10 +21,11 @@ BlackoilState::BlackoilState( size_t num_cells , size_t num_faces , size_t num_p
 }
 
 BlackoilState::BlackoilState( const BlackoilState& other )
-    : SimulationDataContainer(other)
+    : SimulationDataContainer(other),
+      hydrocarbonstate_(other.hydroCarbonState())
 {
     setBlackoilStateReferencePointers();
-    hydrocarbonstate_ = other.hydroCarbonState();
+
 }
 
 BlackoilState& BlackoilState::operator=( const BlackoilState& other )
--- a/opm/core/simulator/BlackoilState.cpp
+++ b/opm/core/simulator/BlackoilState.cpp
@@ -24,12 +24,14 @@ BlackoilState::BlackoilState( const BlackoilState& other )
     : SimulationDataContainer(other)
 {
     setBlackoilStateReferencePointers();
+    hydrocarbonstate_ = other.hydroCarbonState();
 }
 
 BlackoilState& BlackoilState::operator=( const BlackoilState& other )
 {
     SimulationDataContainer::operator=(other);
     setBlackoilStateReferencePointers();
+    hydrocarbonstate_ = other.hydroCarbonState();
     return *this;
 }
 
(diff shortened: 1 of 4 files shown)


atgeirr:
While I am not completely against this, it has to compete with the variables embodying the same information currently stored in the model class itself. I would be happy to replace all of them with something like this, but adding one more I am very reluctant to do!

opm/core/simulator/BlackoilState.cpp
```diff
@@ -24,12 +24,14 @@ BlackoilState::BlackoilState( const BlackoilState& other )
     : SimulationDataContainer(other)
 {
     setBlackoilStateReferencePointers();
+    hydrocarbonstate_ = other.hydroCarbonState();
```
atgeirr: Use the initializer list to initialize members in constructors:
```
 BlackoilState::BlackoilState( const BlackoilState& other )
      : SimulationDataContainer(other),
        hydrocarbonstate_(other.hydrocarbonstate_)
{
...
```

opm/core/simulator/BlackoilState.hpp
```diff
@@ -30,6 +30,12 @@
 namespace Opm
 {
 
+enum HydroCarbonState {
```
atgeirr: Indent one more level (to the same level as the class).

opm/core/simulator/BlackoilState.hpp
```diff
@@ -62,17 +68,22 @@ namespace Opm
         std::vector<double>& surfacevol  () { return *surfacevol_ref_;  }
         std::vector<double>& gasoilratio () { return *gasoilratio_ref_; }
         std::vector<double>& rv ()          { return *rv_ref_;          }
+        std::vector<int>& hydroCarbonState() { return hydrocarbonstate_;  }
```
atgeirr: Use std::vector<HydroCarbonState> instead of std::vector<int>.

opm/core/utility/initHydroCarbonState.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Opm
+{
+
+void initHydroCarbonState(BlackoilState& state, const PhaseUsage& pu, const int num_cells) {
+    enum { Oil = BlackoilPhases::Liquid, Gas = BlackoilPhases::Vapour, Water = BlackoilPhases::Aqua };
+    // hydrocarbonstate is only used when gas and oil is present
+    assert(pu.phase_used[Oil]);
+    if (!pu.phase_used[Gas]) {
+        return; // do nothing
```
atgeirr: Should it not be initialized to OilOnly?

atgeirr:
No more comments here, will be ready to merge (along with downstream PR) after update and build test.

opm/core/utility/initHydroCarbonState.hpp
totto82: Why do you think so?

atgeirr: Otherwise, the hydrocarbonstate will be invalid (empty), and code might fail further down the line.

totto82: Ok. No I see your point. You mean the case without gas.

atgeirr: Yes, exactly.

PR #1019, opened 2016-05-20
Write more messages into OpmLog.

qilicun:
summary:
- Add enum to string functions for injection and production specification.
- log well-related information 
- log information in parameter group
- some other information.

### DIFF
--- a/opm/core/props/rock/RockCompressibility.cpp
+++ b/opm/core/props/rock/RockCompressibility.cpp
@@ -22,6 +22,7 @@
 #include <opm/core/utility/parameters/ParameterGroup.hpp>
 #include <opm/core/utility/Units.hpp>
 #include <opm/common/ErrorMacros.hpp>
+#include <opm/common/OpmLog/OpmLog.hpp>
 #include <opm/core/utility/linearInterpolation.hpp>
 
 #include <opm/parser/eclipse/EclipseState/Tables/RocktabTable.hpp>
@@ -64,14 +65,16 @@ namespace Opm
             if (rockKeyword.size() != 1) {
                 // here it would be better not to use std::cout directly but to add the
                 // warning to some "warning list"...
-                std::cout << "Can only handle a single region in ROCK ("<<rockKeyword.size()<<" regions specified)."
-                          << " Ignoring all except for the first.\n";
+                OpmLog::warning("Can only handle a single region in ROCK ("
+                                + std::to_string(rockKeyword.size())
+                                + " regions specified)."
+                                + " Ignoring all except for the first.");
             }
 
             pref_ = rockKeyword.getRecord(0).getItem("PREF").getSIDouble(0);
             rock_comp_ = rockKeyword.getRecord(0).getItem("COMPRESSIBILITY").getSIDouble(0);
         } else {
-            std::cout << "**** warning: no rock compressibility data found in deck (ROCK or ROCKTAB)." << std::endl;
+            OpmLog::warning("**** warning: no rock compressibility data found in deck (ROCK or ROCKTAB).");
         }
     }
 
--- a/opm/core/props/rock/RockCompressibility.cpp
+++ b/opm/core/props/rock/RockCompressibility.cpp
@@ -74,7 +74,7 @@ namespace Opm
             pref_ = rockKeyword.getRecord(0).getItem("PREF").getSIDouble(0);
             rock_comp_ = rockKeyword.getRecord(0).getItem("COMPRESSIBILITY").getSIDouble(0);
         } else {
-            OpmLog::warning("**** warning: no rock compressibility data found in deck (ROCK or ROCKTAB).");
+            OpmLog::warning("No rock compressibility data found in deck (ROCK or ROCKTAB).");
         }
     }
 
--- a/opm/core/utility/parameters/ParameterGroup_impl.hpp
+++ b/opm/core/utility/parameters/ParameterGroup_impl.hpp
@@ -45,6 +45,7 @@
 #include <opm/core/utility/parameters/ParameterTools.hpp>
 #include <opm/core/utility/parameters/Parameter.hpp>
 #include <opm/common/ErrorMacros.hpp>
+#include <opm/common/OpmLog/OpmLog.hpp>
 
 namespace Opm {
     namespace parameter {
@@ -143,9 +144,9 @@ namespace Opm {
 		    assignments.push_back(std::make_pair(name, value));
 		    continue;
 		}
-		std::cout << "WARNING: Too many assignements  (' "
-                          << ID_delimiter_assignment
-                          << "') detected in argument " << i << ".\n";
+		OpmLog::warning("WARNING: Too many assignements  (' "
+                        + ID_delimiter_assignment
+                        + "') detected in argument " + to_string(i) +);
 	    }
 	    for (int i = 0; i < int(files.size()); ++i) {
 		std::pair<std::string, std::string> file_type = filename_split(files[i]);
@@ -186,12 +187,7 @@ namespace Opm {
 		if (parent_ != 0) {
 		    // If we have a parent, ask it instead.
 		    if (output_is_enabled_) {
-			//TermColors::Red();
-			std::cout << name;
-			//TermColors::Normal();
-			std::cout << " not found at "
-                                  << (path() + ID_delimiter_path)
-                                  << ", asking parent." << std::endl;
+                OpmLog::error(name + path() + ID_delimiter_path + ", asking parent.");
 		    }
 		    return parent_->get<T>(name, r);
 		} else {
@@ -208,11 +204,7 @@ namespace Opm {
 		T val = this->translate<T>(*it, r);
 		it->second->setUsed();
 		if (output_is_enabled_) {
-		    //TermColors::Green();
-		    std::cout << name;
-		    //TermColors::Normal();
-		    std::cout << " found at " << (path() + ID_delimiter_path)
-			      << ", value is " << to_string(val) << std::endl;
+            OpmLog::info(name + " found at " + path() + ID_delimiter_path + ", value is " + to_string(val));
 		}
 		return val;
 	    } else {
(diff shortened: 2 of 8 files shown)


opm/core/utility/parameters/ParameterGroup_impl.hpp
```diff
@@ -186,12 +187,7 @@ namespace Opm {
 		if (parent_ != 0) {
 		    // If we have a parent, ask it instead.
 		    if (output_is_enabled_) {
-			//TermColors::Red();
-			std::cout << name;
-			//TermColors::Normal();
-			std::cout << " not found at "
-                                  << (path() + ID_delimiter_path)
-                                  << ", asking parent." << std::endl;
+                OpmLog::error(name + path() + ID_delimiter_path + ", asking parent.");
```
pgdr: some strange indentation?

atgeirr: Also: this is not an error.

pgdr: And the message `not found at` is missing?  It seems like that information is interesting.

opm/core/utility/parameters/ParameterGroup_impl.hpp
```diff
@@ -208,11 +204,7 @@ namespace Opm {
 		T val = this->translate<T>(*it, r);
 		it->second->setUsed();
 		if (output_is_enabled_) {
-		    //TermColors::Green();
-		    std::cout << name;
-		    //TermColors::Normal();
-		    std::cout << " found at " << (path() + ID_delimiter_path)
-			      << ", value is " << to_string(val) << std::endl;
+            OpmLog::info(name + " found at " + path() + ID_delimiter_path + ", value is " + to_string(val));
```
pgdr: Indentation again?  Please correct all indentation issues.

opm/core/props/rock/RockCompressibility.cpp
```diff
@@ -64,14 +65,16 @@ namespace Opm
...
+                                + std::to_string(rockKeyword.size())
+                                + " regions specified)."
+                                + " Ignoring all except for the first.");
             }
 
             pref_ = rockKeyword.getRecord(0).getItem("PREF").getSIDouble(0);
             rock_comp_ = rockKeyword.getRecord(0).getItem("COMPRESSIBILITY").getSIDouble(0);
         } else {
-            std::cout << "**** warning: no rock compressibility data found in deck (ROCK or ROCKTAB)." << std::endl;
+            OpmLog::warning("**** warning: no rock compressibility data found in deck (ROCK or ROCKTAB).");
```
atgeirr: Here the ad-hoc prefix should be removed as well.

opm/core/utility/parameters/ParameterGroup_impl.hpp
```diff
@@ -265,23 +253,16 @@ namespace Opm {
 		    }
 		}
 		if (output_is_enabled_) {
-		    //TermColors::Blue();
-		    std::cout << name;
-		    //TermColors::Normal();
-		    std::cout << " not found. Using default value '"
-                              << to_string(default_value) << "'." << std::endl;
+            OpmLog::info(name + " not found. Using default value '" + to_string(default_value) + ".");
```
pgdr: missing `'` after `default_value` (before period).

opm/core/utility/parameters/ParameterGroup_impl.hpp
```diff
@@ -143,9 +144,9 @@ namespace Opm {
 		    assignments.push_back(std::make_pair(name, value));
 		    continue;
 		}
-		std::cout << "WARNING: Too many assignements  (' "
-                          << ID_delimiter_assignment
-                          << "') detected in argument " << i << ".\n";
+		OpmLog::warning("WARNING: Too many assignements  (' "
```
atgeirr: Remove ad-hoc prefix. Is it not lovely to get rid of all these slightly-different ways to mark something as a warning?

opm/core/wells/InjectionSpecification.hpp
```diff
@@ -25,7 +26,9 @@ namespace Opm
         };
 
         InjectionSpecification();
-
+        const std::string ControlMode2String(const ControlMode& mode);
+        const std::string InjectorType2String(const InjectorType& type);
+        const std::string GuideRateType2String(const GuideRateType& type);
```
pgdr: Please consider using `toString` instead of `2String`, and also consider overloading on input, and have method names `toString()` only.

atgeirr:
I have a general concern with this: this forces all programs that use classes that now use the logging mechanism to also use the logging mechanism. In itself, that is a good thing, and I will not let that prevent this from being merged. It does however give us the problem of quickly and easily adding logging to many programs.

I propose that we make a static method in the OpmLog class, for example OpmLog::setupSimpleDefaultLogging() that creates a StreamLog backend on cout (with prefixes and perhaps coloring) and adds it. Then adding minimal logging support to program not requiring sophisticated support is a simple one-liner. @qilicun if you agree, can you do this?

opm/core/wells/InjectionSpecification.hpp
jokva: My favourite is `to_string` since you already pass different types, and let overloading sort it out.

opm/core/wells/ProductionSpecification.cpp
```diff
@@ -19,4 +19,98 @@ namespace Opm
...
+
+    const std::string
+    ProductionSpecification::ControlMode2String(const ControlMode& mode)
+    {
+        std::string stringValue;
+
+        switch(mode) {
+        case ControlMode::NONE:
+            stringValue = "NONE";
+            break;
```
jokva: Do yourself a favour and just `return` here, no need to assign to a temporary.

`switch( mode ) {
    case ControlMode::NONE: return "NONE";
    case ControlMode::ORAT: return "ORAT";
   [...]
   }
`

opm/core/wells/InjectionSpecification.hpp
atgeirr: I agree, use toString() rather than ControlMode2String(). However to_string is a bit more tricky since it is a standard thing, then you should always do `using std::to_string` in contexts where you want to use it (since using the namespace explicitly as in `std::to_string()` will not see your overload). If you do that however, the code can be more generic. Your call.

jokva: I'm not suggesting extending the standard namespace (which is illegal), just a familiar naming in the opm namespace. It's familiar and is probably a good idea of all `to_string` like functions. And as you say, it makes it easer to write generic code.

And if it discourages `using namespace std` all the better, because that is an abomination and I'm pretty sure it was described by  both Dante and Brad Pitt.

atgeirr: I was not suggesting to add it to std, but to get agrument-dependent lookup you cannot qualify the name (i.e. `to_string(foo)` will work in a generic context, for example a different namespace, if the class of foo (call it Foo) and to_string(Foo) are both in namespace Opm, but `std::to_string(foo)` will fail. To use the former you must do `using std::to_string` to avoid failing when foo is an integer for example.

jokva: Right, sure, in the case you write code that might both want to use std::to_string and Opm::to_string (unlikely, but possible). If you want `std::to_string` you can also specify it with `std::` to skip ADL altogether.

bska: Also, `const std::string` is meaningless as a return type.  Drop the `const` or make the return type into a reference.

qilicun:
>if you agree, can you do this?

Yes.

>OpmLog::setupSimpleDefaultLogging() that creates a StreamLog backend on cout (with prefixes and perhaps coloring) and adds it.

Will this default logging be used to control output for Flow or just for relperm diagnostics?

opm/core/utility/parameters/ParameterGroup_impl.hpp
qilicun: I thought this due to the incompatible of vi and Emacs..

atgeirr: The indentation trouble I believe is due to this file containing some tabs that have not yet been turned into spaces. If you change the tabs to spaces in this file that would improve things (and in github we can add `?w=1` to the URL to see without whitespace changes).

pgdr: I think vi and emacs are perfectly compatible.  It seems that line 206 has leading tabs, whereas line 207 has leading spaces.  We prefer spaces in this project, you can have a separate commit where you fix all tab-to-space fixes.

atgeirr:
> Will this default logging be used to control output for Flow or just for relperm diagnostics?

I'll just add it to all the programs that do *not* control logging more explicitly. Flow (and the standalone diagnostics program) need more detailed control and should not use the default setup.

qilicun:
>I'll just add it to all the programs that do not control logging more explicitly.

When https://github.com/OPM/opm-common/pull/122 is merged, I will I can update that for flow and relperm diagnostics.

atgeirr:
> I can update that for flow and relperm diagnostics.

Maybe we misunderstood each other: Flow should *not* use OpmLog::setupSimpleDefaultLogging(), it is just for all the other programs in opm-simulators etc. to make it easy to add minimal logging support.

qilicun:
>make it easy to add minimal logging support.

That's fine.

opm/core/wells/InjectionSpecification.hpp
```diff
@@ -25,7 +26,9 @@ namespace Opm
         };
 
         InjectionSpecification();
-
+        std::string toString(const ControlMode& mode);
```
atgeirr: I believe these methods should be static (you should also change the callsites from `injSpec().toString(...)` to InjectorSpecification::toString(...)` to make it clear that it does not depend on the actual injSpec() object). Similar for production. Sorry for not noticing this the first time!

atgeirr:
This looks very good now, with just a tiny issue (`static`) remaining. I'll merge when that is fixed.

atgeirr:
Thanks for your effort, I will merge. Just a note: if you update a PR with new commits, we do not get automatic email about it, so it is often a good idea to add a short note that the PR has been updated. Otherwise days can go by without the change being noticed.

Issue #1020, opened 2016-05-27
opm-core build fails


pgdr:
opm-core build fails.  I don't know what's the reason, nor why Travis is green.  But Jenkins fails and I can't build it either.

    /var/jenkins/workspace/opm-core-install-RH6/examples/compute_initial_state.cpp: In function ‘void {anonymous}::outputData(const string&, const string&, const std::vector<double>&)’:
    /var/jenkins/workspace/opm-core-install-RH6/examples/compute_initial_state.cpp:64:28: error: variable ‘std::ofstream file’ has initializer but incomplete type
             std::ofstream file(fname.str().c_str());
                                ^
    Linking CXX executable bin/tutorial4
    /var/jenkins/workspace/opm-core-install-RH6/examples/compute_tof_from_files.cpp: In function ‘int main(int, char**)’:
    /var/jenkins/workspace/opm-core-install-RH6/examples/compute_tof_from_files.cpp:89:35: error: variable ‘std::ifstream poro_stream’ has initializer but incomplete type
             std::ifstream poro_stream(param.get<std::string>("poro_filename").c_str());

blattms:
I experienced the same fixed with PR #1021 here and OPM/opm-simulators#699 in simulators.

blattms:
travis does not find ERT. No idea why.

Are you sure you are using my branch? I definitely fixed that file:

```
mblatt@loft11004:~/DUNE-test/opm-core$ grep fstream
examples/compute_initial_state.cpp
#include <fstream>
std::ofstream file(fname.str().c_str());
blatt@loft11004:~/DUNE-test/opm-core$ git diff
blatt@loft11004:~/DUNE-test/opm-core$
```
and pushed the changes.

jokva:
There's a PR on it in common. The ERT C++ interface was changed, and the the one changed bit was the one checked by FindERT.cmake in common. Apply https://github.com/OPM/opm-common/pull/125 and it should be fine.

pgdr:
100% tests passed, 0 tests failed out of 38

PR #1021, opened 2016-05-27
Fix incomplete type error for std::ofstream in examples

blattms:
On my system I got
```c++
error: variable ‘std::ofstream file’ has initializer but incomplete type
         std::ofstream file(fname.str().c_str());
```
This is fixed with this commit by including fstream. Previously, this include might have happened implicitely.

### DIFF
--- a/examples/compute_initial_state.cpp
+++ b/examples/compute_initial_state.cpp
@@ -36,6 +36,8 @@
 
 #include <boost/filesystem.hpp>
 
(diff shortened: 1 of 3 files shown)


blattms:
Closes #1020

blattms:
Check fails due to travis not finding ERT.

atgeirr:
I'll merge this since the test failure is unrelated.

PR #1023, opened 2016-06-06
paramemters output should go in Debug mode.

qilicun:
These parameters notification actually are not very useful for user, we should let these information go into Debug.

output

### DIFF
--- a/opm/core/utility/parameters/ParameterGroup_impl.hpp
+++ b/opm/core/utility/parameters/ParameterGroup_impl.hpp
@@ -204,7 +204,7 @@ namespace Opm {
 		T val = this->translate<T>(*it, r);
 		it->second->setUsed();
 		if (output_is_enabled_) {
-		    OpmLog::info(name + " found at " + path() + ID_delimiter_path + ", value is " + to_string(val));
+		    OpmLog::debug(name + " found at " + path() + ID_delimiter_path + ", value is " + to_string(val));
 		}
 		return val;
 	    } else {
@@ -253,7 +253,7 @@ namespace Opm {
 		    }
 		}
 		if (output_is_enabled_) {
-            	    OpmLog::info(name + " not found. Using default value '" + to_string(default_value) + "'.");
+            	    OpmLog::debug(name + " not found. Using default value '" + to_string(default_value) + "'.");
 		}
 		return default_value;
 	    }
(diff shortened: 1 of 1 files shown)


atgeirr:
Actually, for these messages there already exists a mechanism for controlling the messages: `ParameterGroup` has the following members:
```
	    /// \brief Disables the output from get, getDefault and getGroup.
	    ///        By default, such output is enabled.
	    void disableOutput();

	    /// \brief Enables the output from get, getDefault and getGroup.
	    ///        By default, such output is enabled.
	    void enableOutput();

	    /// \brief Returs true if and only if output from get, getDefault
	    ///        and getGroup is enabled.
	    ///
	    /// \return true if and only if output from get, getDefault and
	    ///         getGroup is enabled.
	    bool isOutputEnabled() const;
```
After thinking through this a little though, I agree with moving to OpmLog::debug(), just thought it was good to be aware of the other options!

blattms:
How do I get these messages back?
Actually, I do not share the view that the parameters are debug options as these settings are needed to reproduce a run. At least if non-default settings were used.

atgeirr:
> Actually, I do not share the view that the parameters are debug options as these settings are needed to reproduce a run

Any non-default parameters are written to the simulation.param file. Defaulted parameters are not however.

By default the debug() messages are written to the new ".CASENAME.DEBUG" file (hidden for not confusing regular users), see OPM/opm-simulators#705.

blattms:
Hopefully, we will never need to replicate runs of regular user... ;)

Sorry, but these message about the parameters do not seem to end up there. `grep default .*.DEBUG` does not reveal any results.

My usual work flow was to pipe all output to a file, that persisted quite some month. Now I have to be careful and backup a hidden file after every run. That is quite a bit more inconvenient.

blattms:
But I recall that I can somehow set a subdirectory for the output, right?`Will that debug file end up there, too?

qilicun:
>Sorry, but these message about the parameters do not seem to end up there. grep default .*.DEBUG does not reveal any results.

I just test norne, ``grep default .*.DEBUG`` works from my side.

>But I recall that I can somehow set a subdirectory for the output, right?`Will that debug file end up there, too?

I think if you set ``output_dir``, the debug file will be there.

blattms:
@qilicun I did not update and compile opm-common. Sorry for the noise.
I guess, I will go with output_dir in the future:

qilicun:
@blattms 
Fine!

PR #1024, opened 2016-06-06
rename some variables from "Lad" to "Eval" (was: to "Dad")

andlaus:
this was an involuntary omission in the "local AD" to "dense AD" rename...

I'm open to change this to a better suffix, so feel free to suggest one in the comments. It should probably be noted that this PR is self contained, but for consistency reasons it would be beneficial to merge it together with its opm-simulators counterpart (OPM/opm-simulators#706).

### DIFF
--- a/opm/core/props/BlackoilPropertiesFromDeck.cpp
+++ b/opm/core/props/BlackoilPropertiesFromDeck.cpp
@@ -244,15 +244,15 @@ namespace Opm
         const auto& pu = phaseUsage();
         const int np = numPhases();
 
-        typedef Opm::DenseAd::Evaluation<double, /*size=*/1> LadEval;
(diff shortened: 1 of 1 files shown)


andlaus:
yay! PR#2^10

atgeirr:
I'll make my comments here: on the whole I still find naming less than enlightening, just "Eval" would make the connection with the Evaluation class clearer. The abbreviation Dad (or Lad before) is not obviously an abbreviation, and you don't want to explain it everywhere. (Also, including the type in the variable name is not necessary, but convenient here because `p` etc. are already used by the method parameters.) So: I'd prefer using Eval instead of Dad.

andlaus:
> So: I'd prefer using Eval instead of Dad.

me too, actually. the patches for the two PRs have been adapted.

PR #1026, opened 2016-06-07
Added utilities to compute a real L-infinity norm in parallel.

blattms:


opm/core/linalg/ParallelIstlInformation.hpp
```diff
@@ -574,6 +575,30 @@ class ParallelISTLInformation
             (std::pointer_to_binary_function<const T&,const T&,const T&>
              ((const T&(*)(const T&, const T&))std::max<T>));
     }
+
+    namespace detail
+    {
+    /// \brief Computes the maximum of theabsolute values of two values.
+    template<typename T>
+    struct MaxAbsFunctor
+    {
```
bska: Lately I've taken a liking to C++14's generic `plus<void>` (and similar) object.  Rather than parameterising the structure it parameterises the operation so we get type deduction and support for mixed arithmetic.
```C++
struct MaxAbsFunctor
{
    using std::max;
    using std::abs;

    template <typename T1, typename T2>
    auto operator(T1&& x, T2&& y)
        -> max(abs(std::forward<T1>(x)), abs(std::forward<T2>(y))
    {
        return max(abs(std::forward<T1>(x)), abs(std::forward<T2>(y)));
    }
};
```
I have not considered all implications of using such an approach in this case.

blattms:
*Feels so old school* :(
Even so your code is really fancy, I do not see any containers that are using mixed types. At least not in OPM. And I doubt that this functor is of more general use.

But as this is open source feel free to change. Your approach should work, too.
For me typing the same line twice for auto type reduction is too much here.

jokva:
It's more about getting type deduction when invoking the operation itself, rather than having to specify the type manually. While the definition has massive syntactic overhead, on call sites it's the difference between

`MaxAbsFunctor< double >( x, y )` and `MaxAbsFunctor( x, y )`.

bska:
> It's more about getting type deduction when invoking the operation itself, rather than having to specify the type manually.

True, that's the main difference.  Explicit template arguments are usually best avoided when the types can be deduced--at least if you do not unequivocally wish to incur a type conversion.

As for the syntactic overhead of the trailing return type, that's "only" needed if we wish to remain compatible with C++11.  In a C++14 context I'd use `decltype(auto)` instead.

atgeirr:
> As for the syntactic overhead of the trailing return type, that's "only" needed if we wish to remain compatible with C++11. In a C++14 context I'd use decltype(auto) instead.

I agree decltype(auto) is nice, the C++11 alternative is not very appealing at all I think. I'll merge this as is.

atgeirr:
I get errors from test_parallelistlinformation.cpp after this:
```
/Users/atgeirr/opm-build-debug/opm-core/tests/test_parallelistlinformation.cpp:63:47: error: call to 'abs' is ambiguous
    BOOST_CHECK(std::get<4>(values)==std::max(std::abs(offset),std::abs(N+offset-1)));
                                              ^~~~~~~~
.
.
.

/Users/atgeirr/opm-build-debug/opm-core/opm/core/linalg/ParallelIstlInformation.hpp:589:29: error: call to 'abs' is ambiguous
            return std::max(std::abs(t1),std::abs(t2));
                            ^~~~~~~~

```
There exists no overload of abs() for unsigned integral types. Doing a reduction using abs on such a type also does not make much sense. So I propose we disable that check for unsigned types in some way. Agree?

atgeirr:
Another (better) way to fix this: add a specialization to avoid this. I'll push a PR.

PR #1027, opened 2016-06-08
Fix two phase equil initialization

totto82:
- add guards against non-existing phase
- shift the capillary pressure values with the oil capillary pressure value to handle cases where oil is not the reference case in opm-material. i.e oil-water problems.   

Tested on the Brugge case. Tested that it does not effect norne, Model 2 and SPE9 initialization. 

Not tested on oil-gas problems.

### DIFF
--- a/opm/core/props/satfunc/SaturationPropsFromDeck.cpp
+++ b/opm/core/props/satfunc/SaturationPropsFromDeck.cpp
@@ -155,11 +155,12 @@ namespace Opm
                 // copy the values calculated using opm-material to the target arrays
                 for (int canonicalPhaseIdx = 0; canonicalPhaseIdx < BlackoilPhases::MaxNumPhases; ++canonicalPhaseIdx) {
                     // skip unused phases
-                    if ( ! phaseUsage_.phase_used[canonicalPhaseIdx])
+                    if ( ! phaseUsage_.phase_used[canonicalPhaseIdx]) {
                         continue;
+                    }
                     const int pcPhaseIdx = phaseUsage_.phase_pos[canonicalPhaseIdx];
 
-                    double sign = (canonicalPhaseIdx == BlackoilPhases::Aqua)? -1.0 : 1.0;
+                    const double sign = (canonicalPhaseIdx == BlackoilPhases::Aqua)? -1.0 : 1.0;
                     // in opm-material the wetting phase is the reference phase
                     // for two-phase problems i.e water for oil-water system,
                     // but for flow it is always oil. Add oil (liquid) capillary pressure value
--- a/opm/core/props/satfunc/SaturationPropsFromDeck.cpp
+++ b/opm/core/props/satfunc/SaturationPropsFromDeck.cpp
@@ -139,6 +139,7 @@ namespace Opm
         assert(cells != 0);
 
         const int np = numPhases();
+
         if (dpcds) {
             ExplicitArraysSatDerivativesFluidState fluidState(phaseUsage_);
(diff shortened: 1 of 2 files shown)


opm/core/props/satfunc/SaturationPropsFromDeck.cpp
```diff
@@ -153,10 +154,13 @@ namespace Opm
                 // copy the values calculated using opm-material to the target arrays
                 for (int pcPhaseIdx = 0; pcPhaseIdx < np; ++pcPhaseIdx) {
```
atgeirr: In this loop (and the same applies to the loop on line 172 [original line numbers]) there are two bugs:
 - we should assert() that both water and oil are active phases (since we use the capillary pressures for those two phases) -- this was already wrong, but should be fixed now
 - since pcPhaseIdx and satPhaseIdx ranges over *active* phases (0, ..., np-1) while `capillaryPressures` are indexed by canonical phase indices (BlackoilPhases::Liquid etc. == 0, 1, 2) we must take great care not to mix them.  Instead of using `capillaryPressures[pcPhaseIdx]` for example one should use `capillaryPressures[canonicalIdx]` where `canonicalIdx = actToCan[pcPhaseIdx]` and actToCan is as computed for example by detail::active2Canonical in BlackoilModelBase_impl.hpp.

atgeirr:
I'm very happy to see that this gets some attention. I believe the code has some bugs still as commented above, they would not trigger on a water-oil case simply because then the active and canonical phase indices coincide.

atgeirr:
I'd like to see this fixed up so I can merge it, I can be of assistance if necessary.

totto82:
I will take a look today.

opm/core/props/satfunc/SaturationPropsFromDeck.cpp
```diff
@@ -151,12 +153,25 @@ namespace Opm
...
                 // copy the values calculated using opm-material to the target arrays
-                for (int pcPhaseIdx = 0; pcPhaseIdx < np; ++pcPhaseIdx) {
-                    double sign = (pcPhaseIdx == BlackoilPhases::Aqua)? -1.0 : 1.0;
-                    pc[np*i + pcPhaseIdx] = sign*capillaryPressures[pcPhaseIdx].value;
-
-                    for (int satPhaseIdx = 0; satPhaseIdx < np; ++satPhaseIdx)
-                        dpcds[np*np*i + satPhaseIdx*np + pcPhaseIdx] = sign*capillaryPressures[pcPhaseIdx].derivatives[satPhaseIdx];
+                for (int canonicalPhaseIdx = 0; canonicalPhaseIdx < BlackoilPhases::MaxNumPhases; ++canonicalPhaseIdx) {
+                    // skip unused phases
+                    if ( ! phaseUsage_.phase_used[canonicalPhaseIdx])
```
atgeirr: Use braces (preferred) or put the continue on the same line.

opm/core/props/satfunc/SaturationPropsFromDeck.cpp
```diff
@@ -151,12 +153,25 @@ namespace Opm
...
+                        continue;
+                    const int pcPhaseIdx = phaseUsage_.phase_pos[canonicalPhaseIdx];
+
+                    double sign = (canonicalPhaseIdx == BlackoilPhases::Aqua)? -1.0 : 1.0;
```
atgeirr: Should be const double.

atgeirr:
Looks good now! A few minor nitpicks aside, this is ready to be merged. I assume you have re-run some of the initialization tests after the update?

PR #1028, opened 2016-06-08
Format message

qilicun:
- Output file name and line number when the deck has more than 1 region of the values.
- Add a blank line before ``***************Saturation Functions Diagnostics***************``


atgeirr:
Good, lots of small improvements make for a much better experience over time!

PR #1034, opened 2016-06-13
Minor format changes

qilicun:
For a message which is more than one line, it is better to add "\n" at the end of the message.


qilicun:
The error is 
``unknown location(0): fatal error in "DefaultTest": std::runtime_error: [/home/travis/build/OPM/opm-core/opm/core/linalg/LinearSolverFactory.cpp:90] Linear solver umfpack is not enabled in this configuration.
``
Is there something wrong with travis?

jokva:
There's something wrong with the build system rather.

https://github.com/OPM/opm-core/pull/1017

PR #1038, opened 2016-06-15
Rise up a warning if equil region has no active cells.

qilicun:
Eclipse output this equilibration check.

### DIFF
--- a/opm/core/simulator/initStateEquil.hpp
+++ b/opm/core/simulator/initStateEquil.hpp
@@ -394,7 +394,7 @@ namespace Opm
                         const auto& cells = reg.cells(r);
                         if (cells.empty())
                         {
-                            OpmLog::warning("Equilibration region " + std::to_string(r) 
+                            OpmLog::warning("Equilibration region " + std::to_string(r + 1) 
                                             + " has no active cells");
                             continue;
                         }
--- a/opm/core/simulator/initStateEquil.hpp
+++ b/opm/core/simulator/initStateEquil.hpp
@@ -390,14 +390,13 @@ namespace Opm
                                  const Grid&                       G    ,
                                  const double grav)
                 {
-                    int regnum = 0;
                     for (const auto& r : reg.activeRegions()) {
-                        regnum++ ;
                         const auto& cells = reg.cells(r);
                         if (cells.empty())
                         {
-                            OpmLog::warning("Equilibration region " + std::to_string(regnum) 
+                            OpmLog::warning("Equilibration region " + std::to_string(r) 
                                             + " has no active cells");
+                            continue;
                         }
                         const int repcell = *cells.begin();
 
--- a/opm/core/simulator/initStateEquil.hpp
+++ b/opm/core/simulator/initStateEquil.hpp
@@ -37,6 +37,7 @@
 #include <opm/parser/eclipse/EclipseState/Tables/TableManager.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/RsvdTable.hpp>
 #include <opm/parser/eclipse/EclipseState/Tables/RvvdTable.hpp>
+#include <opm/common/OpmLog/OpmLog.hpp>
 
 #include <array>
 #include <cassert>
@@ -389,11 +390,14 @@ namespace Opm
                                  const Grid&                       G    ,
                                  const double grav)
                 {
+                    int regnum = 0;
                     for (const auto& r : reg.activeRegions()) {
+                        regnum++ ;
                         const auto& cells = reg.cells(r);
                         if (cells.empty())
                         {
-                            continue;
+                            OpmLog::warning("Equilibration region " + std::to_string(regnum) 
+                                            + " has no active cells");
                         }
                         const int repcell = *cells.begin();
 


atgeirr:
Are you sure this will be correct? I seem to remember that the set of equilibration regions can have arbitrary numbering, not necessarily 1...n. I may be confusing it with some other region though.

qilicun:
Fro equilibration, regions is defined by keyword ``EQUIL`` that said number of the records of this keyword is the number of regions to my understanding.

qilicun:
>Are you sure this will be correct? I seem to remember that the set of equilibration regions can have arbitrary numbering, not necessarily 1...n. I may be confusing it with some other region though.

I'm not sure this. Probably we should get the regions from ``EQLNUM``? that will be correct.

qilicun:
@atgeirr @alfbr 

I test with some situations, in any case, this message will never have the chance to be output. 
If the number of regions from EQLDIM mismatch that from EQUIL and EQLDIM, either parser will throw or segment fault will come out.

opm/core/simulator/initStateEquil.hpp
```diff
@@ -389,11 +390,14 @@ namespace Opm
                                  const Grid&                       G    ,
                                  const double grav)
                 {
+                    int regnum = 0;
                     for (const auto& r : reg.activeRegions()) {
+                        regnum++ ;
```
bska: This is wrong.  The quantity `r` already contains the index (obtained from `EQLNUM`) of the current equilibration region.

opm/core/simulator/initStateEquil.hpp
```diff
@@ -389,11 +390,14 @@ namespace Opm
...
                         const auto& cells = reg.cells(r);
                         if (cells.empty())
                         {
-                            continue;
+                            OpmLog::warning("Equilibration region " + std::to_string(regnum) 
+                                            + " has no active cells");
```
bska: You need to keep the `continue` statement here.  Otherwise you'll be dereferencing an invalid iterator on line 402 in the case of an empty region.

opm/core/simulator/initStateEquil.hpp
qilicun: Tested, you are right

opm/core/simulator/initStateEquil.hpp
qilicun: adding ``continue`` back, but when I replace region 1 with 8 in NORNE, it has a segmentation fault
```c
Program received signal SIGSEGV, Segmentation fault.
0x0000000000856ef4 in void Opm::EQUIL::DeckDependent::InitialStateComputer::calcPressSatRsRv<Opm::RegionMapping<std::vector<int, std::allocator<int> > >, UnstructuredGrid>(Opm::RegionMapping<std::vector<int, std::allocator<int> > > const&, std::vector<Opm::EquilRecord, std::allocator<Opm::EquilRecord> > const&, Opm::BlackoilPropertiesInterface&, UnstructuredGrid const&, double) ()
```

bska: > adding continue back, but when I replace region 1 with 8 in NORNE, it has a segmentation fault

Sorry, I don't understand.  What does "replace region 1 with 8" mean?  Doing a global replace of the number `1` with the number `8` in the `EQLNUM` vector?  If so, does the `EQUIL` data have at least eight records?

qilicun: >I seem to remember that the set of equilibration regions can have arbitrary numbering, not necessarily 1...n.

I'm testing this statement. Replace number 1 with 8 in ``EQLNUM`` input file, ``EQUIL`` still contains 5 recordes.

bska: >> seem to remember that the set of equilibration regions can have arbitrary numbering, not necessarily 1...n.
>
> I'm testing this statement. Replace number 1 with 8 in `EQLNUM` input file, `EQUIL` still contains 5 records.

That's bound to fail.  The maximum value of `EQLNUM` over the active cells must not exceed the number of records in `EQUIL` and the minimum value over the same range must be at least one.  Basically, the following expression must be well defined
```fortran
EQUIL(EQLNUM(I), :)
```

qilicun: Thanks!  
I think this PR is ready to merge.

atgeirr:
@bska is this correct now?

alfbr:
I think we need to be more verbose here. To follow the format used by similar messages, please have the warning message explain the error, and then follow up with a line providing keyword, file and line number. I believe we also need a parsecontext for this one to cater to the NOSIM option, so the message may belong in the parser. I am a bit uncertain of how the initialisation code should handle it (if simulation is attempted, i.e., the NOSIM option is not set). I believe stopping simulation with the message makes sense at this point in time. We can always choose a more forgiving approach to EQLDIM later if users request it.

bska:
> @bska is this correct now?

Looks good.  Merge at will.

atgeirr:
I think stopping the simulation for this is overkill, since there is nothing that will go wrong -- the deck just provides data for a region that is not used.

bska:
> please have the warning message explain the error,

It is not necessarily an error to have an empty equilibration region--i.e., an equilibration region with no active cells in it.  It may point to some inconsistency in the deck but it may also be that the grid analysis ended up deactivating all cells within an equilibration region--e.g., due to low pore-volumes.

qilicun:
I still do not see in what situation this message will be output. Since if the number of regions in ``EQLDIM`` mismatches the number of records in ``EQUIL``, parser will throw any exception. That means Parser will check if they are the same, or according to the number of regions in ``EQLDIM`` to get data in ``EQUIL`` and ``EQLNUM``, for other keywords, like pvt related, parser does the same check. But ECLIPSE accept the mismatch and throw an warning.

atgeirr:
So, to get the behaviour @alfbr wants, changes must be made to the parser, not here, since the parser does not accept empty equil regions (really?).

I'll merge this now, and we can revisit to refine the message after any necessary changes to the parser have been made.

alfbr:
> I think stopping the simulation for this is overkill, since there is nothing that will go wrong -- the deck just provides data for a region that is not used.

Thanks for the clarification. In this case I suggest the message is implemented in parser. A parsecontext is added, so that flow can ignore the error, and that flow does not stop at this error.

PR #1039, opened 2016-06-15
Avoid ambiguous calls to abs with unsigned integers.

atgeirr:
In #1026 a `MaxAbs` reduction was added. It causes ambiguous calls for me, when trying to instantiate `std::abs(x)` with x and unsigned integer type.

The suggested patch avoids this by adding a specialization for `std::size_t` in the reduction code, and by using the operator directly in the test.

### DIFF
--- a/opm/core/linalg/ParallelIstlInformation.hpp
+++ b/opm/core/linalg/ParallelIstlInformation.hpp
@@ -578,18 +578,32 @@ private:
 
     namespace detail
     {
-    /// \brief Computes the maximum of theabsolute values of two values.
-    template<typename T>
-    struct MaxAbsFunctor
-    {
-        typedef T result_type;
-
-        result_type operator()(const T& t1, const T& t2)
+        /// \brief Computes the maximum of the absolute values of two values.
+        template<typename T>
+        struct MaxAbsFunctor
         {
-            return std::max(std::abs(t1),std::abs(t2));
-        }
-    };
+            typedef T result_type;
+
+            result_type operator()(const T& t1, const T& t2)
+            {
+                return std::max(std::abs(t1),std::abs(t2));
+            }
+        };
+
+
+        /// \brief Specialization to avoid ambiguous abs() for unsigned integers.
+        template <>
+        struct MaxAbsFunctor<std::size_t>
+        {
+            typedef std::size_t result_type;
+
+            result_type operator()(const std::size_t& t1, const std::size_t& t2)
+            {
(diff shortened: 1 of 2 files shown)


atgeirr:
Since this is a compile-failure fix: can someone (for example @blattms or @dr-robertk) take a look and possible merge this?

atgeirr:
(otherwise I'll merge it myself tomorrow)

bska:
jenkins build this please

jokva:
I suggest generalising this from `size_t` to all unsigned types.

Untested, but you get the idea.
```
struct maxAbs {
using std::abs;
using std::max;

template< typename T >
typename std::enable_if< std::is_unsigned< T >::value, T > operator()( T x, T y ) {
    return max( x, y );
}
template< typename T >
typename std::enable_if< !std::is_unsigned< T >::value, T > operator()( T x, T y ) {
    return max( abs( x ), abs( y ) );
}
auto operator()( T t, U u ) -> decltype( max( abs( t ), abs( u ) ) ) {
   return max( abs( t ), abs( u ) );
}
};
```

atgeirr:
> I suggest generalising this from size_t to all unsigned types.

Can do that, but I imagine with size_t available all other cases will convert with no problems.

jokva:
You'd think so, but it seems not:

```
#include <iostream>

template< typename T >
struct Foo {
    void operator()( T x ) {
        std::cout << "Foo< T >: " << x << std::endl;
    }    
};

template<>
struct Foo< std::size_t > {
    void operator()( std::size_t x ) {
        std::cout << "Foo< size_t >: " << x << std::endl;
    }    
};

int main() {
    Foo< int >()( 1 ); 
    Foo< long >()( -1 );
    Foo< char >()( 'a' );
    Foo< unsigned char >()( 'b' );

    Foo< std::size_t >()( 2 ); 
    Foo< std::size_t >()( -2 );

    Foo< unsigned int >()( 4 );                                                                                                                                                                                      
    Foo< unsigned long >()( 4 ); 
}
```

prints 
```
Foo< T >: 1
Foo< T >: -1
Foo< T >: a
Foo< T >: b
Foo< size_t >: 2
Foo< size_t >: 18446744073709551614
Foo< T >: 4
Foo< size_t >: 4
```

atgeirr:
Well demonstrated. I will change it tomorrow.

atgeirr:
It was more tricky than expected but the current version should work. The subtleties of enable_if are really nontrivial!

Issue #1036, opened 2016-06-14
Runtime failure in "linsolver" test - umfpack not configured


flikka:
When building via the `build-prereqs.sh`, `clone-opm.sh` and `build-and-test.sh` that lies in the opm-common - I ran in to a situation where the build (or rather the test) failed in opm-core. 

The underlying cause was that I had not installed the suitesparse package. The `HAVE_UMFPACK` and `HAVE_SUITESPARSE_UMFPACK_H` were not set.

I don't know if this should be "fixed" by having an `ifdef` before the test in question (`DefaultTest`), or if it something that should be done in the `LinearSolverFactory`, or something else (or nothing).

Just mentioning this, to hear if anything should be done.

atgeirr:
If you don't have UMFPACK it is unfortunate that it is still the default for the LinearSolverFactory. I think we should fix that.

atgeirr:
Resolved in #1040.

PR #1040, opened 2016-06-15
Set default solver depending on configuration.

atgeirr:
Partially adresses issue #1036. Still possible to fail if you have *no* linear solver available (but that should rather be the build system's job to intercept).

### DIFF
--- a/opm/core/linalg/LinearSolverFactory.cpp
+++ b/opm/core/linalg/LinearSolverFactory.cpp
@@ -68,11 +68,12 @@ namespace Opm
 #elif HAVE_PETSC
         std::string default_solver = "petsc";
 #else
+        std::string default_solver = "no_solver_available";
         OPM_THROW(std::runtime_error, "No linear solver available, you must have UMFPACK , dune-istl or Petsc installed to use LinearSolverFactory.");
 #endif
 
         const std::string ls =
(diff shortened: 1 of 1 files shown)


opm/core/linalg/LinearSolverFactory.cpp
```diff
@@ -61,8 +61,18 @@ namespace Opm
...
+        std::string default_solver = "istl";
+#elif HAVE_PETSC
+        std::string default_solver = "petsc";
+#else
+        OPM_THROW(std::runtime_error, "No linear solver available, you must have UMFPACK , dune-istl or Petsc installed to use LinearSolverFactory.");
+#endif
+
         const std::string ls =
-            param.getDefault<std::string>("linsolver", "umfpack");
+            param.getDefault<std::string>("linsolver", default_solver);
```
bska: Since you're already changing this line, you could remove the explicit template argument too.  The type can now be deduced from the `default_solver`.  By the way, how does this line compile at all if none of the `HAVE_*` symbols are defined?  In that case there is no `default_solver`.

atgeirr:
Updated to address comments.

bska:
jenkins build this please

bska:
Both CI systems are green and I have no further comments.  I'll merge into master.  Thanks a lot!

PR #1042, opened 2016-06-16
add opm-output as a dependency to dune.module

andlaus:
without it, building with dunecontrol fails. (for me.)


atgeirr:
Looks like a simple oversight, thanks for fixing. Merging when test complete.

PR #1043, opened 2016-06-20
Remove warnings (by throwing), added copyright header

pgdr:
Added six `throw`s (that will never be reached), thereby removing warnings

* OPM-core had 6 warnings (control reaches end of non-void function [-Wreturn-type])
* Removed warnings by adding `throw std::domain_error` that will never be reached
* Added copyright headers for `InjectionSpecification.cpp` and `ProductionSpecification.cpp`

### DIFF
--- a/opm/core/wells/InjectionSpecification.cpp
+++ b/opm/core/wells/InjectionSpecification.cpp
@@ -1,5 +1,6 @@
 /*
-  Copyright 2016  The Open Porous Media project.
+  Copyright 2012  Sintef.
+  Copyright 2016  Statoil.
 
   This file is part of the Open Porous Media project (OPM).
 
@@ -18,6 +19,8 @@
 
 #include <stdexcept>
 #include "config.h"
+
+#include <opm/common/ErrorMacros.hpp>
 #include <opm/core/wells/InjectionSpecification.hpp>
 
 namespace Opm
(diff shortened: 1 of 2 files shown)


atgeirr:
Thanks for the warning-squashing and improvement! There are just a few adjustments needed:
 - The copyright headers claim "The Open Porous Media Project" as copyright holder. That is incorrect, primarily because copyright holder for these files are SINTEF (work done in 2012 and later) and Statoil (2016). Second, there is no legal entity called "The Open Porous Media Project" that can own anything. Look at for example opm/core/linalg/ParallelIstlInformation.hpp for an example of a file with multiple rightsholders.
 - Errors should be logged. The easy way to do that is to use the OPM_THROW macro: `OPM_THROW(std::domain_error, "Unknown control mode " << mode << " encountered in injection specification");` I know they should never be reached, but on the day they do anyway, it'll help.

pgdr:
Thanks for the feedback.  Changed the copyright header to now read
```c++
/*
 * Copyright 2012  Sintef.
 * Copyright 2016  Statoil.
 * ...
```
... furthermore, added `OPM_THROW` rather than throwing.

I can squash (and rebase) if requested, but to preserve github links I haven't done that yet.

atgeirr:
> I can squash (and rebase) if requested, but to preserve github links I haven't done that yet

No need to, appreciate the concern for the links... I'll merge!

PR #1044, opened 2016-06-20
fixed: broken array loop in jenkins script

akva2:
python, bash, c++, fortran. sometimes you get lost in the syntax jungle.

### DIFF
--- a/jenkins/build-pr.sh
+++ b/jenkins/build-pr.sh
@@ -26,7 +26,7 @@ then
   OPM_COMMON_REVISION=pull/`echo $ghprbCommentBody | sed -r 's/.*opm-common=([0-9]+).*/\1/g'`/merge
 fi
 
-for upstream in $upstreams
+for upstream in ${upstreams[*]}
 do
   if grep -q "$upstream=" <<< $ghprbCommentBody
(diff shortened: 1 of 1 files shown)


atgeirr:
I'll trust you on this...

jenkins/build-pr.sh
```diff
@@ -26,7 +26,7 @@ then
   OPM_COMMON_REVISION=pull/`echo $ghprbCommentBody | sed -r 's/.*opm-common=([0-9]+).*/\1/g'`/merge
 fi
 
-for upstream in $upstreams
+for upstream in ${upstreams[*]}
```
bska: It's not a problem at present, but I **think** I'd prefer
```sh
for upstream in "${upstreams[@]}"
```
(including double quotes) here.

See difference in
```sh
a=('hell' 'o word')

for w in ${a[*]}; do echo "${w}"; done
for w in "${a[@]}"; do echo "${w}"; done
```

akva2: if we get upstreams with spaces in their name, the scripts will fail in 
any case for various other reasons. but sure, in general..

PR #1045, opened 2016-06-21
Change adaptive time stepping message type to OpmLog::note

qilicun:
- Change adaptive time stepping messages type to OpmLog::note
- Let SimulatorTime return current date time

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -216,10 +216,11 @@ namespace Opm {
                 if( timestep_verbose_ )
                 {
                     std::ostringstream ss;
-                    ss << "well iterations = " << solver.wellIterations()
-		       << ", non-linear iterations = " << solver.nonlinearIterations()
-		       << ", total linear iterations = " << solver.linearIterations() 
-		       << "\n";          
+                    if (solver.wellIterations() != std::numeric_limits<int>::min()) {
+                        ss << "well iterations = " << solver.wellIterations() << ", ";
+                    }
+		    ss << "non-linear iterations = " << solver.nonlinearIterations()
+		       << ", total linear iterations = " << solver.linearIterations();
                     OpmLog::info(ss.str());
                 }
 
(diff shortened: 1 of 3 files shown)


atgeirr:
This will remove these messages from the default terminal output, leaving (by default) no way to see the substep behavior other than through the PRT and debug files. Since this output is not very large, I don't see why this is needed?

The addition of currentDateTime() is good.

qilicun:
> Since this output is not very large, I don't see why this is needed?

Let these messages go to PRT and DEBUG files is the discussion with @alfbr, I think he can comment on this, and also for PR https://github.com/OPM/opm-simulators/pull/734

atgeirr:
I guess underlying my question is another one: what output do we want to keep on the terminal in the long run?

alfbr:
I am happy with this one once https://github.com/OPM/opm-simulators/pull/734 is ready.

totto82:
If you run with solve_welleq_initially=false, well iterations returns a large random number. It should be 0. Or maybe it is better not to report the number well iterations in any case. It can give the wrong impression that the wells are solved de-coupled from the rest of the system.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -215,9 +215,11 @@ namespace Opm {
 
                 if( timestep_verbose_ )
                 {
-                    std::ostringstream ss;                    
-                    ss << "Substep( " << substepTimer.currentStepNum()-1 // it was already advanced by ++
-                                             << " ) finished at time " << unit::convert::to(substepTimer.simulationTimeElapsed(),unit::day) << " (days)." << std::endl << std::endl;
+                    std::ostringstream ss;
+                    ss << "well iterations = " << solver.wellIterations()
```
atgeirr: I agree with @totto82, this should not be printed at all if the separate well solve is off.

Also, I think the trailing "\n" could be dropped here, instead we should add a starting "\n" to the message in SimulatorBase_impl.hpp line 182 (the Time step message). That would make the report/schedule steps stand out more.

atgeirr:
Thanks for updating, I'll merge these PRs now.

PR #1046, opened 2016-06-21
Added DX/DY/DZ to norne_pvt.data

pgdr:


atgeirr:
This file is only used to test pvt classes, why do you want to add the geometry data?

pgdr:
Because they are _essential keywords_ in Eclipse, and because we are going away from the possibility of constructing full `EclipseState` objects on decks without geometry data.

atgeirr:
Thanks for the explanation, I'm just worried about the ever increasing strictness requirements.

joakim-hove:
> I'm just worried about the ever increasing strictness requirements.

The situation we are aiming at is:

1. To create an `EclipseGrid` you need either the cornerpoint keywords `COORD` and `ZCORN` or one of the combinations of cartesian keywords like `DX, DY` and `DZ`. It is not necessary with a valid `GRID` *section* to create an `EclipseGrid`.

2. It is not any longer possible to create limited `EclipseGrid` instances which only carry the dimsions.

3. For functions which currently take an `EclipseGrid` as input, but actually only need the dimensions we are changing the signature to take `GridDims` instead.

So - yes I agree that the current change can be seen to make things stricter, but another way to see it is that:

1. The current testcase had a not fully specified grid, and the `EclipseGrid` constructor had silently reduced the instance to a dimensions only instance, that was never detected because that grid was never really used. Now this is explicit.

2. With the introduction of the `GridDims` we actually facilitate simpler and more explicit testing in the situation where a fully specified grid is not really needed. For the current PR an alternative - probbly better - but more invasive route, could have been to change the test to only use `TableManager` which (will) only use the `GridDims`class - and thereby completely remove the grid dependency for this particular test.

I hope this can be merged.

atgeirr:
> The situation we are aiming at is:
> 
> To create an EclipseGrid you need either the cornerpoint keywords COORD and ZCORN or one of the combinations of cartesian keywords like DX, DY and DZ. It is not necessary with a valid GRID section to create an EclipseGrid.
> 
> It is not any longer possible to create limited EclipseGrid instances which only carry the dimsions.
> 
> For functions which currently take an EclipseGrid as input, but actually only need the dimensions we are changing the signature to take GridDims instead.
> 

This is *very* good.

> So - yes I agree that the current change can be seen to make things stricter, but another way to see it is that:
> 
> The current testcase had a not fully specified grid, and the EclipseGrid constructor had silently reduced the instance to a dimensions only instance, that was never detected because that grid was never really used. Now this is explicit.
> 
> With the introduction of the GridDims we actually facilitate simpler and more explicit testing in the situation where a fully specified grid is not really needed. For the current PR an alternative - probbly better - but more invasive route, could have been to change the test to only use TableManager which (will) only use the GridDimsclass - and thereby completely remove the grid dependency for this particular test.
> 

This was my hope, and I see that we are aligned in meaning this would be the best.
Requiring only what is needed (GridDims instead of a full EclipseGrid) gives better flexibility.

I am still a bit worried that all essential classes do or will require EclipseState, and that the EclipseState is becoming impossible to create without a 100% complete and valid deck, but I believe you are aware of this and we'll be fine moving forward.

> I hope this can be merged.

I will.

PR #1049, opened 2016-06-27
Return -1 in WellState::numPhases if there are no wells.

blattms:
This is an alternative to #1047. Closes OPM/opm-simulators#731

### DIFF
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -207,9 +207,17 @@ namespace Opm
         }
 
         /// The number of phases present.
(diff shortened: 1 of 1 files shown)


blattms:
On second thought this is also not a nice fix. Now we might get a segmentation fault later on instead of a floating point exception where the the problem is. Seems much harder to debug.

blattms:
Wrong usage might still result in using a wrong number of phases and will lead to bugs that are hard to find. I prefer to get a floating point exception that make debugging much more easy. Closing...

PR #1050, opened 2016-06-27
Make wells_equal work for null wells.

blattms:
This is needed to run test cases without wells with debugging on.
Without this commit we get a segmentation fault in an assert statement which compares to null well pointers.

### DIFF
--- a/opm/core/wells/wells.c
+++ b/opm/core/wells/wells.c
@@ -545,6 +545,12 @@ bool
 wells_equal(const struct Wells *W1, const struct Wells *W2 , bool verbose)
 /* ---------------------------------------------------------------------- */
 {
+    // Cater the case where W1 and W2 are the same (null) pointers.
+    if( W1 == W2 )
+    {
+        return true;
+    }
+
     bool are_equal = true;
     are_equal = (W1->number_of_wells == W2->number_of_wells);
     are_equal = are_equal && (W1->number_of_phases == W2->number_of_phases);
--- a/opm/core/wells/wells.c
+++ b/opm/core/wells/wells.c
@@ -550,7 +550,11 @@ wells_equal(const struct Wells *W1, const struct Wells *W2 , bool verbose)
     {
         return true;
     }
-
+    if( W1 == 0 || W2 == 0)
+    {
+        return false;
+    }
+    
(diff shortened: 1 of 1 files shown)


bska:
This will fail if one of `W1` **or** `W2` is `nullptr`.  This was the case before too, so no real change.  To be safe you could say something like
```C++
if ((W1 == nullptr) && (W2 == nullptr)) { return true; }
if ((W1 == nullptr) || (W2 == nullptr)) { return false; }
```

blattms:
You are right that this case is not catched. But this PR is a change, because where wells_equal is called it is made sure that both wells are null. See [lines 494-497](https://github.com/OPM/opm-core/blob/2303c43c03b57d4f8bf77ba63e36a0783895e7b5/opm/core/wells/wells.c#L494-L497).
Anyway, the case with one null pointer should be fixed now, too.

opm/core/wells/wells.c
```diff
@@ -545,6 +545,16 @@ bool
 wells_equal(const struct Wells *W1, const struct Wells *W2 , bool verbose)
 /* ---------------------------------------------------------------------- */
 {
+    // Cater the case where W1 and W2 are the same (null) pointers.
+    if( W1 == W2 )
+    {
+        return true;
+    }
+    if( W1 == 0 || W2 == 0)
```
bska: This is pure C (I forgot that), so better make it
```C
if ((W1 == NULL) || (W2 == NULL))
```

atgeirr:
Looks good, possible apart from the NULL/0 question: is 0 considered wrong in C? Will it typically generate warnings? If so it should be fixed, otherwise we can merge I think.

bska:
> is 0 considered wrong in C

Yes.  The null-pointer constant is always spelled `NULL` in C.

bska:
> Use `NULL` instead of 0 in null comparison.

Thanks!  I'll merge once Travis gives the thumbs up.

blattms:
0 -> NULL done

bska:
jenkins build this please

bska:
> I'll merge once Travis gives the thumbs up.

Travis approves and so do I.  I'll merge into master.  Thanks a lot!

PR #1051, opened 2016-06-30
Support of the keyword WECON

GitPaean:
Basically, according to economic check based on the production behavior, we need to stop/shut of some non-economical wells or completions. 

The part in opm-core is trying to support this mechanism and the downstream PR in opm-simulator is comings soon too.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -417,6 +417,6 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/core/wells/WellCollection.hpp
         opm/core/wells/WellsGroup.hpp
         opm/core/wells/WellsManager.hpp
-				opm/core/wells/DynamicListEconLimited.hpp
+        opm/core/wells/DynamicListEconLimited.hpp
         opm/core/wells/WellsManager_impl.hpp
 	)
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -417,5 +417,6 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/core/wells/WellCollection.hpp
         opm/core/wells/WellsGroup.hpp
         opm/core/wells/WellsManager.hpp
+				opm/core/wells/DynamicListEconLimited.hpp
         opm/core/wells/WellsManager_impl.hpp
 	)
--- a/opm/core/wells/DynamicListEconLimited.hpp
+++ b/opm/core/wells/DynamicListEconLimited.hpp
@@ -1,0 +1,57 @@
+/*
+  Copyright 2016 SINTEF ICT, Applied Mathematics.
+
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 3 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef OPM_DYNAMICLISTECONLIMITED_HPP
+#define	OPM_DYNAMICLISTECONLIMITED_HPP
+
+#include <vector>
+#include <string>
+
+#include <cassert>
+
+namespace Opm
+{
+
+    /// to handle the wells and connections voilating economic limits.
+    class DynamicListEconLimited
+    {
+    public:
+        bool anyWellEconLimited() const {
+            return !(m_shut_wells.empty());
+        };
+
+        bool wellEconLimited(const std::string& well_name) const {
+            return std::find(m_shut_wells.begin(), m_shut_wells.end(), well_name) != m_shut_wells.end();
+        };
+
+        void addShuttedWell(const std::string& well_name) {
+            // the well should not be in the list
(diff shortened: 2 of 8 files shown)


CMakeLists_files.cmake
```diff
@@ -417,5 +417,6 @@ list (APPEND PUBLIC_HEADER_FILES
         opm/core/wells/WellCollection.hpp
         opm/core/wells/WellsGroup.hpp
         opm/core/wells/WellsManager.hpp
+				opm/core/wells/DynamicListEconLimited.hpp
```
atgeirr: Use the same indentation as the lines surrounding it (just copy a line then change it). There are some tabs remaining here I think, we should remove eventually.

opm/core/wells/DynamicListEconLimited.hpp
```diff
@@ -0,0 +1,92 @@
...
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef OPM_DYNAMICLISTECONLIMITED_HPP
+#define	OPM_DYNAMICLISTECONLIMITED_HPP
```
atgeirr: Weird indent?

opm/core/wells/DynamicListEconLimited.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include <vector>
+#include <string>
+#include <map>
+
+#include <cassert>
+
+namespace Opm
+{
+
+    /// to handle the wells and connections voilating economic limits.
```
atgeirr: Typo: violating.

opm/core/wells/DynamicListEconLimited.hpp
```diff
@@ -0,0 +1,92 @@
...
+    class DynamicListEconLimited
+    {
+    public:
+
+        bool wellShuttedEconLimited(const std::string& well_name) const {
```
atgeirr: All use of "shutted" should be changed to "shut".

opm/core/wells/DynamicListEconLimited.hpp
```diff
@@ -0,0 +1,92 @@
...
+        void addStoppedWell(const std::string& well_name) {
+            assert( !wellShuttedEconLimited(well_name) );
+            assert( !wellStoppedEconLimited(well_name) );
+
+            m_stopped_wells.push_back(well_name);
+        }
+
+
+        // TODO: maybe completion better here
+        bool connectionClosedForWell(const std::string& well_name) const {
```
atgeirr: To make it clear that this is intended to indicate if any connection has been closed, would not "anyConnectionClosedForWell()" be a better name?

opm/core/wells/DynamicListEconLimited.hpp
```diff
@@ -0,0 +1,92 @@
...
+            return (m_cells_closed_connections.find(well_name) != m_cells_closed_connections.end());
+        }
+
+        const std::vector<int>& getClosedConnectionsForWell(const std::string& well_name) const {
```
atgeirr: This method should have the same name as the mutable variant below.

Also: why is the mutable variant needed?

opm/core/wells/WellsManager.cpp
```diff
@@ -332,12 +332,15 @@ namespace Opm
                                const double* permeability)
         : w_(0), is_parallel_run_(false)
     {
-        std::vector<double> dummy_well_potentials;
+        std::vector<double> dummy_well_potentials;;
```
atgeirr: Extra semicolon.

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -175,6 +185,15 @@ void WellsManager::createWellsFromSpecs(std::vector<const Well*>& wells, size_t
                     else
                     {
                         int cell = cgit->second;
+                        // check if the connection is closed due to economic limits
+                        if (!cells_connection_closed.empty()) {
+                            const bool connection_found = std::find(cells_connection_closed.begin(), cells_connection_closed.end(), cell)
```
atgeirr: Very long line, please split.

opm/core/wells/DynamicListEconLimited.hpp
GitPaean: > Also: why is the mutable variant needed?
We need to update it.

atgeirr:
Aside from minor nitpicks this is good. I'd like the mutable closedConnectionsForWell() to go away if possible though.

Will merge when updated, and review is done in opm-simulators.

opm/core/wells/DynamicListEconLimited.hpp
atgeirr: Then the addClosedConnectionsForWell() method should be used. In the implementation of that method you can directly access the member so you don't need this method.

Also (noticed just now): that method should be renamed addClosedConnectionForWell since it can only add a single and not multiple connections.

opm/core/wells/DynamicListEconLimited.hpp
GitPaean: Yes. it is better. I will fix it.

opm/core/wells/WellsManager_impl.hpp
bska: > Very long line, please split.

I would have thought that the return value was a greater concern than the line length.  Function `std::find()` returns an iterator, not a `bool`.

GitPaean: Sorry for the confusion. 

```const bool connection_found = std::find(cells_connection_closed.begin(), cells_connection_closed.end(), cell) != cells_connection_closed.end() ```

PR #1054, opened 2016-07-14
Drop useage of std::numeric_limits

qilicun:
Adapt to the changes in https://github.com/OPM/opm-simulators/pull/758


qilicun:
@bska 
This PR is related to https://github.com/OPM/opm-simulators/pull/758, please merge it if you think it's ok.

bska:
Looks good.  I'll merge into master.

PR #1055, opened 2016-07-19
fix incorrect derivative  of rock compressibility w.r.t. pressure

andlaus:
this is a relatively minor SNAFU over which I stumbled when trying to compare the Jacobian matrices produced by ebos with the ones from flow. I only ran SPE1 and SPE9 on it but did not run Norne. (I did not notice any significant difference for those.)

since

f(x) = 1 + g(x) + 0.5*g(x)*g(x)

the derivative is

f'(x) = 0 - g'(x) + 2*0.5*g(x) * g'(x) = g'(x) + g(x)*g'(x)

note that the previous incorrect values do not affect the quality of
the obtained results (if the tolerance of the non-linear solver is
chosen to be small enough), but it may have deteriorated convergence
rates.

### DIFF
--- a/opm/core/props/rock/RockCompressibility.cpp
+++ b/opm/core/props/rock/RockCompressibility.cpp
@@ -101,7 +101,8 @@ namespace Opm
         if (p_.empty()) {
             // Approximating poro multiplier with a quadratic curve,
             // we must use its derivative.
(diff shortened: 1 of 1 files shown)


bska:

PR #1059, opened 2016-08-30
Create eclipsegrid

joakim-hove:
With this PR the GridManager has got a new method `createEclipseGrid( )` which can be used to create an EclipseGrid for output. 

Depends on: 
 1. https://github.com/Ensembles/ert/pull/1217
 2. https://github.com/OPM/opm-parser/pull/905

Downstream PR in opm-simulators is coming, in addition to a similar PR in opm-grid.

OK - I encourage you to read the comment in `GridManager.hpp`; the gist of the matter is that creating a guaranteed 100% correct cornerpoint representation from the `struct UnstructureGrid` implementation is a large task - which will almost certainly require changes to the old C code in `opm-core/opm/core/grid/cpgpreprocess/`.

 1. Say that the current approach is good enough.
 2. Say that the current approach is good enough; but make the necessary changes to the `DuneCpGrid` class (I have not looked into this).
 3. Make the necessary changes to the `struct UnstructeredGrid` implementation.

Going all the way with 3) is probably *quite non-trivial*.

**Update:** The final commit has been completely reworked since the initial review. The fundamental idea is still the same, but with the following non-trivial changes:

1. The `ZCORN` copy is attached to the `struct UnstructuredGrid` instead of the `GridManager`.
2. The functionality to actually create the `EclipseGrid` instance is a loose function in the `UgGridHelpers` namespace instead of a method on the `GridManager` class.

The reason for these changes is to facilitate uniform access to the EclipseGrid from both `struct UnstructuredGrid` and `Dune::CpGrid` - see https://github.com/OPM/opm-grid/pull/230 for corresponding changes in opm-grid.

### DIFF
--- a/opm/core/grid/GridHelpers.cpp
+++ b/opm/core/grid/GridHelpers.cpp
@@ -149,5 +149,29 @@ FaceCellTraits<UnstructuredGrid>::Type faceCells(const UnstructuredGrid& grid)
 {
     return FaceCellsProxy(grid);
 }
+
+
+Opm::EclipseGrid createEclipseGrid(const UnstructuredGrid& grid, const Opm::EclipseGrid& inputGrid ) {
+    const int * dims = UgGridHelpers::cartDims( grid );
+
+    if ((inputGrid.getNX( ) == static_cast<size_t>(dims[0])) &&
+        (inputGrid.getNY( ) == static_cast<size_t>(dims[1])) &&
+        (inputGrid.getNZ( ) == static_cast<size_t>(dims[2]))) {
+        std::vector<int> updatedACTNUM;
+        const int* global_cell = UgGridHelpers::globalCell( grid );
+
+        if (global_cell) {
+            updatedACTNUM.assign( inputGrid.getCartesianSize( ) , 0 );
+            for (int c = 0; c < numCells( grid ); c++) {
+                updatedACTNUM[global_cell[c]] = 1;
+            }
+        }
+
+        return Opm::EclipseGrid( inputGrid, grid.zcorn, updatedACTNUM );
+    } else {
+        throw std::invalid_argument("Size mismatch - dimensions of inputGrid argument and current UnstructuredGrid instance disagree");
+    }
+}
+
 }
 }
--- a/opm/core/grid/GridHelpers.hpp
+++ b/opm/core/grid/GridHelpers.hpp
@@ -22,6 +22,7 @@
 #define OPM_CORE_GRIDHELPERS_HEADER_INCLUDED
 
 #include <opm/core/grid.h>
+#include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
 
 #include <opm/common/utility/platform_dependent/disable_warnings.h>
 #include <boost/range/iterator_range.hpp>
@@ -33,7 +34,7 @@ namespace Opm
 namespace UgGridHelpers
 {
 
-/// \brief Allows viewing a sparse table consisting out of C-array 
+/// \brief Allows viewing a sparse table consisting out of C-array
 ///
 /// This class can be used to convert two int array (like they are
 /// in UnstructuredGrid for representing the cell to faces mapping
@@ -47,7 +48,7 @@ public:
         typedef boost::iterator_range<const int*> BaseRowType;
         typedef BaseRowType::size_type size_type;
         typedef int value_type;
-        
+
         IntRange(const int* start_arg, const int* end_arg)
             : BaseRowType(start_arg, end_arg)
         {}
@@ -183,6 +184,31 @@ struct CellVolumeIteratorTraits<UnstructuredGrid>
 {
     typedef const double* IteratorType;
 };
+
+    /**
+       Will create an EclipseGrid representation (i.e. based on
+       ZCORN and COORD) of the current UnstructuredGrid
+       instance. When creating the UnstructuredGrid the detailed
+       cornerpoint information is discarded, and it is difficult
+       to go backwards to recreated ZCORN and COORD.
+
+       The current implementation is based on retaining a copy of the
+       zcorn keyword after the Minpvprocessor has modified it.
+
+       We then create a new EclipseGrid instance based on the original
+       input grid, but we "replace" the ZCORN and ACTNUM keywords with
+       the updated versions.
+
+       If the tolerance in the call to create_grid_cornerpoint( ) is
+       finite the grid processing code might collapse cells, the z
+       coordinate transformations from this process will *not* be
+       correctly represented in the EclipseGrid created by this
+       method.
+    */
+
+/// \brief Construct an EclipseGrid instance based on the inputGrid, with modifications to
+/// zcorn and actnum from the dune UnstructuredGrid.
+Opm::EclipseGrid createEclipseGrid(const UnstructuredGrid& grid, const Opm::EclipseGrid& inputGrid );
 
 /// \brief Get an iterator over the cell volumes of a grid positioned at the first cell.
 const double* beginCellVolumes(const UnstructuredGrid& grid);
--- a/opm/core/grid/GridManager.cpp
+++ b/opm/core/grid/GridManager.cpp
@@ -19,6 +19,7 @@
 
 #include "config.h"
 
+#include <opm/core/grid/GridHelpers.hpp>
 #include <opm/core/grid/GridManager.hpp>
 #include <opm/core/grid.h>
 #include <opm/core/grid/cart_grid.h>
@@ -132,6 +133,7 @@ namespace Opm
                                           const std::vector<double>& poreVolumes)
     {
         struct grdecl g;
+        int cells_modified = 0;
         std::vector<int> actnum;
         std::vector<double> coord;
         std::vector<double> zcorn;
@@ -157,7 +159,7 @@ namespace Opm
             // Currently the pinchProcessor is not used and only opmfil is supported
             //bool opmfil = inputGrid.getMinpvMode() == MinpvMode::OpmFIL;
             bool opmfil = true;
-            mp.process(poreVolumes, minpv_value, actnum, opmfil, zcorn.data());
+            cells_modified = mp.process(poreVolumes, minpv_value, actnum, opmfil, zcorn.data());
         }
 
         const double z_tolerance = inputGrid.isPinchActive() ? inputGrid.getPinchThresholdThickness() : 0.0;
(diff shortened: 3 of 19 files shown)


opm/core/grid/GridManager.cpp
```diff
@@ -137,38 +137,55 @@ namespace Opm
...
 
+    Opm::EclipseGrid GridManager::createEclipseGrid(const Opm::EclipseGrid& inputGrid ) {
+        std::vector<int> updatedACTNUM;
+
+        if (ug_->global_cell) {
+            const int* global_cell = ug_->global_cell;
+            updatedACTNUM.assign( inputGrid.getCartesianSize( ) , 0 );
+
+            for (int c = 0; c < ug_->number_of_cells; c++)
+                updatedACTNUM[global_cell[c]] = 1;
```
atgeirr: Braces, please...

opm/core/grid/MinpvProcessor.hpp
```diff
@@ -126,10 +122,12 @@ namespace Opm
                                 setCellZcorn(ii, jj, kk + 1, cz_below, zcorn);
                             }
                         }
+                        cells_modified++;
```
atgeirr: It's ok to keep it as is, but usually pre-increment should be preferred.

opm/core/grid/MinpvProcessor.hpp
```diff
@@ -114,6 +109,7 @@ namespace Opm
                         }
                         setCellZcorn(ii, jj, kk, cz, zcorn);
 
+
```
atgeirr: Usually avoid double blank lines inside functions (to contrast with more significant distinctions like between functions). Don't see the point here?

opm/core/grid/GridManager.cpp
```diff
@@ -137,38 +137,55 @@ namespace Opm
...
-            eclipseGrid->getPinchThresholdThickness() : 0.0;
+        const double z_tolerance = inputGrid.isPinchActive() ? inputGrid.getPinchThresholdThickness() : 0.0;
         ug_ = create_grid_cornerpoint(&g, z_tolerance);
         if (!ug_) {
             OPM_THROW(std::runtime_error, "Failed to construct grid.");
         }
     }
 
 
+    Opm::EclipseGrid GridManager::createEclipseGrid(const Opm::EclipseGrid& inputGrid ) {
```
atgeirr: It seems to me that inputGrid is expected to be the same grid as the one passed to the constructor. If so, should we not do some sanity checking? And would it not be reasonable to just keep the original shared_ptr around? If the interface is going to change to no longer give out a shared pointer that cannot be done though, I assume that is why you changed this class to take const refs?

atgeirr:
Mostly just nitpicking from me, but apart from that I do think the inputGrid comment above is relevant and should be answered in some way.

opm/core/grid/MinpvProcessor.hpp
joakim-hove: ,As long we are different developers on the project there must be an acceptance for *some* variation in style - this comment:

> Usually avoid double blank lines inside functions

is below any reasonable nitpick limit.

opm/core/grid/GridManager.cpp
joakim-hove: Yes - the grid passed in is expected to be the same as has previously been passed to the constructor; I agree that some sanity checking would be good. My "design" was at least partly guided by the least-invasive principle. One possibility could be to hold on to a pointer to the `EclipseGrid` instance; currently the other constructors make the solution a bit unclean - but maybe they go anyway?

We are steadily working the towards reducing the (excessive) use of shared pointers in opm-parser, and yes the plan is that `EclipseState::getInputGrid()` will return a const reference instead of a shared pointer in the future.

joakim-hove:
Having thought things over a bit I might propose a slightly different solution tomorrow.

opm/core/grid/MinpvProcessor.hpp
atgeirr: The PR contains this isolated whitespace change, so I assume it is wanted for some reason? I think it is justified to discuss that, since I disagree with the change and don't see the point of it. And I do think it is valid to discuss whitespace and formatting with a view to improving legibility.

opm/core/grid/GridManager.cpp
joakim-hove: Coming.

joakim-hove:
OK - various nitpicks addressed.

atgeirr:
Fine!

I can live with the present solution if we add some sanity checking (checking cartesian size should go a long way towards guarding agains errors), but if you find a better solution that is good!

joakim-hove:
> Having thought things over a bit I might propose a slightly different solution tomorrow.

As indicated in the updated PR description that really failed.

atgeirr:
> As indicated in the updated PR description that really failed.

Current solution is good enough, you have license to merge at will.

bska:
Since I didn't read all of it, I'll just ask a high-level question.  What part of the `EclipseGrid` instance do you (we) need for output purposes?  I gather the COORD/ZCORN information is needed to give some kind of faithful output of Cartesian properties (e.g., TRANX), but is there a reason we can't output a fully unstructured format?

joakim-hove:
>I gather the COORD/ZCORN information is needed to give some kind of faithful output of Cartesian properties (e.g., TRANX), 

Yes

> but is there a reason we can't output a fully unstructured format?

Well - of course we *could* do that; but the purpose of the current effort is to output results in an Eclipse compatible format.

bska:
> the purpose of the current effort is to output results in an Eclipse compatible format.

Very well.  There **is** the unstructured ECLIPSE grid format (NDCOORD, CELLFACS, FACENODS) and we could output all connections as NNCs.  That would be ECL compatible too, and more in line with what we actually get from the [`create_grid_cornerpoint()`](https://github.com/OPM/opm-core/blob/master/opm/core/grid/cornerpoint_grid.c#L163-L164) constructor.  On the other hand, that may be too much work in the output module.

On the other hand, from the point of view of the primary consumer of the grid structure (i.e., the simulator), retaining the Cartesian structure is somewhat wasteful.

bska:
jenkins build this ert=1217 opm-parser=905 please

bska:
jenkins build this opm-parser=905 please

opm/core/grid/grid.c
```diff
@@ -69,6 +71,26 @@ create_grid_empty(void)
...
+      If the allocation fails the grid->zcorn pointer will be
+      NULL. The consequence of that is that if we want to output a
+      cornerpoint grid it will not reflect the adjustments we have
+      done in in the Minpvprocessing. That is a consequence we can
+      live with.
+    */
+
+    G->zcorn = realloc(G->zcorn , zcorn_elements * sizeof * G->zcorn );
+    if (G->zcorn) {
+        memcpy(G->zcorn , zcorn , zcorn_elements * sizeof * G->zcorn );
```
atgeirr: I always get nervous about these sizeof tricks... This will use zcorn_elements multiplied by the size of (*G->zcorn), which is a double, correct?

I wish it could be written sizeof(*G->zcorn) for clarity, but I guess that can segfault when it is a null pointer?

opm/core/grid/GridHelpers.hpp
```diff
@@ -184,6 +185,31 @@ struct CellVolumeIteratorTraits<UnstructuredGrid>
...
+
+       If the tolerance in the call to create_grid_cornerpoint( ) is
+       finite the grid processing code might collapse cells, the z
+       coordinate transformations from this process will *not* be
+       correctly represented in the EclipseGrid created by this
+       method.
+    */
+
+/// \brief Construct an EclipseGrid instance based on the inputGrid, with modifications to
+/// zcorn and actum from the dune CPGrid
```
atgeirr: No CPGrid here, also typo: actum->actnum.

opm/core/grid/GridHelpers.cpp
```diff
@@ -149,5 +149,29 @@ FaceCellTraits<UnstructuredGrid>::Type faceCells(const UnstructuredGrid& grid)
...
+        const int* global_cell = UgGridHelpers::globalCell( grid );
+
+        if (global_cell) {
+            updatedACTNUM.assign( inputGrid.getCartesianSize( ) , 0 );
+            for (int c = 0; c < numCells( grid ); c++) {
+                updatedACTNUM[global_cell[c]] = 1;
+            }
+        }
+
+        return Opm::EclipseGrid( inputGrid, grid.zcorn, updatedACTNUM );
```
atgeirr: We should probably either assert(grid.zcorn) or even make it throw if it is null. Also, the function doc below should state that it requires the input grid to have a non-null zcorn field.

This is important since not all creation paths will yield a grid with zcorn.

atgeirr:
Having looked over this and the opm-grid PR I think it is well done and can be merged, with the minor issue of preventing calling this with null zcorn. As stated here I'd like that requirement to be stated in docs and checked inside the function, that also applies in opm-grid.

However, a better solution might be to make it ok to pass nullptr for zcorn, meaning that it is unchanged from the input grid. That would require the ert function to accept that, and the opm-parser method must not unconditionally call data() in the version that takes a vector but something like `zcorn.empty() ? nullptr : zcorn.data()`.

opm/core/grid/grid.c
```diff
@@ -69,6 +71,26 @@ create_grid_empty(void)
...
+
+    /*
+      If the allocation fails the grid->zcorn pointer will be
+      NULL. The consequence of that is that if we want to output a
+      cornerpoint grid it will not reflect the adjustments we have
+      done in in the Minpvprocessing. That is a consequence we can
+      live with.
+    */
+
+    G->zcorn = realloc(G->zcorn , zcorn_elements * sizeof * G->zcorn );
```
bska: I don't care about the comment.  Do not ever do this.  The safe way of using `realloc()` is
```C
void *x;
/* ... */
x = realloc(p, n * sizeof *p);
if (x != NULL) { p = x; }
```

opm/core/grid/grid.c
bska: > I always get nervous about these sizeof tricks

It's not a trick.  The `sizeof` operator either operates on a type (syntax `sizeof(T)`) or an expression (syntax `sizeof expr`).  As `sizeof` is an unevaluated context (its result being fully available at compile time) we don't have to worry about null-pointers.  That said, if I were writing the above, I'd put the dereference operator on `G->zcorn`, e.g.,
```C
memcpy(G->zcorn, zcorn, zcorn_elements * sizeof *G->zcorn)
```

joakim-hove:
OK - comments from @atgeirr and @bska addressed in updated version.

opm/core/grid/grid.c
```diff
@@ -83,6 +97,9 @@ allocate_grid(size_t ndims     ,
     G = create_grid_empty();
 
     if (G != NULL) {
+        /* zcorn cache - only for output. */
+        G->zcorn = NULL;
```
bska: Strictly speaking, you don't really need this.  Function `create_grid_empty()` already makes all pointer fields `NULL`.  It doesn't hurt but it may lead to some proliferation of unneeded assignments.

PR #1060, opened 2016-09-01
Output writer cell data

babrodtk:
Enables outputting cell data to UNRST files.

See also sibling pull requests in OPM/opm-simulators#805 and OPM/opm-output#84


bska:
jenkins build this please

PR #1061, opened 2016-09-06
Warning fixes

babrodtk:
This fixes some warnings that appear on my setup.


opm/core/grid/GridHelpers.hpp
```diff
@@ -190,10 +190,6 @@ const double* beginCellVolumes(const UnstructuredGrid& grid);
 /// \brief Get an iterator over the cell volumes of a grid positioned after the last cell.
 const double* endCellVolumes(const UnstructuredGrid& grid);
 
-/// \brief Get the cell centroid of a face.
-/// \param grid The grid whose cell centroid we query.
-/// \param face_index The index of the corresponding face.
-const double* faceCentroid(const UnstructuredGrid& grid, int face_index);
```
babrodtk: This is a duplicate of the FaceCentroidTraits<UnstructuredGrid> version below.

opm/core/utility/Units.hpp
```diff
@@ -216,6 +218,16 @@ namespace Opm
                 return q / unit;
             }
         } // namespace convert
+
+
+#ifndef HAS_ATTRIBUTE_UNUSED
```
babrodtk: OPM_UNUSED is a macro defined in opm/material/common/Unused.hpp that perhaps could be refactored into opm-common, and then used here?

andlaus: >  that perhaps could be refactored into opm-common, and then used here?

sure. (the reason that it is in opm-material is that it seems like  so far nobody upstream of that module seemed to have an interest in it/was aware of it.)

atgeirr:
Thanks, will merge. I agree we should put Unused.hpp in opm-common, will you do a PR @andlaus? I think opm/common/utility/Unused.hpp is a good path for it.

Since HAS_ATTRIBUTE_UNUSED comes from the config.h though I am not sure it can reliably be used in public headers though?

PR #1062, opened 2016-09-09
[WIP] Use provided set of deactivated wells in parallel.

blattms:
Before this commit we tried to compute whether a well is represented on the processor using the grid information. Due to the overlap region and possible completion on deactivated cells of the global grid this is not even possible. E.g. we cannot distinguish whether a completion is just not represented on the domain of a process or the corresponding cell is not active in the simulation.

With this commit we refactor to passing the well manager an explicit list of name of wells that should be completely neglected. This information can easily by computed after the loadbalancer has computed partitions.

This is work in progess. We need to change opm-grid, and opm-simulators to make this run in parallel correctly, too. For non-parallel runs everything should be fine as we stay backwards compatible


blattms:
Discarded the WIP tag as accompanying PRs have been published.

bska:
jenkins build this please

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -121,6 +121,7 @@ void WellsManager::createWellsFromSpecs(std::vector<const Well*>& wells, size_t
                                         const double* permeability,
                                         const NTG& ntg,
                                         std::vector<int>& wells_on_proc,
+                                        const std::unordered_set<std::string>& ignored_wells,
```
GitPaean: something very tiny here. Will it be beneficial to use `deactivated_wells` all the places?

blattms:
On Thu, Sep 15, 2016 at 08:02:07AM -0700, Kai Bao wrote:
> GitPaean commented on this pull request.
> 
> 
> 
> > @@ -121,6 +121,7 @@ void WellsManager::createWellsFromSpecs(std::vector<const Well*>& wells, size_t
>                                          const double* permeability,
>                                          const NTG& ntg,
>                                          std::vector<int>& wells_on_proc,
> +                                        const std::unordered_set<std::string>& ignored_wells,
> 
> something very tiny here. Will it be beneficial to use `deactivated_wells` all the places?
> 

It might be a good idea to use consistent names. Unfortunately I am
very bad in naming things. Maybe we should use
locally_(deactivated|ignored)_wells. Maybe ignored is better since
these wells might still be active when looking on the global view of
the simulation?

Anyway, we could merge before changing names.

atgeirr:
Looks ok in itself, reviewing the whole set of PRs now.

PR #1063, opened 2016-09-13
Bugfix: Let NTMISC determine number of MISC tables

totto82:
Number of misc tables is given by NTMISC (MISCNUM) and not NTSFUN
(SATNUM)

Tested on Model 2 with the solvent option.

### DIFF
--- a/opm/core/props/satfunc/RelpermDiagnostics.cpp
+++ b/opm/core/props/satfunc/RelpermDiagnostics.cpp
@@ -167,7 +167,7 @@ namespace Opm{
 
 
         if (deck->hasKeyword("MISCIBLE")) {
-            const int numMiscNumIdx = deck->getKeyword("MISCIBLE").getRecord(0).getItem("NTMISC").get< int >(0);
+            const int numMiscNumIdx = miscTables.size();
             const std::string msg = "Number of misc regions: " + std::to_string(numMiscNumIdx) + "\n";
             OpmLog::info(msg);
             for (int miscNumIdx = 0; miscNumIdx < numMiscNumIdx; ++miscNumIdx) {
--- a/opm/core/props/satfunc/RelpermDiagnostics.cpp
+++ b/opm/core/props/satfunc/RelpermDiagnostics.cpp
@@ -160,13 +160,21 @@ namespace Opm{
             if (deck->hasKeyword("SSFN")) {
                 ssfnTableCheck_(ssfnTables.getTable<SsfnTable>(satnumIdx), satnumIdx+1);
             }
-            if (deck->hasKeyword("MISC")) {
-                miscTableCheck_(miscTables.getTable<MiscTable>(satnumIdx), satnumIdx+1);
-            }
             if (deck->hasKeyword("MSFN")) {
                 msfnTableCheck_(msfnTables.getTable<MsfnTable>(satnumIdx), satnumIdx+1);
             }
         }
+
+
+        if (deck->hasKeyword("MISCIBLE")) {
+            const int numMiscNumIdx = deck->getKeyword("MISCIBLE").getRecord(0).getItem("NTMISC").get< int >(0);
+            const std::string msg = "Number of misc regions: " + std::to_string(numMiscNumIdx) + "\n";
+            OpmLog::info(msg);
+            for (int miscNumIdx = 0; miscNumIdx < numMiscNumIdx; ++miscNumIdx) {
+                miscTableCheck_(miscTables.getTable<MiscTable>(miscNumIdx), miscNumIdx+1);
+            }
+        }
+
     }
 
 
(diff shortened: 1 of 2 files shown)


joakim-hove:
Are you certain that these changes should not be in the parser?

totto82:
Yes. The number of MISC tables is given my NTMISC in the MISCIBLE kwy. This is currently specified in the parser.

joakim-hove:
There might be duplication here, but I really think the fix should be in the parser - see EclipseState/Tables/TableManager.cpp.

totto82:
I still don't understand what you mean when you say the fix should be in the parser. The parser currently loads NTMISC tables for MISC. The problem occur when the diagnostics tries to get access to MISC tables.  In the diagnostics MISC is assumed to use SATNUM. It is in the same loop as the SWAT etc kwy. This PR just makes a separate loop for MISCNUM related keywords.

joakim-hove:
OK - I might be completely wrong. Will analyse properly later today.

joakim-hove:
OK - I am probably wrong here.

opm/core/props/satfunc/RelpermDiagnostics.cpp
```diff
@@ -160,13 +160,21 @@ namespace Opm{
...
+
+        if (deck->hasKeyword("MISCIBLE")) {
+            const int numMiscNumIdx = deck->getKeyword("MISCIBLE").getRecord(0).getItem("NTMISC").get< int >(0);
+            const std::string msg = "Number of misc regions: " + std::to_string(numMiscNumIdx) + "\n";
+            OpmLog::info(msg);
+            for (int miscNumIdx = 0; miscNumIdx < numMiscNumIdx; ++miscNumIdx) {
+                miscTableCheck_(miscTables.getTable<MiscTable>(miscNumIdx), miscNumIdx+1);
+            }
+        }
+
```
joakim-hove: OK - what I reacted to here was that it should not be necessary to consult the deck at all. The following code should do the trick:
```
const std::string msg = "Number of misc regions: " + std::to_string(miscTables.size()) + "\n";
for (int i=0; i < miscTables.size(); i++) {
     if (miscTables.hasTable(i))
          miscTableCheck_(miscTables.getTable<MiscTable>(i), i+1);
}
```

I guess the same changes could be applied to the other tables.

atgeirr: Does that mean that the size() of miscTables is always the same as NTMISC?

Also: is it so that any individual table in that container may not be present (i.e. `hasTable(i)` is false) so it's a possibly sparse thing? I don't see any handling of that situation in the original code or the PR (apart from you comment), that would be a bug then? The `miscTableCheck_()` method unconditionally accesses the various columns.

joakim-hove: > Does that mean that the size() of miscTables is always the same as NTMISC?

Yes

>Also: is it so that any individual table in that container may not be present (i.e. hasTable(i) is false) so it's a possibly sparse thing? 

Yes; it is a sparse thing. But the [TableContainer](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Tables/TableContainer.hpp) has some machinery so that it always safe to ask for a table in the range [0,size()).

>I don't see any handling of that situation in the original code or the PR (apart from you comment), that would be a bug then? 

Not really a bug - but potentially the same underlying table can be checked repeatedly.

>The miscTableCheck_() method unconditionally accesses the various columns.

Columns is something different?

atgeirr: > Columns is something different?

Not sure what you mean, the code I refer to is:
```
const auto& frac = miscTables.getSolventFractionColumn();
const auto& misc = miscTables.getMiscibilityColumn();
```
If that is ok to do when `hasTable(i)` is false I guess there is no real problem. I take your comment to mean that in that situation we get an adjacent table if we still insist on asking for table `i`?

joakim-hove: > If that is ok to do when hasTable(i) is false I guess there is no real problem. 

I agree.

> I take your comment to mean that in that situation we get an adjacent table if we still insist on asking for table i?

Yes - then we will get table `i - 1`, or `i - 2` .... 

As a detail on naming it would in my opinion be better if these were named in singular:
```C++
const auto& frac = miscTable.getSolventFractionColumn();
const auto& misc = miscTable.getMiscibilityColumn();
```
as `miscTable` - because this code is run in a single-table context.

atgeirr:
Ok I think we have a consensus. This is correct and can be merged when the line
```
const int numMiscNumIdx = deck->getKeyword("MISCIBLE").getRecord(0).getItem("NTMISC").get< int >(0);
```
is replaced with using the size() of the container.

atgeirr:
Thanks for the patch and update, merging!

PR #1064, opened 2016-09-15
Timestepper based on user input

totto82:
A new timestepper that reads timesteps from "timesteps" generated using
ecl_summary "DECK" TIME > timesteps
and applies it to the simulator

Also a parameter timestep.initial_step_length (default 1 day) is added
to controll the frist timestep.

Can be tested specifying 
timestep.control=hardcoded and
timestep.control.filename=timesteps

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -72,7 +72,7 @@ namespace Opm {
         , solver_restart_max_( param.getDefault("solver.restart", int(10) ) )
         , solver_verbose_( param.getDefault("solver.verbose", bool(true) ) && terminal_output )
         , timestep_verbose_( param.getDefault("timestep.verbose", bool(true) ) && terminal_output )
-        , suggested_next_timestep_( -1.0 )
+        , suggested_next_timestep_( unit::convert::from(param.getDefault("timestep.initial_step_length", double(1.0) ), unit::day) )
         , full_timestep_initially_( param.getDefault("full_timestep_initially", bool(false) ) )
     {
         // valid are "pid" and "pid+iteration"
@@ -95,6 +95,10 @@ namespace Opm {
             const double decayrate  = param.getDefault("timestep.control.decayrate",  double(0.75) );
             const double growthrate = param.getDefault("timestep.control.growthrate", double(1.25) );
             timeStepControl_ = TimeStepControlType( new SimpleIterationCountTimeStepControl( iterations, decayrate, growthrate ) );
+        } else if ( control == "hardcoded") {
+            const std::string filename    = param.getDefault("timestep.control.filename", std::string("timesteps"));
+            timeStepControl_ = TimeStepControlType( new HardcodedTimeStepControl( filename ) );
+
         }
         else
             OPM_THROW(std::runtime_error,"Unsupported time step control selected "<< control );
@@ -202,7 +206,7 @@ namespace Opm {
 
                 // compute new time step estimate
                 double dtEstimate =
-                    timeStepControl_->computeTimeStepSize( dt, linearIterations, relativeChange );
+                    timeStepControl_->computeTimeStepSize( dt, linearIterations, relativeChange, substepTimer.simulationTimeElapsed());
 
                 // limit the growth of the timestep size by the growth factor
                 dtEstimate = std::min( dtEstimate, double(max_growth_ * dt) );
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -72,7 +72,7 @@ namespace Opm {
         , solver_restart_max_( param.getDefault("solver.restart", int(10) ) )
         , solver_verbose_( param.getDefault("solver.verbose", bool(true) ) && terminal_output )
         , timestep_verbose_( param.getDefault("timestep.verbose", bool(true) ) && terminal_output )
-        , suggested_next_timestep_( unit::convert::from(param.getDefault("timestep.initial_step_length", double(1.0) ), unit::day) )
+        , suggested_next_timestep_( -1.0 )
         , full_timestep_initially_( param.getDefault("full_timestep_initially", bool(false) ) )
     {
         // valid are "pid" and "pid+iteration"
--- a/opm/core/simulator/TimeStepControl.cpp
+++ b/opm/core/simulator/TimeStepControl.cpp
@@ -84,29 +84,25 @@ namespace Opm
     ////////////////////////////////////////////////////////
 
     HardcodedTimeStepControl::
-    HardcodedTimeStepControl( const std::string filename)
+    HardcodedTimeStepControl( const std::string& filename)
     {
         std::ifstream infile (filename);
-        double time;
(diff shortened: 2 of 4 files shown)


opm/core/simulator/TimeStepControl.hpp
```diff
@@ -111,12 +111,33 @@ namespace Opm
...
+    ///  Input generated from summary file using the ert application:
+    ///  ecl_summary DECK TIME > filename
+    //
+    ///////////////////////////////////////////////////////////////////////////////////////////////////////////////
+    class HardcodedTimeStepControl : public TimeStepControlInterface
+    {
+    public:
+        /// \brief constructor
+        /// \param filename   filename contaning the timesteps
+        HardcodedTimeStepControl( const std::string filename);
```
atgeirr: Constructor should be `explicit`.

atgeirr: Also: prefer const reference for argument (possible exception: when you know you are going to copy the argument inside the function, passing by value may allow move optimization when called with a temporary).

opm/core/simulator/TimeStepControl.hpp
```diff
@@ -111,12 +111,33 @@ namespace Opm
...
-        double computeTimeStepSize( const double dt, const int iterations, const RelativeChangeInterface& relativeChange ) const;
+        double computeTimeStepSize( const double dt, const int iterations, const RelativeChangeInterface& relativeChange, const double /*simulationTimeElapsed */ ) const;
 
     protected:
         const int     target_iterations_;
     };
 
+    ///////////////////////////////////////////////////////////////////////////////////////////////////////////////
+    ///
+    ///  HardcodedTimeStepControl
```
atgeirr: The doc should clearly state that the input file is assumed to have timesteps in days. Nice touch to show how to generate it from summary!

opm/core/simulator/TimeStepControl.cpp
```diff
@@ -74,6 +77,38 @@ namespace Opm
...
+    //
+    ////////////////////////////////////////////////////////
+
+    HardcodedTimeStepControl::
+    HardcodedTimeStepControl( const std::string filename)
+    {
+        std::ifstream infile (filename);
+        double time;
+        std::string line;
+        std::getline(infile, line); //assumes two lines before the timing starts.
```
atgeirr: This is a bit weird, but probably correct. I'd like for it to verify the format however. If not, it is easy for a user to assume that this will accept a file with just the days.

Even better, it should accept both. That can be done by ignoring lines starting with `--` in the loop.

opm/core/simulator/TimeStepControl.cpp
```diff
@@ -74,6 +77,38 @@ namespace Opm
...
+        std::getline(infile, line);
+
+        while (!infile.eof() ) {
```
bska: I have not looked at the context but this loop control is hardly ever correct.  C++ is not Pascal.

opm/core/simulator/TimeStepControl.hpp
```diff
@@ -111,12 +111,33 @@ namespace Opm
...
+    ///  Input generated from summary file using the ert application:
+    ///  ecl_summary DECK TIME > filename
```
joakim-hove: If interesting it would be a ten minutes job to add a switch e.g. `--timesteps` to the summary application which could dump in whatever format convenient.

opm/core/simulator/TimeStepControl.hpp
totto82: done

opm/core/simulator/TimeStepControl.cpp
totto82: good idea. done.

opm/core/simulator/TimeStepControl.cpp
totto82: Thanks for the comment.  
Is while(std::getline(infile, line) safer?

opm/core/simulator/TimeStepControl.hpp
totto82: I think the current format is ok. It actually reads the actual time and not the timesteps as such. 
I will change the variable name and add a comment to avoid confusion.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -72,7 +72,7 @@ namespace Opm {
         , solver_restart_max_( param.getDefault("solver.restart", int(10) ) )
         , solver_verbose_( param.getDefault("solver.verbose", bool(true) ) && terminal_output )
         , timestep_verbose_( param.getDefault("timestep.verbose", bool(true) ) && terminal_output )
-        , suggested_next_timestep_( -1.0 )
+        , suggested_next_timestep_( unit::convert::from(param.getDefault("timestep.initial_step_length", double(1.0) ), unit::day) )
```
atgeirr: Sorry I missed this the first time through: when requesting user parameter input in non-SI units we should make the expected unit obvious, therefore the parameter should be "timestep.initial_step_length_days".

atgeirr:
Travis complains about this:
```
For keyword FGOR at reportstep 1
(days, reference value) and (days, check value) = (10, 242.703) and (10, 242.726)
The absolute deviation is 0.0233307. The tolerance limit is 0.02
The relative deviation is 9.61194e-05. The tolerance limit is 1e-05
Program threw an exception: [/home/travis/build/OPM/opm-output/opm/test_util/summaryRegressionTest.cpp:79] Deviation exceed the limit.
95% tests passed, 1 tests failed out of 21
Total Test time (real) =  62.95 sec
The following tests FAILED:
	 21 - compareECLFiles_SPE9_CP (Failed)
Errors while running CTest
```
Perhaps the change to suggested_next_timestep_ could be upsetting things. It seems to me you now initialize that to 1 day (unless explicitly setting it) whereas it used to be -1 second (an invalid value). I have not checked carefully, but I believe this might cause all runs with adaptive timestepping to default to an initial 1 day step whereas before the default would be 0.3 * initial report step size.

Is this a wanted change? It does not seem related to the new timestepper class.

totto82:
I think this is  a wanted change. It probably belongs in a different PR as it is unrelated to the rest of this PR. The reason for the deviation for SPE9_CP is that TUNING is used to specify the initial step to be 4. 

Start using values from TUNING in AdaptiveTimeStepping instead of the current approach where param_ is used will fix this problem. I will revert the initial timestepping change and make a new PR where TUNING is used.

opm/core/simulator/TimeStepControl.cpp
```diff
@@ -87,6 +87,7 @@ namespace Opm
     HardcodedTimeStepControl( const std::string& filename)
     {
         std::ifstream infile (filename);
+        if (infile.is_open()) {
         std::string::size_type sz;
```
atgeirr: Please either reindent the section below so that indentation is consistent, or (better): check for the inverse condition `!infile.is_open()` and put the throw up here. Then the remainder need not be in the scope of the if-statement any more.

totto82: Sorry. I tried to multitask and that's usually not a good idea. Is something missing now?

atgeirr:
Forgot merging this before the weekend -- better late than never.

Issue #1066, opened 2016-09-21
WellState::init has memory problems


blattms:
I just ran SPE1 in parallel. In this case the situation occurs that the last (second of two) wells has no connections and memory bounds are violated in https://github.com/OPM/opm-core/blob/ce78dc5560a492f1e89d64caa1c0816723ed1a9e/opm/core/simulator/WellState.hpp#L134
The  problem is that `wells->well_connpos[1]` is 1 and that is also the size of the `wells->well_cells` array.

As I do not have the complete overview on the well treatment I would like to get some input here. Shall we either disregard wells with no connections completely here or treat them like a wells for which `well_controls_get_current_type(ctrl) == BHP` holds?

BTW this feels like a Déjà vu

blattms:
Same issue here:
https://github.com/OPM/opm-core/blob/ce78dc5560a492f1e89d64caa1c0816723ed1a9e/opm/core/simulator/WellState.hpp#L94
Not sure whether this handled correctly for multi segment wells.

blattms:
Even for sequential runs we could have this situation if the last well is shut and all compeltions are in non-active cells.

atgeirr:
> BTW this feels like a Déjà vu

Not again...?

Joke aside, I think we should disregard wells with no connections completely. The initialization code makes no sense for such wells, and should not be needed as the wells are not connected in any way with the rest of the problem.

I am not quite sure what happens to such wells in the simulator though -- I thought that at some point such wells would be treated as shut and totally ignored, but that may not be 100% correct. In any case, initializing their bhps with a simple default (1 atmosphere?) should not do any harm as it should never interact with the rest of the problem. A slightly more sophisticated way may be to see if the connection-less well has a bhp control, and use that limit (currently that is only used if bhp is the current active control).

blattms:
I guess this is connected to output. If we totally neglect them then they will not appear there (no well curves).

There also might be cases where perforations appear at a later stage or vanish. Therefore a well with no valid perforations might have some at another point in the simulation.

atgeirr:
Wells can shut, open, shut again etc. so if they participate in the simulation can change. The point about how output treats this is a good one: should we arbitrary insert zeros for bhps of shut wells in the output for example? Perhaps @joakim-hove can shed a little light on this issue?

blattms:
I will repeat myself here, too (sorry for the noise).

The problem I experienced was caused by buggy opm-grid loadbalancing code that was never in master.

[...]

PR #1067, opened 2016-09-21
Zero initialize wells with no perforations.

blattms:
For these wells access its well_cells might read over the bounds
an array if they are the last wells in the struct. Therefore
we cannot initialize first_cell and the well control is uninitialized,
to.

With this commit theses wells are now detected and the bhp, thp, and well_rates
are initialized to zero.

### DIFF
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -65,6 +65,7 @@ namespace Opm
                 for (int w = 0; w < nw; ++w) {
                     assert((wells->type[w] == INJECTOR) || (wells->type[w] == PRODUCER));
                     const WellControls* ctrl = wells->ctrls[w];
+                    const int num_perf_this_well = wells->well_connpos[w + 1] - wells->well_connpos[w];
 
                     // setup wellname -> well index mapping
                     {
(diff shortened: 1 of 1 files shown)


blattms:
Fixes #1066

atgeirr:
Looks fine. Will merge when green.

blattms:
It seems like I might have produced this situation myself due to local changes in the loadbalancing of opm-grid. Not sure whether this will appear with the master of all repositories.

blattms:
I am inclined to close this.

 The problem was caused by mess in the loadbalancer code on my system. It is not an issue for master for the test cases that I ran.

Nevertheless this situation can occur for open wells with all completions in active cells or closed according to the limiter. Which means that in eclipse the well is open and has open perforations but these are closed in the simulator. Not sure how and where to handle this. Skipping those wells in the simulator would be beneficial.

blattms:
Interesting discussion in OPM/opm-output#93

atgeirr:
> I am inclined to close this.

I agree that while according to your tests the PR may not be necessary, it still makes it more robust to "unexpected" input (completionless wells) even if that input may not occur with the current code.

blattms:
Your call.

atgeirr:
I'll merge this. The alternative would be an assert() guarding the use of the first connection, but this is clearer and more robust.

PR #1068, opened 2016-09-22
[WIP] hack: derive second saturation component from the two loaded data

akva2:
this is just a hack highlighting what i believe is a flaw in the restart handling.

only two saturations are loaded from file and the second component is left with some default values.


opm/core/utility/Compat.hpp
```diff
@@ -129,6 +129,12 @@ inline void solutionToSim( const data::Solution& sol,
                 state.saturation() );
     }
 
+    if (stride == 3) {
+      // properly initialize last saturation component
+      for (size_t i = 0; i < state.saturation().size(); i += stride)
+        state.saturation()[i+1] = 1.0-state.saturation()[i]-state.saturation()[i+2];
```
joakim-hove: Oh my oh my; I can recognize magic numbers and hardcoding when I see it! This might very well be correct, and also fit nicely with the current implementation, but it does not look good.

akva2: yes, it's a fugly hack which highlights a real problem. i thought the commit message and PR description made that clear enough.

akva2:
taken care of now.

PR #1069, opened 2016-09-22
Add solvent to solution vector if present

totto82:
Also move BlackoilSolventState from opm-simulators to opm-core.


atgeirr:
Actually I'd rather move the other state classes in the opposite direction. Any particular reason not to?

totto82:
They are needed in opm-output.

atgeirr:
> They are needed in opm-output

This cannot be right, as opm-output does not depend on opm-core (it's the other way around).

totto82:
You are right. They are not needed in output, but in Compat.hpp (in opm-core). Is it safe to move Compat.hpp to opm-simulators? As fare as I see it is only used in  SimulatorFullyImplicitBlackoilOutput.cpp.

atgeirr:
I think that should be safe to do, yes.

jokva:
It would be even better if you could kill off Compat.hpp.

totto82:
This PR is supersede by #1071

PR #1070, opened 2016-09-22
[WIP] Draft fix restart

akva2:
This allows for perfect restart, see https://github.com/OPM/opm-output/pull/95

i'm in particular worried about the templatification in Compat.hpp - the data is not available in the base WellState class so it was a quick fix.


jokva:
> i'm in particular worried about the templatification in Compat.hpp - the data is not available in the base WellState class so it was a quick fix.

Compat.hpp should die, and I thought I was rather clear on that. It was a hack to not break everything when opm-output was split from core, but the design decisions here are up to the core/simulator crew and probably requires some refactoring.

akva2:
properly taken care of now.

PR #1071, opened 2016-09-23
Remove Compat.hpp

totto82:
Compat.hpp is moved to opm-simulators

This PR supersedes  #1069


atgeirr:
Bug: file must be removed from cmake file list, and added to the list in opm-simulators.

totto82:
@atgeirr I have removed the file from the list in opm-core and added it to opm-simulators

PR #1072, opened 2016-09-23
Fix a few warnings

atgeirr:
After this, opm-core compiles warning-less for me.

### DIFF
--- a/opm/core/props/satfunc/RelpermDiagnostics.cpp
+++ b/opm/core/props/satfunc/RelpermDiagnostics.cpp
@@ -109,8 +109,10 @@ namespace Opm{
                                          std::shared_ptr< const Deck > deck)
     {
         const int numSatRegions = deck->getKeyword("TABDIMS").getRecord(0).getItem("NTSFUN").get< int >(0);
-        const std::string msg = "Number of saturation regions: " + std::to_string(numSatRegions) + "\n";
-        OpmLog::info(msg);
+        {
+            const std::string msg = "Number of saturation regions: " + std::to_string(numSatRegions) + "\n";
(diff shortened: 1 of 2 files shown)


opm/core/props/satfunc/RelpermDiagnostics.cpp
```diff
@@ -109,8 +109,10 @@ namespace Opm{
                                          std::shared_ptr< const Deck > deck)
     {
         const int numSatRegions = deck->getKeyword("TABDIMS").getRecord(0).getItem("NTSFUN").get< int >(0);
-        const std::string msg = "Number of saturation regions: " + std::to_string(numSatRegions) + "\n";
-        OpmLog::info(msg);
+        {
+            const std::string msg = "Number of saturation regions: " + std::to_string(numSatRegions) + "\n";
+            OpmLog::info(msg);
+        }
```
joakim-hove: Ahhh - finally someone else creating random scopes in the middle of a function; I :heart: it!

atgeirr: Not sure if you are ironic or not... It was the easiest way to avoid a warning when the `msg` name was reused further down, in any case.

joakim-hove:
I'm not ironic; personally I really like local scopes as a way to create structure and visualise variable lifetime within a function.

jokva:
If you make your variables `const` you don't need compound statements to restrict them. :)

atgeirr:
> If you make your variables const you don't need compound statements to restrict them

Sure, and by default I make all local variables const these days (I think most people do). In this case the variables were all const, so it did not have anything to do with constness though.

PR #1073, opened 2016-09-26
Only call writeTimeStep for real sub steps.

blattms:
Previously, we also called it when the full time step was done.

As the simulator writes that information anyway and we cannot call it a sub step, we omit the final write in the adaptive time stepper.

This is the patch proposed in OPM/opm-simulators#823. Closes OPM/opm-simulators#823.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -230,7 +230,10 @@ namespace Opm {
                 }
 
                 // write data if outputWriter was provided
-                if( outputWriter ) {
+                // if the time step is done we do not need
+                // to write it as this will be done by the simulator
+                // anyway.
+                if( outputWriter && !substepTimer.done() ) {
                     bool substep = true;
                     const auto& physicalModel = solver.model();
                     outputWriter->writeTimeStep( substepTimer, state, well_state, physicalModel, substep);


blattms:
I am wondering whether there is an easy way to inspect which substep (times) where written.

atgeirr:
This looks correct, and fixes a problem, so it should be merged. I just want to ask @joakim-hove or @jokva first: 

With this, for a single report step that is split in 4 substeps, writeTimeStep() will be called  four times, the three first after the first 3 substeps, with the `substep` argument equal to `true` and finally once after the complete report step, with `substep` equal to `false`.

Is that how it should be done? What is the effect of the `substep` argument?

jokva:
`isSubstep = false` means summary and restart files won't be written.

atgeirr:
> isSubstep = false means summary and restart files won't be written.

That sounds strange to me, are you sure it is not the opposite?

jokva:
I mean exactly the opposite, yes. My bad.

atgeirr:
OK, in that case I guess this PR is ok and should go in. Merging!

atgeirr:
Post-merge comment: is it not slightly odd to have substep mean "don't write summary and restart"? It would be more obvious if the parameter was named something like "write_restart_and_summary" (and default etc. flipped to true)?

Also: is the only thing preventing us from having output at all substeps just flipping the bool in the call  here, or are there other problems (apart from the question of "have the user asked for it in the deck").

jokva:
We've considered that before yes, it's inherited from the opm-core code.

The deck also kicks in for restart, but I think you should get a restart step for all substeps *if* that step was requested in RestartConfig.

atgeirr:
Sorry, but I have some trouble mapping my questions to your answer(s)... Could you clarify?

jokva:
Summary is written when `isSubstep = false`. Restart checks ` if(!isSubstep && restart.getWriteRestartFile(report_step))`, so I honestly doubt it works well with substeps at all (my immediate guess is overwrite, but @joakim-hove might be able to clarify).

This is behaviour from the opm-core implementation.

https://github.com/OPM/opm-output/blob/master/opm/output/eclipse/EclipseWriter.cpp#L657

joakim-hove:
> With this, for a single report step that is split in 4 substeps, writeTimeStep() will be called four times, the three first after the first 3 substeps, with the substep argument equal to true and finally once after the complete report step, with substep equal to false.

This is correct. 

The concept of substep maybe takes a bit too much room in the API - there is also probably some legacy here:

1. The normal behavior is to write summary data for every timestep, and the restart data only for complete report steps,  the purpose of the `substep` switch is to support this behavior. However; in Eclipse (not OPM yet) it is actually possible to configure both summary results *only* at the report steps and also restart files at *every* timestep. 

2. The output format (filenames & headers) distinguishes between results at report steps and at substeps, and the `substep` boolean flag plays a role in this as well.

Due to the second point (I think) the output routine needs to know whether this is a substep or not, but the question of *whether* to write restart and summary results to disk should ideally be resolved with a config object from the `EclipseState` - we actually have it.

atgeirr:
So since we are passing all the required information in any case, substep or not (at least I believe it is true), the details of whether to write summary and/or restart at any given step can actually be handled fully inside opm-parser? That sounds good.

joakim-hove:
> the details of whether to write summary and/or restart at any given step can actually be handled fully inside opm-parser? 

Yes - that is right. Admittedly the relevant config objects can not yet properly internalize the freak situations: "Output restart for every timestep", and "Output summary only for report steps" - but that can be fixed.

blattms:
> So since we are passing all the required information in any case,
> substep or not (at least I believe it is true), the details of
> whether to write summary and/or restart at any given step can
> actually be handled fully inside opm-parser? That sounds good. 

Not in my ears. I would appreciate if it would be possible to only
pass the information that will be written. This would allows to limit
the data needed to be transferred in a parallel run.

atgeirr:
> This would allows to limit
> the data needed to be transferred in a parallel run

Good point. To do this properly then, some configuration-query and related logic will be required also on the simulator side.

alfbr:
There are config objects in opm-parser that should contain all the information the simulator needs to know what to output. @joakim-hove and @jokva probably has full overview on that. I support Markus, some effort on the simulator side to enable on-demand output there would be great.

joakim-hove:
> There are config objects in opm-parser that should contain all the information the simulator needs to know what to output. 

Yes that is true. However there are some significant caveats to be aware of:

1. Currently the simulator calls *one single method* - [`writeTimestep()`](https://github.com/OPM/opm-output/blob/master/opm/output/eclipse/EclipseWriter.cpp#L618) which will check the relevant configuration objects and write what is requested of summary, restart and rft. In principle the simulator could check the relevant config objects itself, and then directly call the correct correct write routine(s). This could potentially reduce the amount of of copying, but at the price of polluting the simulator with output details.

2. Both the summary writer and the RFT writer uses field data of e.g. pressure and in-place values as rawdata. So in a situation where we want to output e.g. block pressures in the summary file we still need a full[1] transfer of the pressure field from the simulator to the output, even if we are not writing a restart file. 

So; although we in principle have the information required to reduce the amount of transfer, it comes with a significant price in complexity. 

[1]: In principle we could only transfer the requested values, but that would *really* pollute the simulator.

atgeirr:
> So; although we in principle have the information required to reduce the amount of transfer, it comes with a significant price in complexity.

Let's not move in this direction yet, then: getting output correct and stable is more important than this, although parallel performance will suffer eventually. However: any fully scalable output solution will probably not involve writing eclipse binary files as we know them today (imagine 10000 processes trying to coordinate the writing of a unified restart file), but require something designed for the parallel case.

PR #1076, opened 2016-09-28
Add contructor for adaptiveTimeStepper that uses values from TUNING

totto82:
Some of the tuning values from the TUNING keywords is used to tune the
time-stepping instead of the equivalent ones in param_.

This only implements part of the TUNING keyword. (The ones that where already implemented in OPM) Even thought I don't have any immediate plans for extending the usage of TUNING in the simulator, using TUNING instead of specifying parameters directly saves me some time in my work.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping.hpp
@@ -45,6 +45,17 @@ namespace Opm {
         AdaptiveTimeStepping( const parameter::ParameterGroup& param,
                               const bool terminal_output = true );
 
+        //! \brief contructor taking parameter object
+        //! \param tuning Pointer to ecl TUNING keyword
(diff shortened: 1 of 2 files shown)


atgeirr:
The content of this is fine, and it will be merged. There are two things that should be changed first, however:
 - The body of the constructor is identical to the body of the old (still present) constructor. That should be moved to a private init() method to avoid duplication.
 - The constructor should take all const arguments, and the shared_ptr should be changed to a const reference. The shared_ptr provided at this point by Schedule will probably be gone soon.

As a followup of the second point, OPM/opm-simulators#843 must then be changed slightly to dereference the shared pointer (when opm-parser changes the interface to return a const reference, that will be the only location to require change).

atgeirr:
jenkins build this opm-simulators=843 please

PR #1077, opened 2016-09-28
Log the message of exceptions in the catch clause of adaptive time stepper

blattms:
In parallel each breakdown of the solver was reported once for every process. WIth this commit we only do this logging on one process. Note that this patch is based on PR OPM/opm-common#179. Without each message will get logged twice.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -57,6 +57,17 @@ namespace Opm {
                 return solver_.model().relativeChange( previous_, current_ );
             }
         };
+
+        template<class E>
+        void logException(const E& exception, bool verbose)
+        {
+            if( verbose )
+            {
+                std::ostringstream message;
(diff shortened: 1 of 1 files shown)


blattms:
I have edited the description as I accidentally pressed the button before writing one.

atgeirr:
I don't see why this requires the opm-common PR since it does not involve the macro at all?

In any case, the change here looks good, and ensures the catching message is logged on one process instead of written to cerr on all (the throwing is a separate issue, as discussed in the opm-common PR), thus fixing something that was forgotten when cout usage was converted to using the log some time ago.

So: I'd like to merge this, is there something I have misunderstood, since I think it should be done without regard to the fate of the opm-common PR?

blattms:
You are right. There is no dependency.

atgeirr:
> There is no dependency

Then I'd like to merge this, if you'd rebase it. However since I just broke the Travis chain I'll not merge it until tomorrow in any case.

blattms:
rebased on current master.

PR #1078, opened 2016-09-28
set the Rv and Rs factors to the saturated values for cells which have no gas and no oil

andlaus:
the purpose of this is to get a more defined behaviour when doing the
gravity correction/upstream cell determination in the flux term.

I consider this to be just a kludge, so if anyone has a better idea of
what the composition for the non-existing gas and oil phases is,
please tell me. (note that generic compositional models do not exhibit
this issue because the composition of all fluids is always fully
defined because each component is assumed to dissolve in every phase.)

also be aware that this PR is independent of OPM/opm-simulators#845.

### DIFF
--- a/opm/core/simulator/initStateEquil.hpp
+++ b/opm/core/simulator/initStateEquil.hpp
@@ -25,7 +25,7 @@
 #include <opm/core/grid/GridHelpers.hpp>
 #include <opm/core/simulator/EquilibrationHelpers.hpp>
 #include <opm/core/simulator/BlackoilState.hpp>
-#include <opm/core/props/BlackoilPropertiesInterface.hpp>
+#include <opm/core/props/BlackoilPropertiesFromDeck.hpp>
 #include <opm/core/props/BlackoilPhases.hpp>
 #include <opm/core/utility/RegionMapping.hpp>
 #include <opm/core/utility/Units.hpp>
@@ -163,7 +163,7 @@ namespace Opm
         phaseSaturations(const Grid&             grid,
                          const Region&           reg,
                          const CellRange&        cells,
-                         BlackoilPropertiesInterface& props,
+                         BlackoilPropertiesFromDeck& props,
                          const std::vector<double> swat_init,
                          std::vector< std::vector<double> >& phase_pressures);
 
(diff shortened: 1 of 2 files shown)


atgeirr:
It's a known singular thing, and we just hope it does not bite us... I have some problem understanding why this should matter for the upstream determination, could you point to the code that is affected?

> note that generic compositional models do not exhibit
> this issue because the composition of all fluids is always fully
> defined because each component is assumed to dissolve in every phase

Is this true? If you use component fractions (that sum to one for a phase) they would have the same kind of singularity when a phase disappears? I guess that means you do not use such fractions?

andlaus:
> I have some problem understanding why this should matter for the upstream determination, could you point to the code that is affected?

yeah, it is quite subtle: what's the upstream cell for a phase depends on the phase density because of the gravity term, and phase density depends on the composition (i.e., Rs and Rv). Note that is  does not depend on the phase actually is present or not in a given cell.

In practice this affected the values of the threshold pressure, which lead to me discovering this in the first place.

maybe it can be argued that the Rs and Rv should be zero for the water-only case, but this also leads to a discontinuity if a hydrocarbon phase appears.

> Is this true? If you use component fractions (that sum to one for a phase) they would have the same kind of singularity when a phase disappears? I guess that means you do not use such fractions?

I suppose it depends on the exact formulation: The compositional models I dealt with so far all assume chemical equilibrium, i.e., each component can always dissolve into every phase to some extend. (that is if their fugacity coefficient for the phase is not infinite, but in the infinite case you'll probably have larger problems anyway.)

atgeirr:
> In practice this affected the values of the threshold pressure, which lead to me discovering this in the first place

I see (I think). My take: arbitrary Rs/Rv yields arbitrary upwind directions for the phase that is missing. While this should result in no difference in flux (since the phase is missing) if the situation is encountered during simulation, the arbitrary potential difference comes into play through the threshold pressures, which will have an effect once the (missing-on-initialization) phase appears at the thresholded connection. Does that look right?

atgeirr:
I have looked at the code, and intend to merge this.

Since this will have a non-zero (if minor) effect on solutions, as usual I'll ask @totto82 to verify that it does not negatively effect model 2(.2).

andlaus:
> Does that look right?

yes :+1:

totto82:
I will test it and report back. 
Also @osae and I have noticed convergence issues in flow when gas or oil appears in a water only cell. We are testing out different solutions, but it should not effect this PR.

totto82:
Tests are ok.

atgeirr:
Thanks for patch and testing, merging!

PR #1079, opened 2016-09-29
update petsc code

akva2:
- api changes in newer versions
- do not manually destroy the preconditioner. this is, and has always
  been, owned by the ksp object and dies with its destruction.


atgeirr:
Looks ok, but I assume this is not actually tested by the Travis and Jenkins systems? I'm perfectly willing to merge this small PR in any case...

akva2:
there is no testing on travis or jenkins, but i have tested a bit it locally, regression test and some other small setups.  note that you also need the prerequsite in opm-common for this to take much effect.

petsc doesn't go very deep in opm, in particular nothing below opm-core supports it.

atgeirr:
Good enough, I think. Merging!

PR #1080, opened 2016-09-30
changed: generate unit type conversion constants on compile-time

akva2:
as a bonus it avoids unused variable warnings in compile
units only using parts of the conversion units.

ref https://github.com/OPM/opm-parser/pull/893

### DIFF
--- a/opm/core/utility/Units.hpp
+++ b/opm/core/utility/Units.hpp
@@ -48,13 +48,13 @@ namespace Opm
     namespace prefix
     /// Conversion prefix for units.
     {
-        const double micro = 1.0e-6;  /**< Unit prefix [\f$\mu\f$] */
-        const double milli = 1.0e-3;  /**< Unit prefix [m] */
-        const double centi = 1.0e-2;  /**< Non-standard unit prefix [c] */
-        const double deci  = 1.0e-1;  /**< Non-standard unit prefix [d] */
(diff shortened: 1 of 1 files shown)


akva2:

opm/core/utility/Units.hpp
```diff
@@ -75,8 +75,8 @@ namespace Opm
     {
         ///\name Common powers
         /// @{
-        inline double square(double v) { return v * v;     }
-        inline double cubic (double v) { return v * v * v; }
+        constexpr double square(double v) { return v * v;     }
```
atgeirr: Are you sure we can remove the `inline` here? I am not familiar enough with the details of constexpr to know.

akva2: constexpr is something that is evaluated at compile time, not runtime. so inline makes no sense in the context.

atgeirr: But it is also a function, right? Can I pass a pointer to it for example?

akva2: only if you point to some constexpr object.

atgeirr: I was unclear, will try again: can I pass a function pointer to square() to something that takes function pointers?

akva2: no. square does not exist as a function after compilation.

PR #1081, opened 2016-10-04
Header file moved in opm-output.

joakim-hove:
Followup to: https://github.com/OPM/opm-output/pull/101


atgeirr:
Merge at will when/if merging the other one.

PR #1084, opened 2016-10-10
update packaging to reflect new dependency tree

akva2:


akva2:
shameless self-merge.

PR #1088, opened 2016-10-14
group control_updating well production targets within a group. 

GitPaean:
This is an improvement (no complete solution) towards better group control, building based our current way in handling well control and previous implementation of WellsGroup. Some old implementation from WellsGroup is changed based on current understanding, and also the PR is tailored to satisfy the current need, which is focusing on the redistribution of well targets within one production group.

I believe the redistribution logic of group targets from multi-layer group structures will be similar, while with some more complications possible. We are not addressing complete and general group control yet, which needs more examples designed to help figure all the possible logic can be involved.

`FLD` keywords from GCONPROD is addressed for one level of group control. 

The guide rate is always based on the well potentials for now and also the examples we need to address, so we need to use option ‘compute_well_potentials=true’ when running example with group control. There should be some real logic based on the specification to decide whether we need to compute well potentials.

The PR from the opm-simulator and some testing results are also coming.


atgeirr:
Thanks! I changed the title to indicate that this is work in progress, since it has certain limitations that might want to address before we merge the feature.

GitPaean:
This PR is ready for review and discussion. It depends on the PR OPM/opm-parser#955 and is required by PR OPM/opm-simulators#862.

GitPaean:
The last commit causes some problems. My compilation script reported the compilation error while did not rebuild flow yesterday. I did not notice that. 
Sorry about that. Working on fixing it.

GitPaean:
The results just got recovered.

atgeirr:

opm/core/wells/WellCollection.hpp
```diff
@@ -110,16 +114,50 @@ namespace Opm
...
+        bool needUpdateProductionTargets() const;
+
+        /// Number of the well nodes.
+        size_t numNode() const;
+
+        /// Getting the ith well node.
+        WellNode* getNode(size_t i) const;
+
+        /// Updating the well targets based on the well rates.
+        void updateWellTargets(const std::vector<double> well_rates);
```
atgeirr: Should be a const reference.

opm/core/wells/WellsManager_impl.hpp
```diff
@@ -409,6 +409,7 @@ WellsManager::init(const Opm::EclipseState& eclipseState,
 
         const auto& fieldGroup = schedule.getGroup(fieldNode->name());
 
+        well_collection_.setHavingVREPGroups(false);
```
atgeirr: Rather initialize in definition ( member var = false), then you can remove this call.

opm/core/wells/WellCollection.cpp
```diff
@@ -189,4 +208,118 @@ namespace Opm
...
+                }
+            }
+        }
+    }
+
+    bool WellCollection::havingVREPGroups() const {
+        return having_vrep_groups_;
+    }
+
+    void WellCollection::setHavingVREPGroups(const bool vrep) {
```
atgeirr: This function can be removed, the flag should be initialized to false in the class definition, and updated from class methods only.

opm/core/wells/WellsGroup.cpp
```diff
@@ -619,28 +674,168 @@ namespace Opm
...
+    }
+
+    void WellsGroup::updateWellInjectionTargets(const std::vector<double>& /*well_rates*/)
+    {
+        // NOT doing anything yet.
+        // Will finish it when having an examples with more than one injection wells within same injection group.
+    }
+
+
+    bool WellsGroup::isProducer() const
```
atgeirr: The methods isProducer, isInjector and getProductionRate seems to only be properly implemented for WellNode. Therefore they should not be virtual methods, and not defined for WellsGroup or the base class.

This assumes that we actually have a WellNode when we need to call these methods.

GitPaean: Unfortunately the `children_` is `WellsGroupInterface*`.  I think that is why I decided to make it a virtual function.

opm/core/wells/WellCollection.cpp
GitPaean: It is called in WellsManager.

bska:
jenkins build this please

atgeirr:
Finally I have been able to review this, and with the final revisions I think this is ready to be merged.

Thanks for the effort!

PR #1090, opened 2016-10-17
Convergence failure is "problem" not "error"

atgeirr:
This should address the issue raised in OPM/opm-simulators#824.

Also (not directly related) ensures that logging is only happening on the first rank in a parallel run (the event it is fixed for is the rare event of complete breakdown of time-stepping).

Requires OPM/opm-common#185 (but not vice versa).

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -291,7 +291,12 @@ namespace Opm {
             {
                 // increase restart counter
                 if( restarts >= solver_restart_max_ ) {
-                    OPM_THROW(Opm::NumericalProblem,"Solver failed to converge after " << restarts << " restarts.");
+                    const auto msg = std::string("Solver failed to converge after ")
+                        + std::to_string(restarts) + " restarts.";
+                    if (solver_verbose_) {
+                        OpmLog::error(msg);
+                    }
+                    OPM_THROW_NOLOG(Opm::NumericalProblem, msg);
                 }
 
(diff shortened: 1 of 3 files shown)


opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -291,7 +291,12 @@ namespace Opm {
             {
                 // increase restart counter
                 if( restarts >= solver_restart_max_ ) {
-                    OPM_THROW(Opm::NumericalProblem,"Solver failed to converge after " << restarts << " restarts.");
+                    const auto msg = std::string("Solver failed to converge after ")
+                        + std::to_string(restarts) + " restarts.";
+                    if (solver_verbose_) {
+                        OpmLog::error(msg);
+                    }
+                    OPM_THROW_NOLOG(Opm::NumericalProblem, msg);
```
atgeirr: > Does it not get logged at line 221 AdaptiveTimeStepping_impl.hpp, too?

Yes and no: the underlying reason for failure (for example too many nonlinear iterations or linear solver convergence failure) is already logged. The normal action is then (after logging that as a "problem" but not "error") to reduce the timestep size and try again. The part here however, is only triggered if we try to do this over and over again without succeeding (I think the default is reducing the timestep max 10 times). It is not a situation from which the simulator can recover, and therefore it is an "error" not a "problem".

atgeirr:
With the added commit, the ROCK message issue should be taken care of (from the opm-core side). Do you think this can be merged, @blattms?

blattms:
Sure! Have not tested again with the last commit but I am inclined to
trust jenkins here.

PR #1092, opened 2016-10-19
fix `make install`

andlaus:
CMakeLists_files.txt needed some adaptation after the recent unit
changes. this is analogous to OPM/opm-parser#943


atgeirr:
Thanks, merging.

PR #1093, opened 2016-10-21
Make WellsManager compile with c++-11.

dr-robertk:
This PR replaces the initializer list on the constructor of WellsManager with the equally valid empty construction of objects. Otherwise the code won't compile without the c++14 flag which is not needed anywhere else.

### DIFF
--- a/opm/core/wells/WellsManager.hpp
+++ b/opm/core/wells/WellsManager.hpp
@@ -94,8 +94,8 @@ namespace Opm
                      const double* permeability,
                      const DynamicListEconLimited& list_econ_limited,
                      bool is_parallel_run=false,
-                     const std::vector<double>& well_potentials={},
-                     const std::unordered_set<std::string>& deactivated_wells = {});
+                     const std::vector<double>& well_potentials=std::vector<double>(),
+                     const std::unordered_set<std::string>& deactivated_wells = std::unordered_set<std::string> ());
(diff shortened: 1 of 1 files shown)


atgeirr:
The Travis failure is due to this:
```
/home/travis/build/OPM/opm-simulators/opm/autodiff/FlowMain.hpp:432:106: error: could not convert ‘{{Opm::Log::MessageType::Note, <expression error>}, {Opm::Log::MessageType::Info, <expression error>}, {Opm::Log::MessageType::Warning, <expression error>}, {Opm::Log::MessageType::Error, <expression error>}, {Opm::Log::MessageType::Problem, <expression error>}, {Opm::Log::MessageType::Bug, <expression error>}}’ from ‘<brace-enclosed initializer list>’ to ‘const std::map<long int, int>’
                                                    {Log::MessageType::Bug, msgLimits.getBugPrintLimit(0)}};
```

Will look into it. Could it be a similar kind of error to the one you are fixing here?

dr-robertk:
Yeah. Seems like it. For some reasons the c++-14 did not get activated for my configuration. But I guess c++-11 should be good enough. The rest compiles though.

atgeirr:
jenkins build this please

atgeirr:
Jenkins is green, Jenkins is green, merging.

PR #1095, opened 2016-10-27
Fix saturation family diagnostics for two-phase case.

atgeirr:
Noticed while doing OPM/opm-simulators#878.

@qilicun can you verify this?

### DIFF
--- a/opm/core/props/satfunc/RelpermDiagnostics.cpp
+++ b/opm/core/props/satfunc/RelpermDiagnostics.cpp
@@ -18,6 +18,7 @@
 */
 
 #include <opm/core/props/satfunc/RelpermDiagnostics.hpp>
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

qilicun:
@atgeirr 

This Looks good. I will do some tests. Thanks.

qilicun:
I confirm this works well with 2p case.

atgeirr:
> I confirm this works well with 2p case.

I have checked Norne and SPE, so it should be mergable. I'd appreciate if you'd also carefully read the code to look for mistakes in the logic.

qilicun:
>I have checked Norne and SPE, so it should be mergable. I'd appreciate if you'd also carefully read the code to look for mistakes in the logic.

I didn't find any mistakes in the logic... you can merge at your convenience。

PR #1097, opened 2016-10-28
Make it possible to set initial timestep

totto82:
Default is kept at -1.0. I.e. this PR does not change the current
behavior.

Useful for testing restarts.


atgeirr:
This provides another way to specify this (by parameters), in addition to the existing TUNING (by deck) way. To be clear, that is what you want?

totto82:
Yes. It is useful for testing restarts.

PR #1100, opened 2016-11-04
fixed: use correct indices for perforation pressure

akva2:
This makes restart tests pass again. Please, sanity check, I'm on very thin grounds ;)

### DIFF
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -259,8 +259,8 @@ namespace Opm
 
                     auto& completion = well.completions[ i ];
                     completion.index = active_index;
-                    completion.pressure = this->perfPress()[ well_index + i ];
-                    completion.reservoir_rate = this->perfRates()[ well_index + i ];
+                    completion.pressure = this->perfPress()[ itr.second[1] + i ];
+                    completion.reservoir_rate = this->perfRates()[ itr.second[1] + i ];
                 }
                 assert(num_perf_well == int(well.completions.size()));
             }


blattms:
Well, that does not seem to be your fault. That code is seriously under documented. How should fellow developers know what the map_entry is (besides its type) if this code:
```
        typedef std::array< int, 3 >  mapentry_t;
        typedef std::map< std::string, mapentry_t > WellMapType;
```
lacks any kind of documentation?

akva2:
well i solved that mystery by some grepping. it's (well_index, data_start_idx, num_datas). the problem is more that it's very hard to track which data is indexed in what way with all these flat vectors, in particular since i haven't worked much with reservoir simulators before. i'm fairly certain my fix is okay though as evidenced by passing restarts (and printing and such - we now reproduce well state, before it was shuffled and contained duplicate data for the completions).

opm/core/simulator/WellState.hpp
```diff
@@ -259,8 +259,8 @@ namespace Opm
 
                     auto& completion = well.completions[ i ];
                     completion.index = active_index;
-                    completion.pressure = this->perfPress()[ well_index + i ];
-                    completion.reservoir_rate = this->perfRates()[ well_index + i ];
+                    completion.pressure = this->perfPress()[ itr.second[1]+i ];
+                    completion.reservoir_rate = this->perfRates()[ it.second[1] + i ];
```
blattms: My compiler complains a bit as it cannot see that `it` should be `itr`. Compilers are just dump.

akva2: Even with their newfangled deep leaning and whatnot

atgeirr:
It's embarrassing, but I noticed this error reviewing the corresponding code in opm-simulators while failing to see it in this one. Looks correct now.

Travis fails though, not sure why.

akva2:
travis failures are expected, as there is reference data produced with the broken code.

atgeirr:
OK, makes sense. Merging.

atgeirr:
Actually, not merging!

I just realized, this will cause tests to fail (even though it is a fix) until we have an update ready for opm-data. I think we must wait until we have that, right?

akva2:
yes. i'm on it, but i'm doing the restart tests in parallel and polishing my scripting to share with others in the process. see https://github.com/OPM/opm-simulators/pull/822 - restart test passed.

akva2:
okay. https://github.com/OPM/opm-simulators/pull/822 shows successful execution on jenkins (along with a couple of restart tests).

this pr can be merged in tandem with https://github.com/OPM/opm-data/pull/138

https://github.com/OPM/opm-simulators/pull/822 can be merged separately, but preferably after the others. it also includes (replaces) https://github.com/OPM/opm-simulators/pull/816 which can be closed at the same time.

atgeirr:
Excellent, merging!

atgeirr:
I do not see quite how OPM/opm-simulators#822 includes OPM/opm-simulators#816, I assume you mean that the same case gets compared, not that it includes the implementation?

akva2:
it includes a test that covers the intention of 816.

PR #1104, opened 2016-11-16
rewriting the findWellNode function

GitPaean:
with assuming we should always find the well in the well collection.


atgeirr:
Thanks, merging!

PR #1105, opened 2016-11-16
fix some valgrind errors in the init code

andlaus:
this fixes some valgrind errors which I found while doing the twophase capability for flow_ebos: In all previously tested cases, these errors were probably non-fatal because the memory illegally accessed here is likely to be allocated (but after this function was finished it contained garbage).

note that I'm not completely sure if this patch is semantically correct, so I'd appreciate some input who understands it. (what is a "z" anyway? I remember a computer game by that name, but that's definitely not it.)

### DIFF
--- a/opm/core/simulator/initState_impl.hpp
+++ b/opm/core/simulator/initState_impl.hpp
@@ -896,12 +896,15 @@ namespace Opm
             for (int col = 0; col < np; ++col) {
                 z[0] += A_a[0 + np*col]*s[col];
                 z[1] += A_l[1 + np*col]*s[col];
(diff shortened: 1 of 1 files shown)


bska:
The patch is correct for Water/Oil and Water/Oil/Gas systems.  Oil/Gas and Water/Gas systems are not handled here (and weren't before either).  This is nevertheless a strict improvement over the previous implementation so I'll merge into master once Jenkins approves.

> what is a "z" anyway?

Array of component volumes at surface conditions (for each cell).

bska:
jenkins build this please

PR #1106, opened 2016-11-17
Remove tests that were moved to opm-output.

atgeirr:
I'm pretty sure these tests have been fully incorporated into opm-output and can be deleted.

@joakim-hove, @jokva: Can you confirm that? (You may merge if affirmative.)


jokva:
Haven't I already submitted the PR to remove these? I've made the patch, surely.

But yes.

PR #1112, opened 2016-11-24
Improve simulator report

andlaus:
this allows the performance report at the end of the simulation to be more relevant. in particular, the time needed for linearization and that for linear solves is kept track independently.

note that to take advantage of this, the actual simulators need to use the new possibilities. that is done in a separate opm-simulators PR.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -26,6 +26,7 @@
 #include <opm/core/simulator/SimulatorTimer.hpp>
 #include <opm/core/simulator/AdaptiveSimulatorTimer.hpp>
 #include <opm/core/simulator/TimeStepControl.hpp>
+#include <opm/core/utility/StopWatch.hpp>
 #include <opm/common/Exceptions.hpp>
 #include <opm/common/OpmLog/OpmLog.hpp>
 #include <dune/istl/istlexception.hh>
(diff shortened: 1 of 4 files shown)


bska:
jenkins build this please

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -258,6 +260,7 @@ namespace Opm {
                     restarts = 0;
                 }
 
+#if 0
```
atgeirr: Why is this section disabled? Is this written our somewhere else? If so, where?

If it should be removed it should be deleted instead of #if-ed out.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -268,15 +271,19 @@ namespace Opm {
...
                     OpmLog::info(ss.str());
                 }
+#endif
 
                 // write data if outputWriter was provided
                 // if the time step is done we do not need
                 // to write it as this will be done by the simulator
                 // anyway.
                 if( outputWriter && !substepTimer.done() ) {
+                    Dune::Timer perfTimer;
```
atgeirr: This is a new dependency for opm-core, please use the already-used StopWatch instead (and yes, we *may* unify this but we have not yet...). In any case, the include directive is missing I think.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
andlaus: basicallly to unclutter the terminal output. do you prefer to re-enable or remove it?

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
atgeirr: I was thinking Dune::Timer, not dune-common as such, was a new dep.

I would like StopWatch to go totally std::chrono and move to opm-common, but have not got around to it yet.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
atgeirr: I'm not sure, since I did not get an answer... If it's printed in a different part of the code it should go completely, otherwise I'd like it to be reinstated (it's already in an explicitly verbosity-controlled section).

andlaus: ok, the message is back.

atgeirr:
I'd like to get this merged, but there are two minor things remaining:
 - the MINPV processor was moved to opm-grid, so it cannot be modified here anymore (sorry...!)
 - I'd like the "failed" member to be added as discussed elsewhere (and used as before from the sequential model in opm-simulators).

andlaus:
> - The MINPV processor was moved to opm-grid, so it cannot be modified here anymore (sorry...!)

ok, PR rebased and a separate one opened for opm-grid: OPM/opm-grid#257

> - I'd like the "failed" member to be added as discussed elsewhere (and used as before from the sequential model in opm-simulators).

in principle that's no problem as long as you do not expect these attributes to be correct. I propose to leave this to a separate PR.

atgeirr:
> in principle that's no problem as long as you do not expect these attributes to be correct. I propose to leave this to a separate PR.

OK.

andlaus:
the travis failures seem to be unrelated to this PR. (IIRC there has been some updates to the reference solutions recently?) I suppose the PR is merge-ready?

atgeirr:
Yes, Jenkins was successful in opm-simulators so I'll merge.

PR #1116, opened 2016-12-03
AdaptiveTimeStepping: fix stupid (but harmless) mistake in the sub-step info message

andlaus:
that was a copy-and-pasto: newton iterations = linearizations - 1


andlaus:
since this is quite trivial and fixes an obvious mistake, I'll self-merge.

PR #1121, opened 2016-12-16
add missing semicolon to fix the build

andlaus:
some versions of boost seem to be more picky about this than others...


atgeirr:
Thanks, merging.

PR #1126, opened 2016-12-21
ignore the SWATINIT keyword in the EQUIL initializer

andlaus:
this is probably incorrect from the physical point of view but E100 seems to ignore SWATINIT there too, so go for "bug-for-bug compatibility".

Note: Before this can be merged @totto82 should verify that it improves the match of the results for Model 2.


atgeirr:
Now should this be closed or resurrected?

totto82:
It should be closed.

PR #1127, opened 2016-12-29
Fix order of initialization.

atgeirr:
Silences a compiler warning.


andlaus:
since I have to blame myself for this warning and because the fix is rather simple and won't break anything, I'll merge. Thanks for the fix!

PR #1128, opened 2017-01-02
Remove unused Deck function argument.

atgeirr:
Necessary after OPM/opm-material#191.


bska:
jenkins build this with downstreams opm-material=191 please

PR #1129, opened 2017-01-02
Make it optinal to apply SWATINIT

totto82:
The reasoning behind this PR is to make it possible to initialize the case
without SWATINIT in order to compute the same defaulted THPRES values as
Ecl. The initialization needs to be re-computed to account for SWATINIT
in the simulations. But this is already the case in flow_ebos

### DIFF
--- a/opm/core/simulator/initStateEquil_impl.hpp
+++ b/opm/core/simulator/initStateEquil_impl.hpp
@@ -882,6 +882,8 @@ namespace Opm
      * \param[in] props    Property object, pvt and capillary properties are used.
      * \param[in] deck     Simulation deck, used to obtain EQUIL and related data.
      * \param[in] gravity  Acceleration of gravity, assumed to be in Z direction.
+     * \param[in] applySwatInit     Make it possible to not apply SWATINIT even if it
+     *                              is present in the deck
      */
     template<class Grid>
     void initStateEquil(const Grid& grid,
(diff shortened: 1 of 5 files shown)


opm/core/simulator/initStateEquil_impl.hpp
```diff
@@ -882,17 +882,35 @@ namespace Opm
...
-                        BlackoilState& state)
+                        BlackoilState& state,
+                        bool applySwatinit = true)
     {
+
         typedef EQUIL::DeckDependent::InitialStateComputer ISC;
-        ISC isc(props, deck, eclipseState, grid, gravity);
+        //Check for presence of kw SWATINIT
+        std::vector<double> swat_init = {};
+        if (deck.hasKeyword("SWATINIT") && applySwatinit) {
```
joakim-hove: Please - we are trying hard to reduce the usage of `Deck`outside of opm-parser. It is maybe unrealistic to consider `Deck` as a fully internal datastructure in opm-parser, but when the actual property is already internalized in `EclipseState` that should be used. For this particular case the changes could be like:
```C++
const auto& ecl_props = eclipseState.get3DProperties();
if (ecl_props.hasDeckDoubleGridProperty("SWATINIT") && applySwatinit) {
   ...
}
```

pgdr:
No objections to the patch, but this would be a very good time to stop using `deck` and start using `eclipseState.get3DProperties().hasDeck[Double|Int]GridProperty(...)`

opm/core/simulator/initStateEquil_impl.hpp
totto82: Great. I agree on the goal.  I will fix it.

andlaus:
can I merge this? I'd like to get OPM/ewoms#122 in and this seems to be a prerequisite. (note: if you opt to merge this PR yourself, please also merge OPM/ewoms#122 at the same time.)

totto82:
From my point of view. Yes. It is tested and works as expected. Maybe @atgeirr has an opinion since it changes the interface of the initialization code.

atgeirr:
The interface change is ok, but has it been tested? Do we have code that exercises both the true and false branches?

totto82:
I don't find any tests for SWATINIT at all. I will add a simple test.

joakim-hove:
> I will add a simple test.

image

PR #1131, opened 2017-01-06
Increase max iterations for capillary curve inversion.

atgeirr:
Triggered by a new two-phase case.

It is not clear why the scalar solver fails to converge quickly in this case, but as it
 - only affects the initialization
 - can be easily addressed with no effect on any other cases
... it may not be worth spending a lot of time investigating it.


andlaus:
+1 for merging this!

PR #1132, opened 2017-01-09
Throw if solvent phase injected.

atgeirr:
Replaces #1130.


bska:
Looks good.  Merge at will when Travis approves.

PR #1133, opened 2017-01-11
Added an accumulate method that allows to switch off non-owner entries.

blattms:
It does this using a mask vector with entries 0 or 1. If that is not provided
it falls back to std::accumulate.
This will be needed for the parallel flow_ebos.


atgeirr:
Good, merging!

PR #1134, opened 2017-01-12
updating the group control limit for individual control wells

GitPaean:
to provide a better standard for the wells under individual control to
return to group control. For example, some wells get really big group
control limit and switch to individual control, it is very difficult for
them to return to group control with that kind of unreasonable fixed
group limit.

It will be needed for the downstream opm-simulator development for the group control for flow-ebos.


GitPaean:
With the results recovered in PR OPM/opm-simulators#1021, this PR is ready for review now.

atgeirr:
I think this looks good, will review the opm-simulators PR now.

PR #1135, opened 2017-01-13
(Discussion) for WCONHIST wells, there should be one active control there

GitPaean:
For WCONPROD, the rates or bhps that are not under control mode serves as limits.

For WCONHIST, they should not enter well_controls as limit. The rates not under control mode, they are just rates in case we need them to calculate LRAT or RESV.  Basically, the design of WCONHIST should only allow one control there. 

Not sure should there be a BHP control there always, since we can specify BHP control directly. When not specified BHP control, it should not possibly have a BHP limit either.

In the parser, for WCONHIST wells, we add all kinds of control there to keep the rates values.


GitPaean:
There should always be one BHP limit there.

GitPaean:
closed by OPM/opm-parser#1032.

PR #1137, opened 2017-01-25
set the VREP control as current control when applying VREP control

GitPaean:
Not sure it is always the best thing to do here, while it can help the consistence of the two current controls in the well_controls and well_state, and also the individual_control_. 

It will be required by the incoming downstream group control PR in opm-simulators.


atgeirr:
Logic seems fine, merging.

PR #1138, opened 2017-01-26
do not explicitly pass the permeability to the well model anymore

andlaus:
this information is already part of the EclipseState. The reason why this should IMO be avoided is that this enforces an implementation detail (ordering of the permeability matrices) of the simulator on the well model. If this needs to be done for performance reasons, IMO it would be smarter to pass an array of matrices, instead of passing a raw array of doubles.  I doubt that this is necessary, though: completing the full Norne deck takes about 0.25 seconds longer on my machine, that's substantially less than 0.1% of the total runtime.

in order to avoid code duplication, the permeability extraction function of the RockFromDeck class is made a public static function and used as an implementation detail of the WellsManager.

finally, the permfield_valid_ attribute is removed from the RockFromDeck class because this data was unused and not accessible via the class' public API.


andlaus:
this PR must be merged synchronously with OPM/opm-simulators#1042

opm/core/props/rock/RockFromDeck.cpp
```diff
@@ -105,19 +107,20 @@ namespace Opm
         }
     }
 
-    void RockFromDeck::assignPermeability(const Opm::EclipseState& eclState,
-                                          int number_of_cells,
-                                          const int* global_cell,
-                                          const int* cartdims,
-                                          double perm_threshold)
+    void RockFromDeck::extractInterleavedPermeability(std::vector<double>& permeability,
```
atgeirr: Please change the interface as follows:
 - put output arguments last (or consider return by value instead, I think that may be even better)
 - make input arguments const
 - since this method is now public, it must be documented

andlaus:
> Does it? If so, how?

it reduces the dependency on the old fluid object, i.e., the permeabilities do not need to be brought into a format which makes about zero sense in the context of `flow_ebos` just to allow instantiation of the `WellsManager`.

atgeirr:
> it reduces the dependency on the old fluid object

It retains the dependency on the old rock property class though. What is the permeability format used by flow_ebos? And where is it used (after creating transmissibilities, well connection tranmissibility factors and jfunc factors it should not be needed anymore)?

opm/core/props/rock/RockFromDeck.cpp
andlaus: ok, changed. thanks.

andlaus:
> What is the permeability format used by flow_ebos?

it asks the problem for the permeability of a given local context that depends on the used spatial and temporal discretization, i.e. the generic way to query for permeability is `problem.permeability(context, spaceIdx, timeIdx)`. this method returns a `Dune::FieldVector<double, dim, dim>`. `EclProblem` implements this by storing one field vector per cell in a `std::vector`.

andlaus:
> And where is it used

in the context of ´flow_ewoms` it is only used to compute the transmissibilities and by the well model. That's why I'd like to avoid to add (IMO non-obvious) code just to make `WellsManager` happy.

andlaus:
> OK. I assume you meant FieldMatrix?

of course. With this I wanted to test your attention ;)

opm/core/props/rock/RockFromDeck.cpp
atgeirr: Is it changed? I don't see any changes...

andlaus: ops, sorry seems like I forgot the `-f` when pushing. now it should be there...

atgeirr:
I think the tests run previously shows this was working, so I don't think we need to wait for another Jenkins round just because of the minor revision. Merging.

PR #1139, opened 2017-02-06
changed: pass fipnum array into adaptive time stepping loop

akva2:
needed as substep summary reports requires FIP data to be available.
add calculation of this data if output is requested and summary config holds
relevant keywords.

### DIFF
--- a/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
+++ b/opm/core/simulator/AdaptiveTimeStepping_impl.hpp
@@ -152,15 +152,17 @@ namespace Opm {
     SimulatorReport AdaptiveTimeStepping::
     step( const SimulatorTimer& simulatorTimer, Solver& solver, State& state, WellState& well_state )
     {
-        return stepImpl( simulatorTimer, solver, state, well_state );
+        return stepImpl( simulatorTimer, solver, state, well_state,  nullptr, nullptr );
     }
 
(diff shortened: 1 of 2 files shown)


dr-robertk:
Would it be possible to move the AdaptiveTimeStepping to opm-simulators (including history)? It's not used anywhere else and to would avoid having to have separate PR's all the time.

opm/core/simulator/AdaptiveTimeStepping_impl.hpp
```diff
@@ -279,6 +282,8 @@ namespace Opm {
                 // to write it as this will be done by the simulator
                 // anyway.
                 if( outputWriter && !substepTimer.done() ) {
+                    if (fipnum)
```
atgeirr: Braces please.

akva2: done.

atgeirr:
> Would it be possible to move the AdaptiveTimeStepping to opm-simulators (including history)? It's not used anywhere else and to would avoid having to have separate PR's all the time.

Yes, and I agree we should do it. Will have a go as soon as I can find the time (probably not until next week).

akva2:
merging. permission granted in phone call with @atgeirr

atgeirr:
Seems to compile without this, but to get the expected functionality this should be merged with OPM/opm-simulators#1046, so merging it now.

PR #1140, opened 2017-02-07
fix a GCC 7 warning

andlaus:
GCC 7 warns

```
dynamic exception specifications are deprecated in C++11; use ‘noexcept’ instead [-Wdeprecated]
```

here.

on an unreleated but IMO interesting note, GCC 7 seems to become a great release for out purposes: on my machine the linearization performance improved by about 20% using flow_ebos with the norne deck when compared to GCC 5.4.0:

```
================    End of simulation     ===============

Total time taken (seconds):   661.888
Solver time (seconds):        658.491
 Assembly time (seconds):     284.158
 Linear solve time (seconds): 313.216
 Update time (seconds):       9.94308
 Output write time (seconds): 0.00081154
Overall Well Iterations:      890
Overall Linearizations:       1997
Overall Newton Iterations:    1653
Overall Linear Iterations:    24167
```

this figures need to taken with a grain of salt though: I had to disable output because it segfaulted (this is probably related to #1146) and I had to recompile boost because of linker errors. the latter caused me to loose another 3 hours (or so).

### DIFF
--- a/opm/core/utility/MonotCubicInterpolator.cpp
+++ b/opm/core/utility/MonotCubicInterpolator.cpp
@@ -166,7 +166,7 @@ read(const std::string & datafilename, int xColumn, int fColumn)
 
 void
 MonotCubicInterpolator::
-addPair(double newx, double newf) throw(const char*) {
+addPair(double newx, double newf) {
   if (std::isnan(newx) || std::isinf(newx) || std::isnan(newf) || std::isinf(newf)) {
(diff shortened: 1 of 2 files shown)


alfbr:
If that is really the case, it seems Christmas presents came early this year :-)

andlaus:
or rather late for last year ;)

dr-robertk:
It's more like Christmas and Easter together ;-).

andlaus:
@atgeirr: mind pressing green here?

PR #1142, opened 2017-02-13
adding thp to the report for summary output.

GitPaean:


totto82:
Thanks. I have tested it and it seems to works. ie. ready for merge.

PR #1143, opened 2017-02-17
Fixing thp initialization

GitPaean:
When no THP target/limit is involved in the well control, we should not assign a non-zero value to the THP value.  It means, we will always output zero THP value for this well. 

It does not change previously matched results from the THP control case.

### DIFF
--- a/opm/core/simulator/WellState.hpp
+++ b/opm/core/simulator/WellState.hpp
@@ -84,8 +84,8 @@ namespace Opm
                         for (int p = 0; p < np; ++p) {
                             wellrates_[np*w + p] = 0.0;
                         }
-                        bhp_[w] = 0;
-                        thp_[w] = 0;
+                        bhp_[w] = 0.;
+                        thp_[w] = 0.;
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


GitPaean:
Possibly, the PR can keep the `flow_ebos` from running Model 2.2 until the prediction mode (it stops at an early point now). Please hold this PR until further notice.

GitPaean:
Testing show it is nothing related to this PR. The same behavior is observed with the master branch. This PR does not change anything in the residual history until crash. 

Some very recent change  changed the behavior when running `flow_ebos` with model 2.2

```
Time step   80 at day 549/18022, date = 01-Mar-2002
  Substep 0, stepsize 3 days.
    Substep summary: well iterations = 5, newton iterations = 4, linearizations = 5 (9.9369 sec), linear iterations = 41 (5.59008 sec)
flow_ebos: /home/kaib/OPM-devel-test/debug/opm-material/opm/material/common/Spline.hpp:1759: Scalar Opm::Spline<Scalar>::h_(std::size_t) const [with Scalar = double; std::size_t = long unsigned int]: Assertion `x_(i) > x_(i-1)' failed.
Aborted
```

GitPaean:
The residual history before the crash is not changed either compared with the result from old running. Not sure why the running crash with the above message with the updated master branch.

GitPaean:
The crash is related to OPM/opm-simulators#1056 is fixed by OPM/opm-material#205 . It is not related to this PR. This is PR will make the output of the THP correct. When no THP constraint is involved, the THP will output zero.

GitPaean:
This PR should go in to avoid the current confusing THP output in summary.

PR #1144, opened 2017-02-22
Do not set rs=rsSat and rv=rvSat for the saturated case

totto82:
Makes the initial rs and rv values match the ecl values. 
The simulator independently sets Rs=rsSat and rv=rvSat for the saturated case. This should therefore not change any results. Only tested on Model 2 rel-0. 

Is this initialization code used by other simulators than flow and flow_ebos?

### DIFF
--- a/opm/core/simulator/initStateEquil_impl.hpp
+++ b/opm/core/simulator/initStateEquil_impl.hpp
@@ -920,35 +920,6 @@ namespace Opm
         state.gasoilratio() = isc.rs();
         state.rv() = isc.rv();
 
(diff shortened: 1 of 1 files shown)


atgeirr:
> Is this initialization code used by other simulators than flow and flow_ebos?

Not as far as I know. For the model you tested, did it produce identical numbers as before (for the convergence display for example)?

totto82:
Yes, for this case convergence is exactly the same.

atgeirr:
> Yes, for this case convergence is exactly the same.

Then I'll say your claim that it does not change the results holds. Merging!

PR #1145, opened 2017-03-01
Add sat table id to well struct

totto82:
Note: -1 is used if default saturation table ids.

Depends on OPM/opm-parser#1045

### DIFF
--- a/opm/core/wells.h
+++ b/opm/core/wells.h
@@ -91,6 +91,10 @@ struct Wells
      */
     double *WI;
 
+    /**
+     *  Saturation table number , same size and structure as well_cells.
+     */
+    int *sat_table_id;
 
     /**
      * Well controls, one set of controls for each well.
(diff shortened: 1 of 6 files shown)


joakim-hove:
I have not looked at this code, and I guess it is not  on my list to review it either - but I am **extremely** skeptical. When ohh when are we going to kill this structure?

bska:
> When are we going to kill [`struct Wells`]?

When there is a strictly better alternative in place.  The parser's [representation of wells](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Schedule/Well.hpp) is not.

joakim-hove:
Well there is 90% overlap - if the parser representation is not adequate we should improve it instead of contuing to maintain the 90% overlap.

bska:
> if the parser representation is not adequate we should improve it

Fine by me.  The one missing piece is making the parser aware of the simulation grid or, at minimum, the simulation grid's mapping of active cells.  Using `ACTNUM != 0` is not sufficient.

atgeirr:
> When there is a strictly better alternative in place.

Also the work required to change it all is daunting. And might not matter much.

atgeirr:
The code looks fine, I could have missed some C management details but to me it looks correct.

PR #1147, opened 2017-03-16
Do not extrapolate initial rs and rv values in the depth tables

totto82:
Makes initial rs values when outside depth range compatible with Ecl.

### DIFF
--- a/opm/core/utility/linearInterpolation.hpp
+++ b/opm/core/utility/linearInterpolation.hpp
@@ -70,6 +70,22 @@ namespace Opm
 	return  (yv[ix2] - yv[ix1])/(xv[ix2] - xv[ix1])*(x - xv[ix1]) + yv[ix1];
     }
 
+    inline double linearInterpolationNoExtrapolation(const std::vector<double>& xv,
+                                                     const std::vector<double>& yv, double x)
+    {
+        // Return end values if x is outside xv
+        if (x < xv.front()) {
+            return yv.front();
+        }
+        if (x > xv.back()) {
+            return yv.back();
+        }
+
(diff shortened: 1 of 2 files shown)


opm/core/utility/linearInterpolation.hpp
```diff
@@ -70,6 +70,22 @@ namespace Opm
 	return  (yv[ix2] - yv[ix1])/(xv[ix2] - xv[ix1])*(x - xv[ix1]) + yv[ix1];
     }
 
+    inline double linearInterpolationNoExtrapolation(const std::vector<double>& xv,
+                                      const std::vector<double>& yv, double x)
+    {
+        int ix1 = tableIndex(xv, x);
```
atgeirr: Please move these definitions below the end-value tests.

opm/core/utility/linearInterpolation.hpp
```diff
@@ -70,6 +70,22 @@ namespace Opm
 	return  (yv[ix2] - yv[ix1])/(xv[ix2] - xv[ix1])*(x - xv[ix1]) + yv[ix1];
     }
 
+    inline double linearInterpolationNoExtrapolation(const std::vector<double>& xv,
+                                      const std::vector<double>& yv, double x)
```
atgeirr: Weird indentation, please fix.

totto82:
Fixed. And yes it is tested on model 2 and norne.

atgeirr:
I don't really understand the Travis error, so I'll run Jenkins to check this.

atgeirr:
jenkins build this please

atgeirr:
Thanks for the patch, merging as it is green and you already tested it.

PR #1149, opened 2017-03-21
Temporarily disabled EQUIL + SWATINIT test.

joakim-hove:
As describeded here: https://github.com/OPM/opm-core/issues/1148 the PR: https://github.com/OPM/opm-parser/pull/1051 has broken the test of `EQUIL` in decks with `SWATINIT`. In opm-core - I really do not know how to fix this situation, and no one have stepped up to assist yet.

So - in agreement with @atgeirr I will merge this PR temporarily disabling the test if Travis goes green. Hopefully @totto82, @andlaus  or someone else with intimate knowledge of endscaling and `SWATINIT` can resolve the situation properly.

Sorry about the mess.

Issue #1148, opened 2017-03-20
Broken test


joakim-hove:
Ok - I realise now that opm-parser recently broke the EQUIl test here in opm-core, I am sorry about this. Will look into it tomorrow, but I foresee I already now that I might need some assistance. Feel free to merge a revert in opm-parser.

alfbr:
Be aware that the PR fixed a critical bug. I would rather prefer disabling the test than reverting the PR while the investigation is ongoing.

joakim-hove:
@totto82  /  @andlaus : can you look into this; I really do not have a clue whether this is a bug/regression in the code, or if the test needs to be updated; but since this seems to come from the merging of: https://github.com/OPM/opm-parser/pull/1051 you might know how this should be fixed.

totto82:
I had a quick look. The test needs to be updated. Sorry for late replay. I was traveling. Expect a fix tomorrow.

PR #1151, opened 2017-03-22
Fix test equil

totto82:
This should fix the test. 
See issue #1148 
This PR also undo #1149

### DIFF
--- a/tests/capillarySwatinit.DATA
+++ b/tests/capillarySwatinit.DATA
@@ -11,15 +11,17 @@ OIL
 GAS
 
 DIMENS
-40 40 40 /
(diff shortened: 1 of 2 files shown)


bska:

bska:
Just out of interest: How did you derive the new expected values?

totto82:
I have verified the values against Eclipse. They are within a few %. But I used values generated by opm in order to use a stricter tolerance for the test.

atgeirr:
Jenkins is ok, local test is ok, and it only affects a test in opm-core, so I'll merge. Thanks for reinstating the test!

PR #1152, opened 2017-04-04
[WIP] changes needed to compute well potentials each time step. 

GitPaean:
The major reason for this PR is to help to fulfill the computation of the well potentials for each `time step` instead of using the well potentials from the end of `last report step`. As a result, setting the `guide rate` based on `well potentials` can be done in the initialization of the `WellsManager` anymore. 

And also, make the `distr` in `well_controls` recognize the injection phase. 

More description will be found in the PR description in the OPM/opm-simulators#1118.


GitPaean:
TODO: Well potentials related should be removed from the WellsManager.  It will change  the interface through quite a few legacy simulators.

PR #1154, opened 2017-04-10
SimulatorReport: make it possible to report failed time steps

andlaus:
the method can still be called as it used to be. in this case no failures are reported, though.

### DIFF
--- a/opm/core/simulator/SimulatorReport.cpp
+++ b/opm/core/simulator/SimulatorReport.cpp
@@ -70,23 +70,80 @@ namespace Opm
         }
     }
 
-    void SimulatorReport::reportFullyImplicit(std::ostream& os)
+    void SimulatorReport::reportFullyImplicit(std::ostream& os, const SimulatorReport* failureReport)
     {
         if ( verbose_ )
         {
-            os << "Total time taken (seconds):   " << total_time << "\n"
-               << "Solver time (seconds):        " << solver_time << "\n";
+            double t = total_time;
+            os << "Total time (seconds):         " << t;
+            os << std::endl;
+
+            t = solver_time + (failureReport ? failureReport->solver_time : 0.0);
+            os << "Solver time (seconds):        " << t;
+            os << std::endl;
+
             if (assemble_time > 0.0 || linear_solve_time > 0.0) {
-                os << " Assembly time (seconds):     " << assemble_time << "\n"
-                   << " Linear solve time (seconds): " << linear_solve_time  << "\n"
-                   << " Update time (seconds):       " << update_time  << "\n"
-                   << " Output write time (seconds): " << output_write_time  << "\n";
+                t = assemble_time + (failureReport ? failureReport->assemble_time : 0.0);
+                os << " Assembly time (seconds):     " << t;
+                if (failureReport) {
+                    os << " (Failed: " << failureReport->assemble_time << "; "
+                       << 100*failureReport->assemble_time/t << "%)";
+                }
+                os << std::endl;
+
+                t = linear_solve_time + (failureReport ? failureReport->linear_solve_time : 0.0);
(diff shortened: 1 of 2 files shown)


opm/core/simulator/SimulatorReport.hpp
```diff
@@ -53,7 +53,7 @@ namespace Opm
         /// Print a report to the given stream.
         void report(std::ostream& os);
         /// Print a report, leaving out the transport time.
-        void reportFullyImplicit(std::ostream& os);
+        void reportFullyImplicit(std::ostream& os, const SimulatorReport* failedReport = nullptr);
```
atgeirr: The pointer + default argument approach is a little cludgy, but it works. Good enough for now! I think we should find something more elegant when it gets annoying...

opm/core/simulator/SimulatorReport.cpp
```diff
@@ -70,23 +70,73 @@ namespace Opm
...
     {
         if ( verbose_ )
         {
-            os << "Total time taken (seconds):   " << total_time << "\n"
-               << "Solver time (seconds):        " << solver_time << "\n";
+            double t = total_time;
+            os << "Total time (seconds):         " << t;
+            os << std::endl;
+
+            t = solver_time + (failureReport?failureReport->solver_time:0.0);
```
atgeirr: Spaces around the `?` and `:` please, I actually like the trinary but find it hard to read this way. (Also further down.)

opm/core/simulator/SimulatorReport.cpp
```diff
@@ -70,23 +70,73 @@ namespace Opm
...
+            os << std::endl;
+
             if (assemble_time > 0.0 || linear_solve_time > 0.0) {
-                os << " Assembly time (seconds):     " << assemble_time << "\n"
-                   << " Linear solve time (seconds): " << linear_solve_time  << "\n"
-                   << " Update time (seconds):       " << update_time  << "\n"
-                   << " Output write time (seconds): " << output_write_time  << "\n";
+                t = assemble_time + (failureReport?failureReport->assemble_time:0.0);
+                os << " Assembly time (seconds):     " << t;
+                if (failureReport)
```
atgeirr: Braces please. (Also further down.)

opm/core/simulator/SimulatorReport.cpp
```diff
@@ -70,23 +70,73 @@ namespace Opm
...
+                os << " Output write time (seconds): " << t;
+                os << std::endl;
 
             }
-            os << "Overall Well Iterations:      " << total_well_iterations << "\n"
-               << "Overall Linearizations:       " << total_linearizations << "\n"
-               << "Overall Newton Iterations:    " << total_newton_iterations << "\n"
-               << "Overall Linear Iterations:    " << total_linear_iterations << "\n";
+
+            int n = total_well_iterations + (failureReport?failureReport->total_well_iterations:0);
```
atgeirr: For the following reports, I'd rather have failed counts than percentages (or both, but not only percentages).

andlaus: it already prints both, doesn't it?

andlaus:
ok, I hope to have addressed the issues you've raised in the latest version. thanks for the review.

opm/core/simulator/SimulatorReport.hpp
andlaus: I concur. Besides this I'm a bit annoyed by the `if` jungle introduced by this patch, but I could not come up with a better solution...

opm/core/simulator/SimulatorReport.cpp
atgeirr: Oh, missed that. Don't do reviews late at night I guess...

PR #1155, opened 2017-04-18
Packinging update backport

akva2:
- Add new dependencies (opm-grid related)
- Remove binary package (no binaries produced by opm-core any longer).


atgeirr:
> Remove binary package (no binaries produced by opm-core any longer).

What about the library?

akva2:
that is of course produced. binary = executables, sorry for being imprecise.

PR #1156, opened 2017-04-20
FIX memory leakage in well_controls

totto82:
The Alq and vfp is set free in well_controls_destroy to avoid memory
leakage. Valgrind stopped complaining after this fix. 

@and Can you verify that this fixes the memory issue you reported.


atgeirr:

andlaus:
I can confirm that with this patch address sanitizer no longer complains for SPE1 on my machine.

PR #1157, opened 2017-04-20
fix non-MPI builds

andlaus:
some headers must be included unconditionally.

this patch should also go into the release branch.


andlaus:
jenkins build this please

PR #1163, opened 2017-05-18
Using ppa for newer cmake version on travis.

joakim-hove:


atgeirr:
I assume this should be merged? @akva2 or @joakim-hove feel free to press the green button.

atgeirr:
Rerunning travis, this failed. Not sure if it is worth to pursue further? @joakim-hove please weigh in.

akva2:
afaict it's caused by using a backported compiler with a boost built with an older ABI.

atgeirr:
> afaict it's caused by using a backported compiler with a boost built with an older ABI.

Sounds like a hornet's nest!

akva2:
since travis has been shut down, i'll close this.

PR #1170, opened 2017-07-13
make dune >= 2.4 mandatory

andlaus:
this Dune release is now two years old, so it should not be a major concern to upgrade. Note that there even was some compatibility code for Dune 2.2. The patch deals with this as well.


atgeirr:
jenkins build this please

andlaus:
this is obsolete. closing.

PR #1172, opened 2017-08-04
Add interface for passing cell pressures directly in wellState.init(...)

totto82:
Useful if you don't have a ReservoirState


totto82:

atgeirr:
Merging this first, as it does not require any simultaneous changes.

PR #1177, opened 2017-10-03
Removing redundant checks from Nexus

alfbr:
Two checks were erroneously entered, based on a mix-up of which saturation to check. These are now removed. Simply run flow on the Norne deck to confirm.

### DIFF
--- a/opm/core/props/satfunc/RelpermDiagnostics_impl.hpp
+++ b/opm/core/props/satfunc/RelpermDiagnostics_impl.hpp
@@ -33,7 +33,7 @@ namespace Opm {
                                        const Opm::Deck& deck,
                                        const GridT& grid)
     {
-        OpmLog::info("\n***************Saturation Functions Diagnostics***************");
+        OpmLog::info("\n===============Saturation Functions Diagnostics===============");
         phaseCheck_(eclState);
         satFamilyCheck_(eclState);
         tableCheck_(eclState);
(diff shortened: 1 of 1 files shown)


atgeirr:
> Simply run flow on the Norne deck to confirm.

I don't understand what that would confirm? Are these (wrong) checks among those triggered by the Norne deck?

alfbr:
Yes, they are wrong checks. A run of Norne without this patch would yield:

Warning: For scaled endpoints input, cell(5, 10, 2) SATNUM = 1, SWCR > SOWCR

But this is based on a misread of the Nexus manual. It should be SWCR>(1-SOWCR) which is already covered by the check SOWCR+SWCR>=1.0.

Please note that the PR now also includes a tolerance. We have had false positives from, so errors now need to exceed 1e-6 to emit warning.

atgeirr:
There are 27114 such warnings for Norne right now... If all of those are removed that is nice!

If there is no hurry I'll hold off merging things like this until Jenkins is fully functional. Otherwise, it looks benign enough.

alfbr:
Sure, there is no hurry. Yes, many of them will go away :-)

alfbr:
ping

atgeirr:
jenkins build this with downstreams run norne please

atgeirr:
All ok, merging.

PR #1178, opened 2017-10-04
mark initHydroCarbonState() as inline

andlaus:
this allows it to be used in multiple compile units without the linker running amok.


atgeirr:
Good catch, thanks.

PR #1180, opened 2017-10-11
[cleanup] Removes unnecessary if clauses for unsupported DUNE versions.

blattms:
We are targetting DUNE 2.4.* and 2.5.* currently. Therefore this commit removes
the if checks for lower versions to cleanup the code.


blattms:
jenkins run triggered on OPM/opm-grid#286

PR #1181, opened 2017-10-12
Switch to dune buildsystem

andlaus:
this is the opm-core part of OPM/opm-common#277. please do the generic discussion about it there, i.e., only discuss the opm-core specific issues here.


andlaus:
this PR will soon be replaced by a "more additive" one. closing.

PR #1187, opened 2017-11-20
Remove black oil props from init

totto82:
This PR untangle the EQUIL initialization from opm-core legacy property, blackoilState and blackoilPhase objects.

### DIFF
--- a/opm/core/simulator/EquilibrationHelpers.hpp
+++ b/opm/core/simulator/EquilibrationHelpers.hpp
@@ -538,8 +538,7 @@ namespace Opm
             /**
              * Retrieve pvtIdx of the region.
              */
-            const int
-            pvtIdx() const { return this->pvtIdx_; }
+            int pvtIdx() const { return this->pvtIdx_; }
 
 
         private:
@@ -566,27 +565,29 @@ namespace Opm
                   cell_(cell),
                   target_pc_(target_pc)
             {
-                fluidState_.setSaturation(FluidSystem::waterPhaseIdx, 0.0);
-                fluidState_.setSaturation(FluidSystem::oilPhaseIdx, 0.0);
-                fluidState_.setSaturation(FluidSystem::gasPhaseIdx, 0.0);
-                std::fill(pc_, pc_ + FluidSystem::numPhases, 0.0);
+
             }
             double operator()(double s) const
             {
                 const auto& matParams = materialLawManager_.materialLawParams(cell_);
-                fluidState_.setSaturation(phase_, s);
-                MaterialLaw::capillaryPressures(pc_, matParams, fluidState_);
+                SatOnlyFluidState fluidState;
+                fluidState.setSaturation(FluidSystem::waterPhaseIdx, 0.0);
+                fluidState.setSaturation(FluidSystem::oilPhaseIdx, 0.0);
+                fluidState.setSaturation(FluidSystem::gasPhaseIdx, 0.0);
+                fluidState.setSaturation(phase_, s);
+
+                double pc[FluidSystem::numPhases];
+                std::fill(pc, pc + FluidSystem::numPhases, 0.0);
+                MaterialLaw::capillaryPressures(pc, matParams, fluidState);
                 double sign = (phase_ == FluidSystem::waterPhaseIdx)? -1.0 : 1.0;
-                double pc = pc_[FluidSystem::oilPhaseIdx] + sign *  pc_[phase_];
-                return pc - target_pc_;
+                double pcPhase = pc[FluidSystem::oilPhaseIdx] + sign *  pc[phase_];
+                return pcPhase - target_pc_;
             }
         private:
             const MaterialLawManager& materialLawManager_;
             const int phase_;
             const int cell_;
             const double target_pc_;
-            mutable SatOnlyFluidState fluidState_;
-            mutable double pc_[FluidSystem::numPhases];
         };
 
         template <class FluidSystem, class MaterialLawManager>
@@ -694,22 +695,25 @@ namespace Opm
                   cell_(cell),
                   target_pc_(target_pc)
             {
-                fluidState_.setSaturation(FluidSystem::waterPhaseIdx, 0.0);
-                fluidState_.setSaturation(FluidSystem::oilPhaseIdx, 0.0);
-                fluidState_.setSaturation(FluidSystem::gasPhaseIdx, 0.0);
-                std::fill(pc_, pc_ + FluidSystem::numPhases, 0.0);
             }
             double operator()(double s) const
             {
                 const auto& matParams = materialLawManager_.materialLawParams(cell_);
-                fluidState_.setSaturation(phase1_, s);
-                fluidState_.setSaturation(phase2_, 1.0 - s);
+                SatOnlyFluidState fluidState;
+                fluidState.setSaturation(FluidSystem::waterPhaseIdx, 0.0);
+                fluidState.setSaturation(FluidSystem::oilPhaseIdx, 0.0);
+                fluidState.setSaturation(FluidSystem::gasPhaseIdx, 0.0);
+                fluidState.setSaturation(phase1_, s);
+                fluidState.setSaturation(phase2_, 1.0 - s);
 
-                MaterialLaw::capillaryPressures(pc_, matParams, fluidState_);
+                double pc[FluidSystem::numPhases];
+                std::fill(pc, pc + FluidSystem::numPhases, 0.0);
+
+                MaterialLaw::capillaryPressures(pc, matParams, fluidState);
                 double sign1 = (phase1_ == FluidSystem::waterPhaseIdx)? -1.0 : 1.0;
-                double pc1 = pc_[FluidSystem::oilPhaseIdx] + sign1 *  pc_[phase1_];
+                double pc1 = pc[FluidSystem::oilPhaseIdx] + sign1 *  pc[phase1_];
                 double sign2 = (phase2_ == FluidSystem::waterPhaseIdx)? -1.0 : 1.0;
-                double pc2 = pc_[FluidSystem::oilPhaseIdx] + sign2 *  pc_[phase2_];
+                double pc2 = pc[FluidSystem::oilPhaseIdx] + sign2 *  pc[phase2_];
                 return pc1 + pc2 - target_pc_;
             }
         private:
@@ -718,8 +722,6 @@ namespace Opm
             const int phase2_;
             const int cell_;
             const double target_pc_;
-            mutable SatOnlyFluidState fluidState_;
-            mutable double pc_[FluidSystem::numPhases];
         };
 
 
--- a/opm/core/simulator/EquilibrationHelpers.hpp
+++ b/opm/core/simulator/EquilibrationHelpers.hpp
@@ -1,5 +1,6 @@
 /*
   Copyright 2014 SINTEF ICT, Applied Mathematics.
+  Copyright 2017 IRIS
 
   This file is part of the Open Porous Media project (OPM).
 
@@ -20,7 +21,6 @@
 #ifndef OPM_EQUILIBRATIONHELPERS_HEADER_INCLUDED
 #define OPM_EQUILIBRATIONHELPERS_HEADER_INCLUDED
 
-#include <opm/core/props/BlackoilPhases.hpp>
 #include <opm/core/utility/linearInterpolation.hpp>
 #include <opm/core/utility/RegionMapping.hpp>
 #include <opm/core/utility/RootFinders.hpp>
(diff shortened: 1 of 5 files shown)


bska:
> This PR untangle the EQUIL initialization

I'm not sure "untangle" is the word I'd use for this.  The word "switch" is definitely more descriptive.

The feature set looks okay, but the implementation is not.  I'm not going to merge this in its current state.  There are too many `shared_ptr<X>` parameters that should be `const X&` instead and rely on nested lifetimes.  Moreover, the "always three-phase" behaviour is a step backwards.  Why is that necessary?  Will we reenable two-phase support at a later time?  If not, why not?

Add to this a litany of white-space changes that only obscure the change-set and using [`->template`](https://github.com/OPM/opm-core/pull/1187/commits/42847d37c7b9800d7d984aac1975de1d9aecf8bd#diff-269066f88a1ce3baa7cdcedd29186fe4R105) outside of dependent names in addition to owning raw pointers that should be smart pointers.

In short, this proposal needs additional work before it is suitable for inclusion.

totto82:
> There are too many shared_ptr<X> parameters that should be const X&

You mean the MaterialLawManager. I just followed what's the practice in the rest of the code.  

>Moreover, the "always three-phase" behaviour is a step backwards. Why is that necessary? Will we reenable two-phase support at a later time? If not, why not?

It works perfectly for 2p cases. It is just that the size of the pressure and saturations vectors are always 3 in order to use the fixed phaseIdx from opm-material. I don't think this will be very time critical anyway. 

>Add to this a litany of white-space changes that only obscure the change-set and using ->template outside of dependent names in addition to owning raw pointers that should be smart pointers

I will clean this up.

opm/core/simulator/EquilibrationHelpers.hpp
```diff
@@ -662,70 +549,133 @@ namespace Opm
...
         private:
-            const BlackoilPropertiesInterface& props_;
+
+            std::shared_ptr<MaterialLawManager> materialLawManager_;
             const int phase_;
             const int cell_;
             const double target_pc_;
-            mutable double s_[BlackoilPhases::MaxNumPhases];
-            mutable double pc_[BlackoilPhases::MaxNumPhases];
+            mutable SatOnlyFluidState fluidState_;
```
andlaus: why does this need to mutable?

opm/core/simulator/EquilibrationHelpers.hpp
```diff
@@ -662,70 +549,133 @@ namespace Opm
...
+            mutable double pc_[FluidSystem::numPhases];
```
andlaus: this attribute looks superfluous to me

opm/core/simulator/EquilibrationHelpers.hpp
```diff
@@ -747,37 +697,47 @@ namespace Opm
...
-            const BlackoilPropertiesInterface& props_;
+            std::shared_ptr<MaterialLawManager> materialLawManager_;
             const int phase1_;
             const int phase2_;
             const int cell_;
             const double target_pc_;
-            mutable double s_[BlackoilPhases::MaxNumPhases];
-            mutable double pc_[BlackoilPhases::MaxNumPhases];
+            mutable SatOnlyFluidState fluidState_;
+            mutable double pc_[FluidSystem::numPhases];
```
andlaus: same as above.

opm/core/simulator/initStateEquil_impl.hpp
```diff
@@ -176,45 +165,42 @@ namespace Opm
...
+                    double rs = rs_(depth, press, temp_);
+                    double bOil = 0.0;
+                    if ( !FluidSystem::enableDissolvedGas() || rs >= FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_, temp_, press) ) {
+                        bOil = FluidSystem::oilPvt().saturatedInverseFormationVolumeFactor(pvtRegionIdx_, temp_, press);
+                    } else {
+                        bOil = FluidSystem::oilPvt().inverseFormationVolumeFactor(pvtRegionIdx_, temp_, press, rs);
+                    }
+                    double rho = bOil * FluidSystem::referenceDensity(FluidSystem::oilPhaseIdx, pvtRegionIdx_);
+                    if (FluidSystem::enableDissolvedGas()) {
+                        rho += rs * bOil * FluidSystem::referenceDensity(FluidSystem::gasPhaseIdx, pvtRegionIdx_);
```
andlaus: this can be possibly simplified using `FluidSystem::density()`. That requires a fluid state where pressure, temperature and the phase compositions are already set, though.

bska:
>> There are too many `shared_ptr<X>` parameters that should be `const X&`
>
> You mean the `MaterialLawManager`.

I do.

> I just followed what's the practice in the rest of the code.

Then this is an opportunity to improve upon the existing practice.  An interface that accepts a `shared_ptr<>` explicitly states that it wants to take part in the [lifetime management](https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/) of the underlying object.  The majority of this code does not need to do that.

>> Moreover, the "always three-phase" behaviour is a step backwards. Why is that necessary? Will we reenable two-phase support at a later time? If not, why not?
>
> It works perfectly for 2p cases. It is just that the size of the pressure and saturations vectors are always 3 in order to use the fixed `phaseIdx` from opm-material. I don't think this will be very time critical anyway.

Is that an internal implementation detail or does it leak out to callers?  Always carrying a three-phase saturation vector is definitely not a trivial cost.

opm/core/simulator/initStateEquil.hpp
```diff
@@ -161,12 +142,12 @@ namespace Opm
...
          *                            one saturation value per cell in the region.
          */
-        template <class Grid, class Region, class CellRange>
+        template <class FluidSystem, class Grid, class Region, class CellRange, class MaterialLawManager>
         std::vector< std::vector<double> >
         phaseSaturations(const Grid&             grid,
                          const Region&           reg,
                          const CellRange&        cells,
-                         BlackoilPropertiesFromDeck& props,
+                         std::shared_ptr<MaterialLawManager> materialLawManager,
```
atgeirr: About the shared_ptr issue, I guess this line is typical of what @bska wants to see gone. I agree with him, this essentially says that the function needs to preserve (by sharing ownership to) the MaterialLawManager, possibly beyond the running of the function itself. It also takes a non-const object, so it can modify the MaterialLawManager, which probably is not what is needed (although I see that the old version took a non-const reference to the props object, I don't know why but if there was a good reason it could apply to the new code too).

Passing by const reference should be the default unless we require otherwise. Remember: the parser was in a similar situation, where most of the time shared_ptrs were used. Now those are only used where necessary, and usually const references are used.

totto82: This is where SWATINIT is applied so it can not be const. 
I can double-check that const is added whenever it is possible in the rest of the code . 
I can also change to passing a references instead of shared_ptr if that is preferred.

atgeirr: > This is where SWATINIT is applied so it can not be const.

Aha, thanks for the clarification. Does this have to be simultaneous though? Is it possible to first compute the saturations (with const material laws) and then *afterwards* explicitly update it for swatinit? That would make it a lot less "magical" I think.

As to const and references: yes please and yes please!

opm/core/simulator/initStateEquil_impl.hpp
totto82: I chose this way to avoid creating the fluid state every time I call this function, but maybe you are right. Depending on a explicit test rs >= rsSat to determine if the system is saturated or not may not be a good idea if we decide to change this logic in the future. I will try to rewrite and see how that turns out.

opm/core/simulator/EquilibrationHelpers.hpp
totto82: The operator() is const

totto82:
> Is that an internal implementation detail or does it leak out to callers? Always carrying a three-phase saturation vector is definitely not a trivial cost.

Partly. The caller needs to initializer whatever state object it uses with the correct number of phases. Both Flow and Flow-legacy only carries around the needed phases. (i.e. 2 for 2p problems) 

I changed the shared pointers to ref. as you suggested.  

> and using ->template outside of dependent names in addition to owning raw pointers that should be smart pointers.

Could you be more specific on this?

bska:
>> and using ->template outside of dependent names in addition to owning raw pointers that should be smart pointers.
>
> Could you be more specific on this?

I mean that the `->template` syntax, or its brethren `::template` and `.template`, exist for the purpose of invoking templated identifiers (member functions, member types &c) on objects that depend on a template parameter.  They are not needed to invoke any arbitrary templated member function regardless of context and we generally should not use that syntax outside the implementation of a template (e.g., a function template).

As for owning raw pointers, your [`initDefaultFluidSystem()`](https://github.com/OPM/opm-core/pull/1187/files#diff-269066f88a1ce3baa7cdcedd29186fe4R71) function contains multiple instances of
```C++
X* p = new X(args);
```
when
```C++
auto p = make_shared<X>(args);
```
serves the same purpose in a much safer way, especially since the pointers are subsequently put into `shared_pointer<>` containers anyway.  I think you should apply the patch below and, while you're there, also declare the function `static` (or preferably put it into an anonymous namespace) to address the warnings concerning
```
"no previous declaration for ‘void initDefaultFluidSystem()’ [-Wmissing-declarations]"
```
By the way you should also remove the `const` qualifier in the [return type](https://github.com/OPM/opm-core/pull/1187/files#diff-28e715e320e8b5e39090db6aaa0fd050R541) of
```C++
const int Opm::EQUIL::EquilReg::pvtIdx() const;
```
Returning a `const int` serves no purpose.

---

```diff
diff --git a/tests/test_equil.cpp b/tests/test_equil.cpp
index 3c40a287..b41fd0da 100644
--- a/tests/test_equil.cpp
+++ b/tests/test_equil.cpp
@@ -98,25 +98,25 @@ void initDefaultFluidSystem() {
     FluidSystem::setEnableVaporizedOil(false);
     FluidSystem::setReferenceDensities(rhoRefO, rhoRefW, rhoRefG, /*regionIdx=*/0);
 
-    Opm::GasPvtMultiplexer<double> *gasPvt = new Opm::GasPvtMultiplexer<double>;
+    auto gasPvt = std::make_shared<Opm::GasPvtMultiplexer<double>>();
     gasPvt->setApproach(Opm::GasPvtMultiplexer<double>::DryGasPvt);
-    auto& dryGasPvt = gasPvt->template getRealPvt<Opm::GasPvtMultiplexer<double>::DryGasPvt>();
+    auto& dryGasPvt = gasPvt->getRealPvt<Opm::GasPvtMultiplexer<double>::DryGasPvt>();
     dryGasPvt.setNumRegions(/*numPvtRegion=*/1);
     dryGasPvt.setReferenceDensities(/*regionIdx=*/0, rhoRefO, rhoRefG, rhoRefW);
     dryGasPvt.setGasFormationVolumeFactor(/*regionIdx=*/0, Bg);
     dryGasPvt.setGasViscosity(/*regionIdx=*/0, mug);
 
-    Opm::OilPvtMultiplexer<double> *oilPvt = new Opm::OilPvtMultiplexer<double>;
+    auto oilPvt = std::make_shared<Opm::OilPvtMultiplexer<double>>();
     oilPvt->setApproach(Opm::OilPvtMultiplexer<double>::DeadOilPvt);
-    auto& deadOilPvt = oilPvt->template getRealPvt<Opm::OilPvtMultiplexer<double>::DeadOilPvt>();
+    auto& deadOilPvt = oilPvt->getRealPvt<Opm::OilPvtMultiplexer<double>::DeadOilPvt>();
     deadOilPvt.setNumRegions(/*numPvtRegion=*/1);
     deadOilPvt.setReferenceDensities(/*regionIdx=*/0, rhoRefO, rhoRefG, rhoRefW);
     deadOilPvt.setInverseOilFormationVolumeFactor(/*regionIdx=*/0, Bo);
     deadOilPvt.setOilViscosity(/*regionIdx=*/0, muo);
 
-    Opm::WaterPvtMultiplexer<double> *waterPvt = new Opm::WaterPvtMultiplexer<double>;
+    auto waterPvt = std::make_shared<Opm::WaterPvtMultiplexer<double>>();
     waterPvt->setApproach(Opm::WaterPvtMultiplexer<double>::ConstantCompressibilityWaterPvt);
-    auto& ccWaterPvt = waterPvt->template getRealPvt<Opm::WaterPvtMultiplexer<double>::ConstantCompressibilityWaterPvt>();
+    auto& ccWaterPvt = waterPvt->getRealPvt<Opm::WaterPvtMultiplexer<double>::ConstantCompressibilityWaterPvt>();
     ccWaterPvt.setNumRegions(/*numPvtRegions=*/1);
     ccWaterPvt.setReferenceDensities(/*regionIdx=*/0, rhoRefO, rhoRefG, rhoRefW);
     ccWaterPvt.setViscosity(/*regionIdx=*/0, 1);
```

totto82:
@bska Thanks for the explanation. I will try it out tomorrow.

opm/core/simulator/EquilibrationHelpers.hpp
andlaus: sure, but there is no real reason not to make `pc[BlackoilPhases::MaxNumPhases]` a local variable there?

andlaus: (same for the `fluidState`)

totto82: Good point. I will fix

PR #1188, opened 2017-11-21
Handle WEFAC

totto82:
Added test combining WEFAC and GEFAC

Depends on OPM/opm-parser#1156

### DIFF
--- a/tests/test_wellcollection.cpp
+++ b/tests/test_wellcollection.cpp
@@ -78,3 +78,42 @@ BOOST_AUTO_TEST_CASE(AddWellsAndGroupToCollection) {
     BOOST_CHECK_EQUAL("G2", collection.findNode("PROD2")->getParent()->name());
 }
 
+BOOST_AUTO_TEST_CASE(EfficiencyFactor) {
+    Parser parser;
+    std::string scheduleFile("wells_group.data");
+    ParseContext parseContext;
+    Deck deck = parser.parseFile(scheduleFile, parseContext);
+    EclipseState eclipseState(deck, parseContext);
+    PhaseUsage pu = phaseUsageFromDeck(eclipseState);
+    const auto& grid = eclipseState.getInputGrid();
+    const TableManager table ( deck );
+    const Eclipse3DProperties eclipseProperties ( deck , table, grid);
+    const Schedule sched(deck, grid, eclipseProperties, Phases(true, true, true), parseContext );
+
+    size_t timestep = 2;
+    WellCollection collection;
+    // Add groups to WellCollection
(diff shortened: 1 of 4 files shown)


totto82:

totto82:
Any reason for this not being merged?

atgeirr:
Slipped my mind, not reviewed. @GitPaean can you take a look at it?

tests/test_wellcollection.cpp
```diff
@@ -78,3 +78,42 @@ BOOST_AUTO_TEST_CASE(AddWellsAndGroupToCollection) {
     BOOST_CHECK_EQUAL("G2", collection.findNode("PROD2")->getParent()->name());
 }
 
+BOOST_AUTO_TEST_CASE(EfficiencyFactor) {
+    Parser parser;
+    std::string scheduleFile("wells_group.data");
```
GitPaean: Both tests have some `const` variables without adding `const`. It can be better to add `const` to make it clear they are `const`.

GitPaean:
@totto82 , have you verified against ECL with a simulation? The keyword is not complicated, while it can be good to verify it before it goes in.  Maybe you can add `WEFAC`  to one or two wells in `spe9group/SPE9_CP_GROUP.DATA` in opm-data to protect this functionality.

totto82:
> The keyword is not complicated

I am not sure if I agree. The interaction between efficiency factors and group control is hard to understand. I have tested the WEFAC alone without group control and it seems to work fine. When combined with group control it doesn't. I did have a look in the wellsGroup.*pp and wellCollection.*pp files but got lost...

GitPaean:
Sorry about the misleading comments. 

Would you like to share a result with modified `spe9group/SPE9_CP_GROUP.DATA`, so we can try to fix it together?   I have never had a case with `WEFAC`, that is why it is not handled when implementing `GEFAC`. 

If yes, I can try to modify the SPE9 case a little bit and send to you? Hopefully `WEFAC` does not conflict with `GEFAC` in a bad way. 

And also, it is possible the group  control is buggy.

totto82:
I have a test case that fails based on SPE9_GROUP. It is not the WEFAC <-> GEFAC interaction that is buggy. It is the WEFAC + Group controlled wells. I think WEFAC is supposed to be applied to the well production rate before summed for group rates.  Lets discuss more on the hangout tomorrow.

GitPaean:
> I think WEFAC is supposed to be applied to the well production rate before summed for group rates. Lets discuss more on the hangout tomorrow.

Sure. There can be some tricky parts there. If the ECL result easy to understand, we can fix it to make the results straight.

totto82:

PR #1190, opened 2017-11-27
Fix pvtIndex map for cases with non-active eqlnum regions.

totto82:


opm/core/simulator/initStateEquil.hpp
```diff
@@ -371,13 +372,12 @@ namespace Opm
...
                 void setRegionPvtIdx(const Grid& G, const Opm::EclipseState& eclipseState, const RMap& reg) {
 
-                    regionPvtIdx_.reserve(reg.activeRegions().size());
                     std::vector<int> cellPvtRegionIdx;
                     extractPvtTableIndex(cellPvtRegionIdx, eclipseState, UgGridHelpers::numCells(G), UgGridHelpers::globalCell(G));
                     for (const auto& r : reg.activeRegions()) {
                         const auto& cells = reg.cells(r);
                         const int cell = *(cells.begin());
-                        regionPvtIdx_[r] = cellPvtRegionIdx[cell];
+                        regionPvtIdx_[reg.region(cell)] = cellPvtRegionIdx[cell];
```
bska: How is `reg.region(cell)` different from `r` in this case?

opm/core/simulator/initStateEquil.hpp
```diff
@@ -263,6 +263,7 @@ namespace Opm
                     const auto& tables = eclipseState.getTableManager();
                     // Create (inverse) region mapping.
                     const RegionMapping<> eqlmap(equilnum(eclipseState, G));
+                    regionPvtIdx_.resize(rec.size());
```
atgeirr: Please resize the elements to something invalid, such as -1.

andlaus: do you mean s/resize/initialize/ ?

atgeirr: Yes I do...

```
const int invalidRegion = -1;
regionPvtIdx_.resize(rec.size(), invalidRegion);
```

opm/core/simulator/initStateEquil.hpp
totto82: It doesn't. I miss read the region code.

opm/core/simulator/initStateEquil.hpp
totto82: Fixed.

totto82:
I have verified that the current version solves the problem. I am running a full test of the PR just to be sure that the refactoring of the init stuff doesn't have any other side effects.

totto82:
I can also verify that all the big models are fine with this fix.

atgeirr:

PR #1191, opened 2017-12-04
remove utility files that have moved to opm-common

akva2:
Upstream: https://github.com/OPM/opm-common/pull/293


akva2:
need splitting as sources go to various repos.

PR #1195, opened 2017-12-05
fixed: add ENERGY to switch to quell unhandled enum value warning 

akva2:


akva2:

bska:
While the warning is annoying, it actually serves a purpose (notification of a case that's not handled).  Adding a `default:` clause prevents that for all future cases.

I'd much prefer something along the lines of an `OPM_THROW` inspired by the solvent and polymer cases.

akva2:
the idea was that it would fall through to the generic throw beneath but if you think a throw with "Energy injector not supported" makes more sense that is what you will get.

bska:
> the idea was that it would fall through to the generic throw beneath

That's fine, I don't particularly care about the specifics of the message.  I do however care that the explicit `default:` will prevent the "unhandled case" message for all future instances of this particular problem.

akva2:

bska:
> jenkins build this with downstreams please

Jenkins approves of the change and so do I.  I will merge into master.  Thanks a lot.

PR #1199, opened 2017-12-08
Add dune buildsystem

andlaus:
this is a downstream PR of OPM/opm-common#295. please discuss issues that are specific to opm-core here and issues related to the change in general there.


andlaus:
closing because it there is no hope in getting this merged. #1200 salvages some of the bits from this wreck.

PR #1200, opened 2017-12-13
Some fixes

andlaus:
these contains a few fixes for the some non-buildsystem issues that emerged during #1199.


andlaus:

andlaus:
jenkins and even travis are happy with this. can it be merged?

examples/compute_eikonal_from_files.cpp
```diff
@@ -19,7 +19,7 @@
 
 
 
-#if HAVE_CONFIG_H
+#ifdef HAVE_CONFIG_H
 #include "config.h"
```
dr-robertk: I agree. But if for some reason -DHAVE_CONFIG_H=0 is passed then the new version would fail, right?

andlaus: it would fail, yes. most projects seem to simply use `-DHAVE_CONFIG_H`, though. I think this is a bit of an unfortunate situation. IMO the best way to fix this is to simply include `config.h` unconditionally.

opm/core/linalg/LinearSolverPetsc.cpp
```diff
@@ -19,6 +19,9 @@
 */
 
 #include "config.h"
```
andlaus: notice that here, amongst quite a few other places in opm-core, there is no guard around `config.h`

dr-robertk: Yep, quite the inconsistency. So it's ok with me then.

dr-robertk:
This patch seems ok from my side, especially the fixes in opm/core/linalg/LinearSolverPetsc.cpp and the UMFPACK file should be merged. After those are merged the "hack" in CMakeLists.txt which removes these files in case the packages are not found can be removed.

andlaus:
> After those are merged the "hack" in CMakeLists.txt which removes these files in case the packages are not found can be removed.

Okay, I've added a patch to this PR which does this...

andlaus:
jenkins build this please

andlaus:
@atgeirr: Jenkins and travis are happy and I'm quite sure that it fixes real shenangians. can it be merged?

dr-robertk:
Seems like nobody is objecting and since this are fixes of real issues: Merging.

PR #1202, opened 2017-12-26
Removed travis integration

joakim-hove:


atgeirr:
I agree that Travis is not so useful now. But is there something else we need to do to "turn it off" as well as removing the configuration here?

Feel free to merge (all of these) when ready.

PR #1207, opened 2018-01-11
Utility fixes

atgeirr:
Two minor fixes required before the utilities in question can be moved to opm-common.


atgeirr:

akva2:
merging so i get to do the file moving before i go home.

PR #1212, opened 2018-01-17
mark module as deprecated

akva2:


atgeirr:
Should we also remove the contents of the repository just to guard against possible messing up by users?

akva2:
We did not to that for porsol or polymer but if you think it better i see no reason not to as such.

atgeirr:
There were no problems with this connected to porsol or polymer, so I guess we'll just let it be.

andlaus:
the OPM github organization ( https://github.com/OPM ) gets a bit messy if abandoned repositories are kept. maybe we should add a second organization (e.g., OPM-archive?) and move such repositories there. That said, since modules that do not receive commits are listed at the bottom, I do not think that this is big issue.

